PR #39, opened 2019-08-06
Auto update vs links

orta:
When the site is published to staging, this scripts the NPM API to grab the latest stable version, then looks through the [extensions marketplace](https://marketplace.visualstudio.com/search?term=typescript&target=VS&category=Tools&vsVersion=vs15&subCategory=All&sortBy=Relevance) for an extension by us which has the same version number. If it doesn't it won't change it, and it'll check again on the next deploy.


orta:
Re: https://github.com/microsoft/TypeScript/issues/32729

scripts/updateURLs.js
```diff
@@ -0,0 +1,77 @@
+// @ts-enable
```
weswigham: wait. It's `//@ts-check`.

orta: Hah, that does explain why it wasn't doing it :D

orta: Thanks!

uniqueiniquity:
If the version numbers ever slip out of sync (this happens rarely, but it does happen), will it be a mess to update manually until the following release?

orta:
If the version numbers are messed up it won't change the JSON file, so that can be manually changed 👍

uniqueiniquity (approved):

Issue #25, opened 2018-05-16
Update Playground Examples


RyanCavanaugh:
The TypeScript Playground examples are... not great. We deserve a new set.

General principles
 * Don't lean on classes
 * Show off type inference where possible - minimal annotations
 * Emphasize working with JavaScript
 * Show both error-finding and good intellisense

# Please Suggest More!

-------------

# Basic Types & Inference
```ts
const array = ["red", "green", "blue", "yellow"];
array.forEach(color => {
  // TypeScript knows 'color' is a string and warns on the incorrect method name
  console.log(color.toUppercase());

  // Try typing 'color.' on this line to see members of string
  color
});

// TypeScript knows s has one of two string values.
// This can detect some different errors
const s = Math.random() > 0.5 ? "aaa" : "bbb";
// ... like the spelling error here
if (s === "aa") {
 // oops
}
// ... and the logic error here!
if (s === "aaa" && s === "bbb") {
  // oops
}
```

# Functions
```ts
function check(s: string) {
  return s.length > 3;
}

// TypeScript warns if we get the argument count wrong
check();
// Or if the argument type is wrong
check(10);

// The return type of the function is inferred,
// so errors like this are automatically detected
if (check("hi") === "world") {
  // Never true because 'world' isn't a boolean
}
```

# Union Types
```ts
// TypeScript can work with objects of mixed types
const items = ["red", 10, 15, "blue"];
for (const i of items) {
  if (typeof i === "string") {
    // TypeScript knows 'i' is 'string' here
    console.log(i.toUpperCase());
  } else {
    // 'i' is of type 'number' here
    console.log(i * 2);
  }
}



```

# Generics

# ES6 Classes

# What else?

orta:
Blast from the past on this issue - but I've moved it to the TS website now, I've tried to think of the examples from quite far back - helping folks look for simpler JS use cases (if you're new) all the way you to advanced TS things like conditional types. What do folks think of this general progression?

<img width="707" alt="Screen Shot 2019-07-11 at 6 49 46 PM" src="https://user-images.githubusercontent.com/49038/61090472-ca2cb080-a40c-11e9-9782-977841c37681.png">

( To do it I'd have to add JS support, and type acquisition to the playground - but that's achievable )

orta:
There isn't done for showing how to work with the DOM! Added

orta:
[...]

PR #43, opened 2019-08-12
Adds updated examples

orta:
This PR is a collect-all for playground based documentation. Unlike the handbook which aims to be thorough and definitive - the examples in the playground are based on providing a few primitives per example and a narrative about why/when you'd use them in a browser environment

Screen Shot 2019-08-12 at 12 25 02 PM

Screen Shot 2019-08-17 at 12 03 06 PM

---

You can try it in https://5d6c47101cff21a89609f849--ts-playground.netlify.com/play/index.html?strict=false#example/objects-and-arrays

I'm now good for content feedback!

---

- [x]  JavaScript
    - [x]  External APIs
       - [x]  TypeScript + Deno 
       - [x]  TypeScript + Node 
       - [x]  TypeScript + React x
       - [x]  TypeScript + Web 
       - [x]  TypeScript + WebGL 
    - [x]  Functions with JavaScript
       - [x]  Function Chaining 
       - [x]  Generic Functions 
       - [x]  Typing Functions 
    - [x]  Helping with JavaScript
       - [x]  Errors 
       - [ ]  Quick Fixes 
       - [ ]  Warnings 
    - [x]  JavaScript Essentials
       - [x]  Code Flow 
       - [x]  Functions 
       - [x]  Hello World 
       - [x]  Objects and Arrays 
    - [x]  Modern JavaScript
       - [x]  Async:Await 
       - [x]  Immutability 
       - [x]  Import:Export 
       - [x]  JSDoc Support 
    - [x]  Working With Classes
        - [x]  Classes 101 
        - [x]  Generic Classes 
        - [x]  Mixins 
        - [x]  This 
- [x]  TypeScript
    - [x]  Advanced Meta-Types
       - [x]  Conditional Types 
       - [x]  Discriminate Types 
    - [x]  Language
       - [x]  Soundness 
       - [x]  Type Widening:Narrowing 
    - [x]  Language Extensions
       - [x]  Enums 
       - [x]  Types vs Interfaces 
    - [x]  Meta-Types
       - [x]  Indexed Types 
       - [x]  Mapped Types 
       - [x]  Nominal Typing 
    - [x]  Primitives
       - [x]  Any 
       - [x]  Literals 
       - [x]  Union and Intersection Types 
       - [x]  Unknown and Never 
    - [x]  Type Primitives
        - [x]  Built-in Utility Types 
        - [x]  Nullable Types 
        - [x]  Tuples


Examples/JavaScript/JavaScript Essentials/Hello World.ts
```diff
@@ -0,0 +1,34 @@
...
+const hello = "Hello"
+
+// You can see on the right the result of the TypeScript compiler:
+// this is vanilla JavaScript which can run on browsers, servers
+// or anywhere really.
+
+const world = "World"
+
+// You can see how it makes tiny changes to the code, by converting
+// a "const" to a "var". This is one of the many things TypeScript
```
sandersn: it's not doing this right now

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
```diff
@@ -0,0 +1,96 @@
+//// { order: 1, compiler: { strict: false } }
```
sandersn: these comments appear in the buffer, which is not bad but a little confusing.

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
```diff
@@ -0,0 +1,96 @@
...
+const pie = {
+  type: "Apple"
+}
+
+const purchaseOrder = {
+  owner: userAccount,
+  item: pie
+}
+
+// If you use your mouse to hover over some of these words (try purchaseOrder above)
```
sandersn: on my 1440px screen, I get 76 characters per pane, which cuts off a few of these lines. Is it worthwhile to reformat to avoid this? I don't know how common 1440 and smaller widths are these days.

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
```diff
@@ -0,0 +1,96 @@
...
+console.log(purchaseOrder.item.type)
+
+// If you hover your mouse over each part of the code between the ()s, you can 
+// see TypeScript offering more information about each part. Try re-writing this below:
+
+// Copy this in the next line, character by character: purchaseOrder.item.type
+
+
+
+// TypeScript provides feedback to the Playground about what JavaScript objects
```
sandersn: I would put this, or part of it, before the "Copy this in the next line" to make it clear to expect completions. (The transition from quick info in the previous paragraph wasn't obvious to me.)

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
```diff
@@ -0,0 +1,96 @@
...
+// are available  in this file and lets you avoid typoes and see additional 
+// information without having to look it up in another place.
+
+// TypeScript also offers these same features to arrays. Here's an array with 
+// just our purchase order above in it.
+
+const allOrders = [purchaseOrder]
+
+// If you hover on allOrders, you can tell it's an array because the hover info
+// ends with []. You can access the first order by using square brackets with an
```
sandersn: You might simplify the type of allOrders somehow. The fully inlined type is scary looking for newbies, I think.

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
```diff
@@ -0,0 +1,96 @@
...
+// called mutating the array, because it changes the underlying data inside it.
+
+const poppedFirstOrder = allOrders.pop()
+
+// Now allOrders is empty. Mutating data can be useful for many things, but 
+// one way to reduce the complexity in your codebases is to avoid mutation. 
+// TypeScript offers a way to declare an array readonly instead:
+
+// Originally: const allOrders = [purchaseOrder]
+const readonlyOrders: readonly typeof purchaseOrder[] = [purchaseOrder]
```
sandersn: I think introducing type aliases before now is a good idea so that you can write `readonly Order[]` here. I'm not sure how, but that alias could help simplify the type of `allOrders` as well.

Examples/JavaScript/JavaScript Essentials/Code Flow.ts
```diff
@@ -0,0 +1,60 @@
...
+// In the above case find could fail, and in that case you don't have an
+// object. This is written like: { name:string } | undefined 
+// The pipe acts as the separator between different types.
+
+// You can use this in a few ways, lets start by looking at an array
+// where the variables are not all the same.
+
+const identifiers = ["Hello", "World", 24, 19]
+
+// Could be a vinyl or cassette
```
sandersn: comment seems leftover from a cooler example

Examples/JavaScript/JavaScript Essentials/Functions.ts
```diff
@@ -0,0 +1,63 @@
+//// { order: 2 }
```
sandersn: I would turn off noImplicitAny for this example and remove the ts-ignores, since the examples don't rely on either one really.

Examples/JavaScript/JavaScript Essentials/Functions.ts
```diff
@@ -0,0 +1,63 @@
...
+const add1 = (x: any, y: any) => { return x + y }
+add1("Hello", 23)
+
+// This is legitimate JavaScript (strings can be added like this for example) but isn't optimal
+// for our function which we know is for numbers, so we'll convert the x and y to only be numbers
+
+const add2 = (x: number, y: number) => { return x + y }
+add2(16, 23)
+add2("Hello", 23)
+//   ~~~~~~~ Argument of type '"Hello"' is not assignable to parameter of type 'number'.
```
sandersn: I think I would remove this comment and change the next paragraph to say something like "Great. You can see from the squiggly underline that we get an error when ..."

RyanCavanaugh:
Fixes #25

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
orta: Hah, great point... I'm now [on an ultra wide](https://www.dell.com/en-us/shop/dell-ultrasharp-34-curved-usb-c-monitor-u3419w/apd/210-arcl/monitors-monitor-accessories?prg=1&VEN1=12839518-5513721-AwEAAAAAAAAAAWlf&AID=5513721&cjevent=2e2fc897bd3311e9815d03f50a240614&dgc=CJ&DGSeg=DHS&cid=198375&lid=45846&acd=12309198375458460&VEN3=811705054906337422) so my experiences are _real_ far from reality nowadays. Thanks will try bump it down to 60

Examples/JavaScript/JavaScript Essentials/Objects and Arrays.ts
orta: It won't eventually 👍

Examples/JavaScript/JavaScript Essentials/Hello World.ts
orta: The compiler settings aren't being set yet - it will 👍

Examples/JavaScript/JavaScript Essentials/Code Flow.ts
orta: Yeah, it turned out it was really hard to have an example like:

```ts
const products = [{ name: "Cassette", rewound: true }, { name: "Vinyl", age: 32 }]

// Could be a vinyl or cassette
const firstProduct = [products][0]

// We can use the JavaScript "x in y" syntax to check for one
// property which is only in vinyl. You can hover to see how
// firstProduct changes between different locations
if ("rewound" in firstProduct) {
  firstProduct
} else {
  firstProduct
}
```

because of how we union the objects in the array: 

```ts
const products: ({
    name: string;
    rewound: boolean;
    age?: undefined;
} | {
    name: string;
    age: number;
    rewound?: undefined;
})[]
```

They need to have different type roots for this example to work I think. Maybe I can make that pretty cheap conceptually by using the `typeof` above to make the type definition next to the instance.

orta:
Type Acquisition for lodash and jest is rough - Lodash's types are kinda crazy, and Jests seems to send my browsers into an infinite loop.

orta:
Yep, I got all those tools set up for projects before - will definitely do that 👍

Examples/en/TypeScript/Language/Soundness.ts
```diff
@@ -0,0 +1,109 @@
+// Without a background in type theory, you're unlikely
+// to be familiar with the idea of a type system being "sound"
+
+// Soundness is the idea that the compiler can make guarantees
+// about what types are available at runtime, and not just
```
jack-williams: Maybe: _Soundness is the idea that the compiler can make guarantees about the type a value has a runtime, and not just during compilation._

<sup>Apologies if you aren't taking feedback atm</sup>

orta: I am - this is a solid improvement 👍

orta:
OK, this now has an implementation in the playground https://github.com/orta/typescript-play/pull/17 

This can be tested in [this draft PR](https://5d6c47101cff21a89609f849--ts-playground.netlify.com/play/index.html?strict=false#example/objects-and-arrays)

Examples/en/TypeScript/Language/Soundness.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+function listenForEvent(eventType: "keyboard" | "mouse", handler: (event: InputEvent) => void) {}
+
+// You can re-declare the parameter type to be a subtype of
+// the declaration. Above, handler expected a type InputEvent
+// but in the below usage examples - TypeScript accepts
+// a type which has additional properties.
+
+listenForEvent("keyboard", (event: KeyboardInputEvent) => {});
```
jack-williams: It might be worth mentioning that this unsoundness is correctly detected using `strictFunctionTypes`, or having a reference to why this is unsound (linking to something on strictness settings).

Examples/en/TypeScript/Language/Soundness.ts
```diff
@@ -0,0 +1,110 @@
+// Without a background in type theory, you're unlikely
```
jack-williams: Another unsound feature that might be worth mentioning is this (I can't think of an example that isn't contrived right now, so maybe it's not a good inclusion!)
```ts
// We can always add an extra optional property to a type.
const obj: { x: string; y? : number } = { x: "a property "};

// However in combination with the ability to 'forget' properties this can
// can be unsound.

const objXY = { x: "a property", y: true };
const objX: { x: string } = objXY;
const objXYNumber: { x: string, y?: number } = objX;

const num: number = objXYNumber.y === undefined ? 0 : objXYNumber.y;

// Excess property checking can prevent errors like this when you
// initialise your objects with annotations:

const objXYAnnotation: { x: string } = { x: "a property", y: true }; // error
```

Examples/en/TypeScript/Primitives/Any.ts
```diff
@@ -0,0 +1,33 @@
...
+// Any declares to TypeScript to trust your code as being
+// safe because you know more about it. Even if that is
+// not strictly true. For example, this code would crash:
+
+myObject.x.y.z;
+
+// Using an any gives you the ability to write code closer to
+// original JavaScript with the trade off of type safety.
+
+// Any is considered a top type in type theory, which means
```
jack-williams: This is true in TypeScript, but not true in the gradual typing literature.

I think a more accurate way of describing it from a theoretical point of view (which may not be useful in this context) is roughly:

_any is much like a 'type wildcard' which you can replace with any type  (except never)_ to make one type assignable to the other.

_Each call to debug is allowed because I could replace the `any` with the type of the argument to match._
```ts
declare function debug(value: any);
	
debug("a string");
debug(23);
debug({ color: "blue" });
```

_The call to swap is allowed because the argument type can be matched by replacing the first `any` in `pair` with `number` and the second `any` with `string`._
```ts
declare function swap(x: [number, string]): [string, number]
declare const pair: [any, any];
swap(pair)
```

orta: Ah yeah, a type wildcard is a great analogy 👍

orta:
`keyof` isn't really in here, IMO maybe it should go in mapped types or in the the included types?

Examples/en/TypeScript/Primitives/Union and Intersection Types.ts
```diff
@@ -0,0 +1,59 @@
...
+type OddNumbersUnderTen = 1 | 3 | 5 | 7 | 9;
+type AMessyUnion = "hello" | 156 | { error: true };
+
+// If the use of "open" and "closed" vs string is
+// new to you, check out: example:literals
+
+// We can mix different types into a union, and
+// what we're saying is that the value is one of those types.
+
+// TypeScript will then leave you to figure out how to
```
chriskrycho: A suggestion: linking from here to the bit on widening and narrowing (as you do below for the *inverse*) is likely to be helpful – as it is, this might feel slightly overwhelming to someone brand new to this idea!

Examples/en/TypeScript/Language/Soundness.ts
orta: Added, and updated the handbook with this  - thanks

Examples/en/TypeScript/Primitives/Union and Intersection Types.ts
orta: Added! Thanks

orta:
The newest build now includes a revised root for documentation:

Artboard

alloy:
w00t! So good, thanks @orta 🎉 👏👏👏

PR #47, opened 2019-08-15
Look  at adding Danger + PR builds via GitHub Actions

orta:
This adds the ability for every PR to be previewable - only TS staff members will get the extra build (because it requires a secret token which only they get when making PRs) but Danger will post a message in the PR.

Also does spell checking.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,10 +1,10 @@
-name: "Danger JS"
+name: "PR Checks"
 on: [pull_request]
 
 
 jobs:
   build:
-    name: "Validate it builds"
+    name: "Builds and Deploys"
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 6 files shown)


orta:
OK, this is good.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,36 @@
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    # Now try build the site
+    - run: cd src/play; yarn install; yarn setup; cd ../..
+    - run: bundle install
+    - run: gulp publish
+
+    # Deploy if we can
+    - run: npx now site/ --token=$NOW_ACCESS_TOKEN --name=typescript-playbook-$PR_NUMBER
```
styfle: Nice!

I suspect this won’t work for PRs from users outside the org (unless the secret is shared). 

But it’s great to see ZEIT Now in action! (pun intended 😉)

orta: Hah, I did consider this - but I figured it'd be easier to handle when it's actually on a PR without access to the keys rather than guessing now

PR #70, opened 2019-09-25
Some fixes

elibarzilay:
After some discussion with @orta.


orta:
The red is fine, I know what to do about this 👍

PR #74, opened 2019-09-27
Improve actions

elibarzilay:
Mainly move the shared steps to a single file.


orta:
Solid uses of the `if` options there!

PR #99, opened 2019-10-26
Replace the word "subset" with "superset"

sanjay-sridharan:
TypeScript is not a subset of JavaScript, it is a superset.


orta:
Hah, yeah, great point

PR #103, opened 2019-11-01
TSConfig Reference Overview 

orta:
## End Goal

Screen Shot 2019-11-01 at 11 37 47 AM
 - editor on the left with Monaco +JSON schema autocomplete, cmd clicking on a key scrolls the right hand side
 - right hand side is a long scrollable (linkable) reference for each TSConfig option

## This PR

Creates the long scrollable reference as a markdown file, by collating a series of smaller markdown files into one markdown file per language. Yep, this is internationalized already. 

The markdown ref from the first draft: [English](https://gist.github.com/orta/d6c12748cad968f046db8950277b9f37#declaration) - [Pirate]() ( The copy for the "[basic types](https://gist.github.com/orta/b542d2f780fd26d9db7cbe2e52f0d3e9#Basic_Options_6172)" is different from English (but the URL is the same) and [allowJs](https://gist.github.com/orta/b542d2f780fd26d9db7cbe2e52f0d3e9#allowJs) too. )

Each config option has it's own markdown file, and in this layout we can really fill out those docs and people can trivially make PRs to edit and improve these docs as one-offs.


orta:
I think it's feasible to ship this page as just another markdown file and slowly build it out to be the full designed version with monaco post re-design

Issue #97, opened 2019-10-24
Playground: stuck waiting for unpkg.com


AlCalzone:
**TypeScript Version:**  not applicable

**Search Terms:** playground unpkg

**Code**
![grafik](https://user-images.githubusercontent.com/17641229/67500387-de767f80-f682-11e9-9512-01e25201d856.png)

**Expected behavior:**
The playground loads immediately, then fetches 3rd party definitions

**Actual behavior:**
The playground is stuck waiting for unpkg.com.

**Playground Link:**
Not applicable. Open the playground with a very limited bandwidth.

**Related Issues:** none

orta:
It would be feasible for the playground to host copies of all the monaco-editors we offer in the dropdown, re-hosting them could be an automated step in the install pipeline.

I'll take a look when I get back for the v3 unless unpkg gets real bad. Can't really move off it for ATA though.

AlCalzone:
> Can't really move off it for ATA though.

Can't that be deferred until the editor was initially loaded? So one can at least start typing while ATA happens in the background?

orta:
Yeah ATA is totally async and deffered, what you're seeing is the downloading of Monaco in which case you can't type anything because it's not loaded the editor yet

AlCalzone:
Ok, then feel free to close this.

orta:
Closing, the v3 playground keeps all of the essential code on it's [own CDN](https://github.com/orta/make-monaco-builds/) - only ATA and Playground plugins come from unpkg (which are non-essential, and non-blocking)

PR #106, opened 2019-11-03
Adds search to the site

orta:
Uses [Algolia DocSearch](https://community.algolia.com/docsearch/) which has become a staple in big OSS projects. It works _really_ well. 

Screen Shot 2019-11-03 at 11 07 00 AM

Works fine on mobile also.

To avoid render-blocking networking, the site will download a copy of the JS/CSS during deployment and will work with that. Given the site is deployed every 2/3 days then we'll always be up-to-date, but maybe lag a tiny bit behind. I'd like to do this for the playground too eventually https://github.com/microsoft/TypeScript-Website/issues/97


orta:
Hrm, looks like some of the automation isn't working like I expected in the PR builds.

orta:
TODO: 

- [ ] Try get the center nav centered again (maybe make search the same width as the site logo)
- [ ] see what's up with the PR build
- [ ] investigate why iOS has a slight white offset

orta:
alright, that looks good to go

PR #108, opened 2019-11-04
Removed non-imports and non-working imports from sample

DanielRosenwasser:
The `require` call doesn't work in TS files, and the `.ts` path was still causing errors to be reported


orta:
👍

orta:
I'd like to get the playground to suppress these two errors eventually and bring these back at some point

DanielRosenwasser:
Well the second is an error, but the require call just doesn't work unless you're in a JS file.

orta:
In this case, `import compose from "https://deno.land/x/denofun/lib/compose.ts"` is not an error in Deno. See: https://www.typescriptlang.org/play/index.html?e=5#example/typescript-with-deno

PR #119, opened 2019-11-09
Move over some of the new handbook TSConfig docs

orta:
Moves in existing copy for TSConfig examples, doesn't touch them, will keep that new repo as source of truth until this feels presentable.

Interesting aside: There are three competing standards for how we could do metadata for code samples.

- Fourslashy `// @something` inside the code
- YML front-matter:  https://github.com/orta/lsif-tsc-remark-example/blob/master/example.md
- Prefix HTML comments - 
  ```
  <!--@
  name: shapes.ts
  -->
  ```


orta:
This one is non-controversial

PR #122, opened 2019-11-10
Add PWA support to the new site

orta:
4b2

Well, that wasn't much work. The site will now uses service workers and an offline cache. Gatsby basically did it all.


orta:
OK, that looks good 👍  from 0 -> 1

PR #123, opened 2019-11-10
Ignore pointer events on skyline image

KubaJastrz:
Banner text is not selectable, because the skyline image is above.

---

Other solution would be to move the banner text above the image:
```scss
.banner {
  position: relative;
  z-index: 1;
}
```


orta:
👍 thanks!

Issue #120, opened 2019-11-09
TypeScript-powered Code Blocks


orta:
## Things I want for all code blocks inside website v2:

- Accurate Syntax Highlighting for complex TS ([ex](https://blog.andrewbran.ch/polymorphic-react-components/))
- Hover support for symbols ([ex](https://blog.andrewbran.ch/polymorphic-react-components/))
- Inline reporting of errors ([ex](https://microsoft.github.io/TypeScript-New-Handbook/chapters/more-on-functions/#object))
- The ability to tell a code block to switch from TS code to _any_ arbitrary part of the tsc emit ([ex](https://microsoft.github.io/TypeScript-New-Handbook/chapters/more-on-functions/#object))
  - a JS file 
  - a d.ts file
  - a source map
- The ability to create a working imports inside a codeblock (TS/JS/DTS)
- The ability to highlight lines, a line, or a range inside the line ([ex](https://microsoft.github.io/TypeScript-New-Handbook/everything/#parameter-type-annotations))

#### Sidegoal: 

Make this not directly tied to the website so that anyone with a Gatsby blog can re-use this for their own docs with very little setup.

#### Places with existing infra this can build from:

- [`orta/lsif-tsc-remark-example`](https://github.com/orta/lsif-tsc-remark-example) - an example of using the LSIF format (static language server protocol) to extract metadata. Overall, I think it's too much to go through the entire tsc-lsif, we could probably just steal the syntax for hover tokens and diagnostics and call it enough maybe.

- [`microsoft/TypeScript-New-Handbook`](https://github.com/microsoft/TypeScript-New-Handbook/blob/master/scripts/sample-compiler.ts) the handbook has a good set of existing examples for this, these could be re-formatted into Gatsby/remark plugins.

- [`andrewbranch/gatsby-remark-vscode`](https://github.com/andrewbranch/gatsby-remark-vscode) re-using VSCode at build time to handle the syntax highlighting. 

- [`andrewbranch/blog`](https://github.com/andrewbranch/blog/tree/master/src/utils/typescript) - uses the typescript API to inline annotate embedded TypeScript ( [long form on how it works](https://blog.andrewbran.ch/overengineering-a-blog/) )

/cc @andrewbranch @RyanCavanaugh

orta:
Bikeshed: There are a bunch of ways in which embedded the complier metadata

@andrewbranch's JSON after language

```
'''ts { compiler: 'format' }
const myThing = ""
'''
```

[...]

PR #129, opened 2019-11-15
Adds a new library for code samples

orta:
Step one in the https://github.com/microsoft/TypeScript-Website/issues/120 - creates a library which handles the pre-conversion of a code sample into something which will then be embedded inside code samples.

Basically it takes the new-handbook code sample importer, wraps it with tests via fixtures. Then uses those fixtures as documentation in the README so as it gets built out, there'll be more info.

### DIFF
--- a/packages/ts-twoslasher/.gitignore
+++ b/packages/ts-twoslasher/.gitignore
@@ -1,0 +1,8 @@
+*.log
+.DS_Store
+node_modules
+.rts2_cache_cjs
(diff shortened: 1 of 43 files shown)


orta:
Easy link to the rendered README: https://github.com/microsoft/TypeScript-Website/blob/c6cdfdcc1abf4f3e393cc127eba5c763a30773a1/packages/ts-twoslasher/README.md

PR #131, opened 2019-11-20
Adds twoslash to the gatsby side

orta:
See: https://twitter.com/orta/status/1197228756370366465 because I didn't save the GIF - doh


orta:
This PR turned into a bit more work than I expected, because this was the first time there was interlinking between deps I had to make sure they were all built beforehand.

PR #155, opened 2019-12-19
Bunch of faff to make LSP responses be reliably placed

orta:
- https://github.com/gatsbyjs/gatsby/pull/20189
- Switched to _temporarily_ use there Nord theme for shiki
- improved testing and logging


orta:
I'm gonna the the fails for now

PR #157, opened 2019-12-19
Polish for the TSConfig reference

orta:
- Runs prettier against all docs to remove formatting
- Adds some extra polish to some of the wording
- Makes this the first URL using the new `/en/` prefix.  The TSConfig is translated into two languages, `en` and `pirate`. It's not much, but it proves the concept.
- Makes the page pretty and functional;
- Some docs updates

Screen Shot 2019-12-19 at 3 28 56 PM


orta:
Not done, but good enough to keep iterating.

https://typescript-v2-157.ortam.now.sh/en/tsconfig

PR #160, opened 2019-12-22
Starts the infra to move the playground to v2

orta:
Works in tandem with https://github.com/orta/make-monaco-builds/pull/3

I can't directly tie the playground using something like node_modules, because the JavaScript cannot be built & linked up-front. We need to support running older, nightly  or PR builds of TS. 

This PR hooks up:

-  AMD exports for the typescript sandbox package
- require JS  + Monaco inside the static page, grabbed at runtime
- the require config for using our hosted build of Monaco editor


orta:
OK, that's good to start moving

PR #166, opened 2019-12-27
update type qualifier for url script headers

pnguyen12:
The headers used in the following function `getLatestVSExtensions` does not get reassigned.

This PR includes the change from var to const.


orta:
cool! thanks

PR #167, opened 2019-12-31
Improve the gatsby remark shiki tests

orta:
Adds tests which uses fixtures to generate both HTML and the twoslash as JSON

This is needed because the 2-3 examples in the package work, but there's a bunch of cases where the inline lsp results aren't right in the tsconfig ref.


orta:
Tests pass

PR #173, opened 2020-01-08
Polish plugin support and makes ready for beta testing

orta:
Hooks up all the last buttons for the v2 version, and adds all the previous behavior back.


orta:
this is still the text color from code samples, it's fine for now

PR #175, opened 2020-01-10
Change tuple to array in a few Examples

Userbit:
More probably in these examples assumed of array usage rather than tuple.


orta:
You're right! Thanks

Issue #100, opened 2019-10-27
Website Internationalization Roadmap


orta:
As the new website is being built out bit-by-bit, we're adding support for internationalized content on a per-section basis. I've been working with l33t speak under the language "vo" to test out the infra as it is built.

I expect there are (at least) four main components for creating content on a per-language basis:

- Website copy (navigation bar, index page, community page etc)
- Website (and handbook) content (markdown files mainly)
- Playground examples content, and on-screen copy 
- TSConfig reference content, and on-screen copy

Each one of these needs:

- Working English documentation as reference
- Infra for supporting the creation of internalized content
- Infra for supporting the presentation of internationalized content

Current state of internationalization:

- Website copy. 
  - [x] English
  - [x] Creation
  - [x] Users

- Website content (e.g. the handbook). 
  - [x] English
  - [x] Creation
  - [x] Users

- Playground examples:
  - [x] English
  - [x] Creation
  - [x] Users

- TSConfig:
  - [x] English
  - [x] Creation
  - [x] Users

## How To Contribute

1. Join the [TypeScript Discord](https://discord.gg/typescript) - come say hi the [`ts-web-translate`](https://discord.gg/gG7nsGA)

2. Clone the repo. The work will happen on the [`v2`](https://github.com/microsoft/TypeScript-Website/tree/v2) branch. Deploys of the TS website have this static site set up at http://www.typescriptlang.org/v2/

  ```sh
  git clone https://github.com/microsoft/TypeScript-Website
  cd TypeScript-Website

  code .
  code README.md

  # Follow setup instructions in README
  # then start up the server
  yarn start
  ```

3. Look below to see the sections, then come back up here

4. Read this PR to see an example of adding a language: https://github.com/microsoft/TypeScript-Website/pull/181

5. Pick one of the below sections to start with, it might be easiest to work with samples in the Playground first FWIW.

#### Website

The website supports localization via React-intl, as of today only the navigation bar it translatable, but with time this will grow. The infra just came in with https://github.com/microsoft/TypeScript-Website/pull/180

There are a series of keys and values for each language, with a fallback to English if you miss something. The TypeScript compiler will keep it all in check.

#### Playground

You can cover the examples in the dropdown in the playground (and in the site footer on desktops.) [Full docs here](https://github.com/microsoft/TypeScript-Website/tree/v2/packages/playground-examples)

[...]

PR #180, opened 2020-01-15
Support internationalization of text inside the website

orta:
Screen Shot 2020-01-15 at 2 37 52 PM

Note the nav above: this PR allows code inside React to be translated. Adds the website section to #100

Interesting things because of this:

- Bumped node version of this project to minimum of 13, because that is when internationalization was added natively. We don't use any of those APIs yet (maybe never (like dates/number formatters etc) but better to be sure)

- I switched form "leet" to "vo". The locales are verified to be real in the i8n code, and well, "leet" isn't an ISO 639-1 code. So I found the one with only 20 language speakers in the world: [Volapük](https://en.wikipedia.org/wiki/Volapük)

As an aside, I did some work on the code syntax coloring - it now passes the a11y audits on CI.


orta:
back to green merges 🥇

PR #181, opened 2020-01-15
Create an example PR which shows how to add a language

orta:
I'm not going to merge this PR (most of the translation was done via Google Translate) but this PR exists to show you a working example of adding a language to the TypeScript website:

I added the beginnings of Portuguese, click the images to go through to the PR deploy:

[Screen Shot 2020-01-15 at 6 03 46 PM](https://typescript-v2-181.ortam.now.sh/pt/tsconfig)

[Screen Shot 2020-01-15 at 6 04 10 PM](https://typescript-v2-181.ortam.now.sh/pt/play)

Will leave a review with inline info now

### DIFF
--- a/packages/playground-examples/copy/pt/JavaScript/JavaScript Essentials/Code Flow.ts
+++ b/packages/playground-examples/copy/pt/JavaScript/JavaScript Essentials/Code Flow.ts
@@ -1,0 +1,23 @@
+//// { title: "Fluxo de código", order: 3, compiler: { strictNullChecks: true } }
+
+// Como o código flui dentro de nossos arquivos JavaScript pode afetar
+// os tipos em nossos programas.
+
+const users = [{ name: 'Ahmed' }, { name: 'Gemma' }, { name: 'João' }]
+
+// Vamos ver se conseguimos encontrar um usuário chamado "João".
+const joao = users.find(u => u.name === 'João')
+
+// No caso acima, 'find' pode falhar. Nesse caso, nós
+// não tem um objeto. Isso cria o tipo:
+//
+//   { name: string } | undefined
+//
+// Se você passar o mouse sobre os três usos a seguir de 'joao' abaixo,
+// você verá como os tipos mudam dependendo de onde a palavra está localizada:
+
+if (joao) {
+  joao
+} else {
+  joao
+}
--- a/packages/playground-examples/copy/pt/sections.json
+++ b/packages/playground-examples/copy/pt/sections.json
@@ -1,0 +1,47 @@
+{
+  "sections": [
+    {
+      "name": "JavaScript",
+      "id": "JavaScript",
+      "subtitle": "Veja como o TypeScript aprimora o dia a dia trabalhando com JavaScript com uma sintaxe adicional mínima."
+    },
+    {
+      "name": "TypeScript",
+      "id": "TypeScript",
+      "subtitle": "Explore como o TypeScript estende o JavaScript para adicionar mais segurança e ferramentas."
+    },
+    {
+      "name": "3.7",
+      "id": "3.7",
+      "subtitle": "Consulte <a href='https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/'>Notas da versão</a>."
+    },
+    {
+      "name": "Playground",
+      "id": "Playground",
+      "subtitle": "Saiba o que mudou neste site."
+    }
+  ],
+
+  "sortedSubSections": [
+    // JS
+    "JavaScript Essentials",
+    "Functions with JavaScript",
+    "Working With Classes",
+    "Modern JavaScript",
+    "External APIs",
+    "Helping with JavaScript",
+    // TS
+    "Primitives",
+    "Type Primitives",
+    "Meta-Types",
+    "Language",
+    "Language Extensions",
+    // Examples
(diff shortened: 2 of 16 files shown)


packages/playground-examples/copy/pt/JavaScript/JavaScript Essentials/Code Flow.ts
```diff
@@ -0,0 +1,23 @@
+//// { title: "Fluxo de código", order: 3, compiler: { strictNullChecks: true } }
```
orta: ## Playground support

The original file is "Code Flow.ts" which will override [this file](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/playground-examples/copy/en/JavaScript/JavaScript%20Essentials/Code%20Flow.ts) - in English the filename is the title, but you can change it to your own language

packages/playground-examples/copy/pt/JavaScript/JavaScript Essentials/Code Flow.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Como o código flui dentro de nossos arquivos JavaScript pode afetar
+// os tipos em nossos programas.
+
+const users = [{ name: 'Ahmed' }, { name: 'Gemma' }, { name: 'João' }]
+
+// Vamos ver se conseguimos encontrar um usuário chamado "João".
+const joao = users.find(u => u.name === 'João')
```
orta: I switched the code from "Jon" to "João" to fit the audience. I think _some_ of this is good, but I would ask that you please keep a diverse set of names in the examples (gender / nationality is mainly what I strive for)

packages/playground-examples/copy/pt/sections.json
```diff
@@ -0,0 +1,47 @@
...
+      "name": "3.7",
+      "id": "3.7",
+      "subtitle": "Consulte <a href='https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/'>Notas da versão</a>."
+    },
+    {
+      "name": "Playground",
+      "id": "Playground",
+      "subtitle": "Saiba o que mudou neste site."
+    }
+  ],
```
orta: This file dictates the name and metadata - given that the names were so small, it didn't really need changing

packages/playground-examples/copy/pt/sections.json
```diff
@@ -0,0 +1,47 @@
...
+    "Meta-Types",
+    "Language",
+    "Language Extensions",
+    // Examples
+    "Syntax and Messaging",
+    "Types and Code Flow",
+    "Fixits",
+    // Playground
+    "Config",
+    "Tooling"
```
orta: This uses filenames, so it might be redundant. In the future I might remove this from non-english to simplify translator's jobs.

packages/tsconfig-reference/copy/pt/categories/Additional_Checks_6176.md
```diff
@@ -0,0 +1,5 @@
+---
```
orta: ## TSConfig Support

These are the names of the category of compiler options, as well as an overview of the grouping.

packages/tsconfig-reference/copy/pt/options/files.md
```diff
@@ -0,0 +1,26 @@
+---
+display: "Arquivos"
+oneline: "Incluir uma lista definida de arquivos, não suporta globs"
+---
```
orta: This is the docs for an individual tsconfig option: `"files"` - you can change the name here, but it will only change the header. 

At the top it will still say files:

Screen Shot 2020-01-15 at 6 17 44 PM

Then further down it will show both this `display`, and the compiler flag name:
Screen Shot 2020-01-15 at 6 18 21 PM

`oneline` is used in the playground TSConfig reference, you should have one for every option though.

packages/typescriptlang-org/src/copy/pt.ts
```diff
@@ -0,0 +1,9 @@
+import { defineMessages } from "react-intl"
+import { navCopy } from "./pt/nav"
+import { docCopy } from "./en/documentation"
+import { Copy } from "./en"
+
+export const lang: Copy = defineMessages({
+  ...navCopy,
+  ...docCopy,
+})
```
orta: This is how we do language translation in the React side, you have an object which returns keys and values and you incrementally add new references. In this case only the `navCopy` comes from the `pt` subfolder instead of English.

```ts
import { navCopy } from "./pt/nav"
```

packages/typescriptlang-org/src/copy/pt/nav.ts
```diff
@@ -0,0 +1,16 @@
...
+  nav_documentation_short: "Documentos",
+  nav_download: "Download",
+  nav_connect: "Conectar",
+  nav_playground: "Parque infantil",
+  nav_search_placeholder: "Documentos de pesquisa",
+  nav_search_aria: "Pesquise no site TypeScript",
+  nav_beta: "Pesquise no site TypeScript",
+  nav_beta_notification:
+    "Nota: esta página é uma página beta, não confie nos problemas de URL e de arquivos <a>no microsoft/TypeScript-Website</a>",
+}
```
orta: This object first fills with the English copy, then adds the Portuguese copy.

PR #185, opened 2020-01-17
Adds some 3.8 examples

orta:
I need to deploy a version of 3.8 for the playground (vs the nightly builds) but this should be the docs side


johnnyreilly (approved):
That is one awesome piece of documentation!

PR #212, opened 2020-02-04
make getDTS to store tsWorker.d.ts

Quramy:
Hi @orta . I'm loving Playground Plugin systems you created 😄 

In some plugins, I want to access language service methods using TypeScript worker such as:

```ts
  modelChanged: (sandbox, model) => {
     sandbox.getWorkerProcess.then(worker => {
         return worker.getSemanticDiagnostics(model.uri.toString());
     }).then(result => /* processing diagnositic */)
  },
```

But for now, the `getDTS` script replaces `TypeScriptWorker` type to `any` and I can't auto-complete worker's methods.

I think it's better to fetch tsWorker.d.ts too. How about it?

### DIFF
--- a/packages/create-typescript-playground-plugin/template/scripts/getDTS.js
+++ b/packages/create-typescript-playground-plugin/template/scripts/getDTS.js
@@ -22,13 +22,18 @@ const go = async () => {
   }
 
   await getFileAndStoreLocally(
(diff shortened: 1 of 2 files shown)


orta:
Perfect, I didn't really consider that - nice call 👍

PR #218, opened 2020-02-05
fix: type mismatch #217

### LINKED ISSUE #217: Type mismatch in at Advanced Types handbook
Here https://github.com/microsoft/TypeScript-Website/blob/v2/packages/handbook-v1/en/Advanced%20Types.md#conditional-types at line 1299 it reads:
```ts
type T12 = ReturnType<(<T>() => T)>;  // {}
```
but `T12` type is `unknow` actually.
Which can be verified by compiling:
```ts
type Expect<T, U, B extends boolean = true> = T extends U
 ? B extends true
   ? true
   : never
 : B extends false
   ? false
   : never

type IsUnknow<T> = Expect<T, unknown>

const test1: IsUnknow<ReturnType<<T>() => T>> = true
```
--updated
I would open a PR, but a CLA is required and I not willing to sign it. But this might be a good first issue for someone.

evanSe:
fixes #217

This section of advanced types seems to have been written during TypeScript 2.8 and was correct then. TypeScript 3.0 introduced the unknown top type making the type now `unknown` rather than `{}`

@orta the documentation is not locked to a specific version right, so should all doc follow latest TypeScript version?

### DIFF
--- a/packages/handbook-v1/en/Advanced Types.md
+++ b/packages/handbook-v1/en/Advanced Types.md
@@ -1296,7 +1296,7 @@ class C {
 
 type T10 = ReturnType<() => string>;  // string
 type T11 = ReturnType<(s: string) => void>;  // void
-type T12 = ReturnType<(<T>() => T)>;  // {}
+type T12 = ReturnType<(<T>() => T)>;  // unkown
 type T13 = ReturnType<(<T extends U, U extends number[]>() => T)>;  // number[]
 type T14 = ReturnType<typeof f1>;  // { a: number, b: string }
 type T15 = ReturnType<any>;  // any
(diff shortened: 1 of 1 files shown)


packages/handbook-v1/en/Advanced Types.md
```diff
@@ -1296,7 +1296,7 @@ class C {
 
 type T10 = ReturnType<() => string>;  // string
 type T11 = ReturnType<(s: string) => void>;  // void
-type T12 = ReturnType<(<T>() => T)>;  // {}
+type T12 = ReturnType<(<T>() => T)>;  // unkown
```
orta:
```suggestion
type T12 = ReturnType<(<T>() => T)>;  // unknown
```

orta:
Yep, there's no versioning

PR #219, opened 2020-02-06
Playground example pages

orta:
This creates one-off pages for every playground page in every language, to make them appear in search indexes. The pages redirect you with JavaScript to the right code sample in the playground straight away when loaded.


packages/typescriptlang-org/src/templates/play.tsx
```diff
@@ -8,6 +8,8 @@ import { RenderExamples } from "../components/ShowExamples"
 import { useIntl } from "react-intl";
```
orta: Moves all strings into i8n

PR #229, opened 2020-02-09
Fix Angular link in Web Frameworks section

janclarin:
Link previously pointed to the TS in 5m overview doc.

Fixes this link:

Screen Shot 2020-02-09 at 3 07 42 PM


orta:
Thanks!

PR #258, opened 2020-02-16
[Bugfix] Windows related errors

KingDarBoja:
This does fix the path and EOL issues after trying to run `yarn bootstrap` on Windows by using NodeJS built-in methods.

Also, the [fs-extra](https://www.npmjs.com/package/fs-extra) package is need for `playground-examples` copy script to properly work on all OS. However, it is not being listed as devDependency after performing the install step. 🤔


packages/playground-examples/scripts/generateTOC.js
```diff
@@ -124,10 +126,11 @@ langs.forEach(lang => {
...
     // Ensure all subfolders are in the sorted section
     const allSubFolders = []
-    all.forEach(path => {
-      const subPath = dirname(path)
-        .split('/')
+    all.forEach(filepath => {
+      const subPath = dirname(filepath)
+        .split(path.sep)
         .pop()
+      console.log(`Subpath: ${subPath}`)
```
KingDarBoja: Oops! Forgot to remove this log.

orta: Fixed - looks good to me, thanks!

PR #266, opened 2020-02-19
Add gatsby-plugin-typegen instead of codegen

item4:
I think website of TypeScript must be type-safe, right?
But, there is few type definition only.
Almost props of components use `(props: any)`. 😱 
So... I make type definition file automatically with gatsby-plugin-typegen.
This plugin support autogenerated type-def file with gatsby together.

Please read this PR and consider merge.


orta:
Ha nice, I only removed it a day or two ago because I introduced the shared fragments and the old one couldn't handle it!

If you can add the generated types to `gitignore` then this is good to go thanks! Wooo

orta:
GitHub is having issues, I'm gonna manually merge this! Thanks!

cometkim:
Hey @item4 and @orta, I'm author of the plugin. I'm really happy to see that the official TypeScript webpage uses my plugins, appreciates all of your work.

The plugin is currently actively developing. so If you have any trouble to use the plugin, please give feedback.

packages/typescriptlang-org/src/components/IntlLink.tsx
```diff
@@ -1,11 +1,7 @@
 import React from "react"
 import { GatsbyLinkProps, Link, graphql } from "gatsby"
-
-export type AllSitePage = {
-  nodes: {
-    path: string
-  }[]
-}
+import { SitePage } from "../__generated__/gatsby-types";
+export type AllSitePage = { readonly nodes: ReadonlyArray<Pick<SitePage, 'path'>> };
```
orta: I just changed this to:

```ts
import { AllSitePageFragment } from "../__generated__/gatsby-types";
export type AllSitePage = AllSitePageFragment["allSitePage"];
```

in case you don't know about the type accessor feature in TS

orta:
👍  it's working so far

packages/typescriptlang-org/src/components/IntlLink.tsx
item4: oh. It's nice. thanks your kindness :)

Urigo:
that's awesome!
We also use that plugin (we are a bit biased as it's using https://github.com/dotansimha/graphql-code-generator under the hood ;) )

@orta if you think this is good, I think it would be a good idea to update your amazing blog post on Gatsby and Typescript.
I assume a lot of people will read it and use it as a resource!

PR #272, opened 2020-02-20
Update to TS 3.8 (and bump any older deps too)

orta:
- 3.7 -> 3.8 required no breaking changes
- new tsconfig compiler option values 
 - #266 required a bit of tweaking


orta:
The white gap isn't showing up in builds - so I'm accepting this

PR #275, opened 2020-02-23
Fix removeComments config example's description.

Quramy:
I found this in my jp translation(#220) work and it seems that the description about `removeComment` is wrong. If tsconfig has no  `removeComments` setting, TypeScript emits original comments.


orta:
Perfect- thanks!

PR #276, opened 2020-02-23
fix: Mixins may not specify visibility modifiers

Gerrit0:
Removes this error:

```
Property '_scale' of exported class expression may not be private or protected.(4094)
```

https://www.typescriptlang.org/v2/en/play/?q=101#example/mixins


orta:
Cool, thanks!

PR #287, opened 2020-02-26
Use twoslash for example codes of "strictBindCallApply"

Quramy:
I think it's better to use `twoslash` feature to explain which errors `strinctBindCallApply` option throws.

スクリーンショット 2020-02-26 15 57 02
[left: before (latest v2), right: after (this PR)]


orta:
I agree!

PR #339, opened 2020-03-14
Remove fixed line-height to improve readability

Livven:
The fixed line-height of 18px looks very cramped. Simply removing it and using the default improves the readability a lot, e.g. for the paragraphs under "What is TypeScript?" on the landing page. You may try this out using the CSS inspector.


orta:
I'm afraid this is far too big of a design change for me to not do it myself as it effects typography everywhere.

I'll take a look - but there's a reason I would have set this manually.

orta:
Thanks @Livven - I gave it a test run and it looks good

PR #366, opened 2020-03-17
Fixed sintax error in example for set and get accessor in ambient context

jrumandal:
The current example is wrong as tslint return a sintax error.
As per TS annotation: "A 'set' accessor cannot have a return type annotation.(1095)"


orta:
Indeed, so it does! Thanks!

PR #405, opened 2020-03-22
Simplified twoslash renderer

orta:
Trying to simplify the Shiki twoslash output so that I can understand it, and so that it can hopefully not exhibit the occasional missing character


orta:
### Before: 

It would chop many strings into components, then flatten them down into a single string. This... was complex. I barely grok'd it and felt kinda lucky that it was working at all.

### After

We create an array of all things to be inserted into the string, then go from the last one inserting them incrementally - so that offsets don't roll into each other

orta:
Cool! This greened

PR #423, opened 2020-03-29
Use webworkers for twoslash

orta:
Was an interesting thought experiment, but really really hard to set up and not done


orta:
This is now possible via a mon-typescript improvement instead https://github.com/microsoft/monaco-typescript/blob/master/test/custom-worker.html

Issue #283, opened 2020-02-25
Syntax mistake in enum documentation


shajanjp:
Syntax for enum data type is wrong on page [Basic Types](https://www.typescriptlang.org/v2/docs/handbook/basic-types.html)
Open brace is missing.
![Screenshot from 2020-02-25 11-34-37](https://user-images.githubusercontent.com/2943609/75219477-e24a1800-57c2-11ea-9c1e-053794fe250f.png)

lgordey:
I guess it's [twoslasher](https://github.com/microsoft/TypeScript-Website/tree/v2/packages/ts-twoslasher) bug or not? 🤔 

Source code looks correct
```ts twoslash
enum Color {
  Red,
  Green,
  Blue
}
let c: Color = Color.Green;
```

orta:
Yeah, I think this is a twoslash bug - I need to get https://github.com/microsoft/TypeScript-Website/issues/202 in which should fix a bunch of weird errors like this

orta:
Confirmed this is fixed by #405 - http://www.staging-typescript.org/docs/handbook/basic-types.html#enum

PR #425, opened 2020-03-30
Remove the banner, and the twoslash popovers for identifiers

orta:
This unblocks #283 - but doesn't solve the underlaying problem ( #405 will eventually )


orta:
lol, not quite green there

PR #426, opened 2020-03-30
Add more relaxed default line height

andrewbranch:
| Before | After |
|--------|------|
| Screen Shot 2020-03-30 at 2 35 12 PM | Screen Shot 2020-03-30 at 2 35 29 PM |


orta:
The reds are fine, I need to fix those

PR #427, opened 2020-03-30
Fix next button chevron in Safari

andrewbranch:
| Before | After |
|--------|------|
| Screen Shot 2020-03-30 at 2 40 52 PM | Screen Shot 2020-03-30 at 2 40 46 PM |


orta:
Perfect!

Issue #336, opened 2020-03-11
Feature request: "on this page" should float as you scroll and highlight current section


ur92:
Seems that if "on this page" will be always visible and highlight the current section, it will make navigation a lot easier. 
See [Jest docs](https://jestjs.io/docs/en/getting-started.html) for example 
![image](https://user-images.githubusercontent.com/6205315/76468552-e0a06700-63f4-11ea-95ba-9f4eeca20064.png)

orta:
looks like there was already code for it, but it doesn't seem to be working

pamtbaau:
I would like to expand on this by requesting to also scroll the left side menu.

orta:
I tried that and went back on it, it didn't feel right and scrolling in it is awkward

orta:
This is fixed

PR #431, opened 2020-03-31
Fix nav scrolling with page

tjamesmac:
Fixes the `#handbook-content` lack of stickiness when scrolling down the page in reference to https://github.com/microsoft/TypeScript-Website/issues/336

| **Before**  | **After** |
| ------------- | ------------- |
| 336 before  | 336 after  |

I noticed that the links don't highlight the correct section when clicked on, should I raise a separate issue for that?

### DIFF
--- a/packages/typescriptlang-org/src/templates/handbook.scss
+++ b/packages/typescriptlang-org/src/templates/handbook.scss
@@ -8,7 +8,6 @@
 #handbook-content {
   margin-left: 2rem;
   margin-right: 2rem;
-  overflow: hidden;
 
   @media (max-width: $screen-xs) {
     margin-left: 0;


orta:
Hi, great (CI will be red for now, I'm fixing it in #430) - thanks for the gifs!

Re highlighting, yeah, I thought I had some code for that. I couldn't repro it jumping to the wrong headers, so I'm open to an issue (and I'll look) or a PR. I think I took the code for generating the links directly from the remark titles plugin, so I'd hope they were the same

orta:
interesting, the scrolling indicators work on my setup (light mode, mac, edge)  but they probably need to be offsetted because I have some margins to ensure that clicking links doesn't hide the title

tjamesmac:
Hi, no problem. Think I should have been more clear, the correct section is jumped to. But the blue highlight on the current link is incorrect and is showing the previous 'section' instead.
| **Example**  | 
| ------------- | 
|336 clicking|

Hopefully, that'll explain it better. I'm using dark/mac/chrome by the way.

PR #439, opened 2020-04-01
Moves example underlines to be span

### LINKED ISSUE #428: Error span squiggles are still off
image

orta:
This should probably fix #428


orta:
Let's give it a run

PR #459, opened 2020-04-04
Fix permalink icon colour in dark mode

AbhishekKrishna123:
Changed the colour of permalink icon in dark mode from black to white.

**Before**:
image
**After**:
image

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/main.scss
+++ b/packages/typescriptlang-org/src/components/layout/main.scss
@@ -73,12 +73,13 @@ html {
 }
 
 // Style the permalink icon in dark colour-scheme
-.raised svg path {
+.raised .markdown h1 svg path,
+.raised .markdown h2 svg path,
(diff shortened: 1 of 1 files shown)


packages/typescriptlang-org/src/components/layout/main.scss
```diff
@@ -72,6 +72,13 @@ html {
   }
 }
 
+// Style the permalink icon in dark colour-scheme
+.raised svg path {
```
orta, line 76: I think the s probably too loose (it would apply to almost every SVG in the site!) - if you can narrow it down to something like `.raised .markdown h1 a svg path, .raised .markdown h2 a svg path, .raised .markdown h3 a svg path` ?

AbhishekKrishna123: Thanks, done.

orta:
Thanks!

PR #461, opened 2020-04-04
Adds Ember as additional web framework

scalvert:
This PR adds Ember as an additional Web Framework in the `https://www.typescriptlang.org/docs/home` page. Ember both uses TS internally for large portions of its development (ember.js, glimmer-vm, etc), but also supports TypeScript for development. Ember also recently went through a huge modernization, which both simplifies the framework _and_ takes advantage of a lot of newer language features.

Screen Shot 2020-04-04 at 2 14 14 PM

These changes add Ember in alphabetical order relative to the other JS frameworks (Angular, React, Vue). Additionally ensures that the blurb is translated.

### DIFF
--- a/packages/typescriptlang-org/src/lib/release-info.json
+++ b/packages/typescriptlang-org/src/lib/release-info.json
@@ -3,15 +3,15 @@
   "tags": {
     "stableMajMin": "3.8",
     "stable": "3.8.3",
-    "betaMajMin": "3.8",
-    "beta": "3.8.0-beta",
+    "betaMajMin": "3.9",
+    "beta": "3.9.0-beta",
     "rc": "3.8.1-rc",
     "rcMajMin": "3.8"
   },
   "isRC": false,
-  "isBeta": false,
+  "isBeta": true,
   "releaseNotesURL": "/docs/handbook/release-notes/typescript-3-8.html",
-  "betaPostURL": "https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/",
+  "betaPostURL": "https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/",
   "rcPostURL": "https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-rc/",
   "vs": {
     "stable": {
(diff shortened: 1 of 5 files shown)


orta:
I originally included this in my designs, but then realized that there's no TypeScript docs nor does the CLI help you get set up with TS in Ember.

So, I'm open to having it in it once there's an obvious path for someone wanting to set up a  TypeScript project - which is what all these are for.

orta:
( Not trying to use this to force official doc changes or anything, I'd be happy if it linked to a reputable source of "ember + TS" kinda docs too )

scalvert:
Makes perfect sense. Ember has an official addon that enables Typescript development - [ember-cli-typescript](https://github.com/typed-ember/ember-cli-typescript). Would it be better to link to that?

orta:
Yep, that works!

scalvert:
OK, done. The actual link I used was https://ember-cli-typescript.com, which is the documentation site for the aforementioned github repo.

packages/typescriptlang-org/src/lib/release-info.json
```diff
@@ -3,24 +3,24 @@
...
       "vs2019_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-38beta"
     },
     "beta": {
-      "vs2017_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-38beta",
-      "vs2019_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-38beta"
+      "vs2017_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-373",
+      "vs2019_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-373"
     },
     "rc": {
       "vs2017_download": "https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.typescript-38beta",
```
orta, line 26: Interesting, I thought this was gitignored - but being up to date is no problem 👍

PR #468, opened 2020-04-05
Adds next and previous buttons to the handbook

orta:
Adds next/prev buttons to the bottom of the handbook when pages are chonologically sorted

Screen Shot 2020-04-05 at 10 56 04 AM

For non-chronologically ordered - we could consider re-listing the sections below


orta:
Looks good to me 👍

Issue #484, opened 2020-04-08
Idea: Change order of items in Handbook Navigation for better UX


taghikhavari:
**Improvement Idea:**
Change the order of items in Handbook Navigation for modules and namespaces to a better conceptually order

**Why:**
Usually when developers want to read documents they starts from top to bottom and in "Module Resolution" and "Namespaces and Modules" sections we have some kind of dependency on "Modules" and "Namespaces" sections and right now we have to go back and forth because the order isn't conceptually right

**Alternative ideas:**

orta:
PR was submitted 👍  thanks

PR #485, opened 2020-04-08
Change order of items in Handbook Reference Navigation

taghikhavari:
I change Order because in "Module Resolution" and "Namespaces and Modules" sections we have dependencies on "Modules" and "Namespaces" section and in this order there is no need to go back and forth and have a better user experience.
This PR will close issue #484


orta:
👍

PR #486, opened 2020-04-08
Adds a branding page, fixes some CSS in tool

orta:
I took a few shortcuts (used SVGs instead of perfecting the layouts with divs) but I don't think it's too offensive as I'm mostly funneling people into downloading the asset pack to grab all the files.


orta:
Looks good, and verified it works fine on a phone and just sent a commit for dark mode

PR #497, opened 2020-04-10
Make the playground run in other pages, and start of a workbench page

orta:
After thinking about it for a bit, getting to something like this:

Re: https://github.com/microsoft/TypeScript/issues/35389

Would either mean:
 - taking the sandbox and wrapping it with a custom sidebar infrastructure
 - taking the playground and allowing for custom sidebar plugins

The choice ended up being pretty easy in the end, so the playground now supports custom sets of plugins - which is how this is implemented

Result of this PR:

https://typescript-v2-497.ortam.now.sh/dev/bug-workbench/

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": false,
-  "printWidth": 120
+  "printWidth": 120,
+  "arrowParens": "avoid"
 }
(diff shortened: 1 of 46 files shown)


orta:
and it sets up prettier correctly everywhere too

orta:
Got all the infra hooked up - looks like a bug with grabbing tthe queries - (I bet its an offset thing)

Screen Shot 2020-04-11 at 8 20 14 PM

orta:
Cool, this is good enough for a few of my bugs - will expand it out in time

PR #500, opened 2020-04-12
(fix): strict* checks are true when `strict`

agilgur5:
- in the docs, "Strict Checks" include noImplicitAny, noImplicitThis,
  strictNullChecks, strictFunctionTypes, strictBindCallApply,
  strictPropertyInitialization, and alwaysStrict
  - so they should all have consistent defaults

- noImplicitReturns is *not* part of "Strict Checks" and is actually
  part of "Linter Checks", so this was incorrectly documented
  - so its default should be plain false


orta:
Thanks!

PR #501, opened 2020-04-12
Remove repeated word from handbook-v1 / The Handbook.md page

kyanny:
The word `rush` appears twice in the same sentence, but it looks redundant.


orta:
Good spot! thanks

PR #509, opened 2020-04-13
Make the error pane jut out less

DanielRosenwasser:
**Warning: This was a very naive change**

Personally I prefer the more subtle outline

Before

image

After

image


DanielRosenwasser:
Maybe an alternative: move back by 4px, expand the border by 1px?

image

orta:
I'm into both of those designs, I'll clone it locally to verify all the permutations 👍

orta:
Ran it all through locally - I'm good with this as it is. You did good, this is better.

PR #518, opened 2020-04-15
A few "let's" -> "lets"

DanielRosenwasser:


DanielRosenwasser:
I don't get it, isn't that what's in the document?

orta:
it is - I usually just ignore those reports

PR #522, opened 2020-04-17
fix link for angular

willnorris:
this has already been corrected [in the v2 site](https://www.typescriptlang.org/v2/docs/home).


orta:
Nice, thanks!

PR #527, opened 2020-04-18
Add tsconfig sections

orta:
Before: 
 - One big section of all categories!

After:
 - Intro markdown
 - All Top level fields except `compilerOptions` and `watchOptions` 
 - All `compilerOptions`
 - All `watchOptions` 

Each is its own section. This solves a few design problems:
 - compilerOptions has _a lot_ of categories and felt crushed in the nav bar
 - we can now give information in the sidebar about the prefix
 -


orta:
merge on green

PR #532, opened 2020-04-20
[v2] Point the link for Ruby on Rails to the official rails/webpacker repo

yuki24:
The repo https://github.com/typescript-ruby/typescript-rails doesn't seem to be very active these days, and Rail's official JS integration has evolved in the mean time. I think it's time to switch to [Rails official TS guide](https://github.com/rails/webpacker/blob/master/docs/typescript.md).


orta:
thanks!

PR #540, opened 2020-04-22
Correct point example

crookedneighbor:
Replacement for https://github.com/microsoft/TypeScript-Handbook/pull/1315

```ts
interface Point {
    distanceFromOrigin(point: Point): number;
}
Point.prototype.distanceFromOrigin = function(point: Point) {
    return this.getDistance({ x: 0, y: 0});
}
```
```ts
Point.prototype.distanceFromOrigin = function(this: Point) {
    return this.getDistance({ x: 0, y: 0});
}
```

Trying to make sense of what the `point: Point` is supposed to be doing. It looks like it's an argument for the function, but the function doesn't use it. I suspect it's a typo, but would love to know for sure.

This PR removes it, under the assumption that it is a typo.

### DIFF
--- a/packages/handbook-v1/en/tutorials/Migrating from JavaScript.md
+++ b/packages/handbook-v1/en/tutorials/Migrating from JavaScript.md
@@ -423,9 +423,9 @@ class Point {
 
 // Reopen the interface.
 interface Point {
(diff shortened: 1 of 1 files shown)


orta:
Yeah, I can't see a way in which that's not a typo 👍

PR #543, opened 2020-04-30
Start on updating the templates for modules

orta:


packages/handbook-v1/en/declaration files/templates/global-plugin.d.ts.md
```diff
@@ -4,6 +4,276 @@ layout: docs
 permalink: /docs/handbook/declaration-files/templates/global-plugin-d-ts.html
 ---
 
+## _UMD_
```
sandersn: I'm not sure UMD is a common term (any more?) so perhaps the title should use a more compositional term like .. uh .. "Dual module/global imports"?

I talked to a customer today who missed UMD `export as namespace` completely because it was described on a page named `module.d.ts` and he was looking for a page named `global-mixed.d.ts` or something.

orta:
Paired with @Ethan-Arrowood today on this and got it quite a bit closer to ocmpletion

PR #551, opened 2020-05-06
fix tsconfig include path

sakito21:
This does not work.
スクリーンショット 2020-05-07 1 21 03

Deviation from the English version.
https://github.com/microsoft/TypeScript-Website/blob/v2/packages/tsconfig-reference/copy/en/options/include.md


orta:
Thanks!

PR #553, opened 2020-05-08
Adds an attribution and CTA to help out on handbook pages

orta:
The bits are in, but without polish
Screen Shot 2020-05-07 at 11 33 55 PM

Eventually 

Screen Shot 2020-05-08 at 12 14 53 PM


orta:
Perfect, the CI fail is not legit but the windows stuff not failing is enough

PR #559, opened 2020-05-09
Replace broken link to SystemJS@baseUrl

dandv:
I [can no longer find any documentation about `baseUrl`](https://github.com/systemjs/systemjs/search?q=baseurl&unscoped_q=baseurl).


orta:
Nice one, thanks

PR #563, opened 2020-05-09
Switch button colors to blue.

DanielRosenwasser:
**Don't merge**: realized that the hover colors have to be fixed.

The second half of the screen on a page load is a bar of grey followed by white. I feel like (especially in light mode) this can feel very bland. I think we at least need an accent color here.

Before:

image

After:

image


orta:
I think change throws the above the fold visual hierarchy off quite a lot - I'll leave my feedback in #564

orta:
We discussed offline

PR #566, opened 2020-05-10
Fix wrong link

XYShaoKang:
I added `/docs/` to the href of `Template: Global Module`.

Clicking on `Template: Global Module` in the document does not jump to the corresponding page. 
After searching, it is found that the corresponding link in the source code is missing `/docs/`


orta:
Thank you!

PR #570, opened 2020-05-11
Fix undefined class if className is undefined

ulrichstark:
Section will get 'undefined' class if no className is provided as property

Anmerkung 2020-05-11 102637

This commit fixes the issue


orta:
Great spot!

PR #573, opened 2020-05-11
Swap function type comments

jaimemarijke:
I'm brand-new to TypeScript, so I'm not sure if this is correct, but it looked to me like these two comments were supposed to be swapped.


orta:
Agree! Thanks!

PR #574, opened 2020-05-11
Remove Location and clarify ImageControl error

jaimemarijke:
There seem to a few issues with this code, comments, and descriptions. I took a stab at fixing them. (But I'm brand new to TypeScript, so I might be interpreting it incorrectly).

The Location class doesn't seem to have anything to do with the rest of the example code, and it also throws an unrelated error "Duplicate identifier 'Location'.", which is confusing.

The comment on line 655 also seems incorrect - it seems vestigial from some previous iteration of code. The actual error shown is: "Class 'ImageControl' incorrectly implements interface 'SelectableControl'.  Types have separate declarations of a private property 'state'."

### DIFF
--- a/packages/handbook-v1/en/Interfaces.md
+++ b/packages/handbook-v1/en/Interfaces.md
@@ -652,13 +652,10 @@ class TextBox extends Control {
   select() {}
 }
 
-// Error: Property 'state' is missing in type 'Image'.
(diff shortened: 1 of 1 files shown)


orta:
Yep, I agree with these! 
Thanks

orta:
Thanks for helping out

PR #577, opened 2020-05-13
Fixes dead link.

MicahZoltu:
For whatever reason, this link had an extra `-` which caused it to not match the anchor on the page.


orta:
Thanks!

PR #589, opened 2020-05-18
Use correct quotes for file extensions and settings examples

102:
https://www.typescriptlang.org/docs/handbook/compiler-options.html

| the problem |
| ------ | 
| image |
| image |


orta:
Great! Thanks

PR #591, opened 2020-05-18
Update getLength function example to return length

markhaslam:
From the name of the 'getLength' function it looks like it was intended to return 'obj.length' and not just 'obj'.


orta:
Agree! Thanks

PR #593, opened 2020-05-19
Explain bigint type in handbook

corollari:
[Since typescript 3.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html), ts supports a number type called `bigint` but the handbook says that the only numeric type supported is `number`. This PR fixes that.

### DIFF
--- a/packages/handbook-v1/en/Basic Types.md
+++ b/packages/handbook-v1/en/Basic Types.md
@@ -25,6 +25,7 @@ These floating point numbers get the type `number`, while BigIntegers get the ty
 In addition to hexadecimal and decimal literals, TypeScript also supports binary and octal literals introduced in ECMAScript 2015.
 
 ```ts twoslash
+// @target: ES2020
 let decimal: number = 6;
 let hex: number = 0xf00d;
 let binary: number = 0b1010;
--- a/packages/handbook-v1/en/Basic Types.md
+++ b/packages/handbook-v1/en/Basic Types.md
@@ -20,8 +20,8 @@ let isDone: boolean = false;
 
 # Number
 
-As in JavaScript, all numbers in TypeScript are floating point values.
-These floating point numbers get the type `number`.
+As in JavaScript, all numbers in TypeScript are either floating point values or BigIntegers.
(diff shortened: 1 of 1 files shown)


corollari:
CI is failing because of the compiler target that the snippets are tested against:
```
BigInt literals are not available when targeting lower than ES2020
```

I'm not sure what to do now, I've looked at previous issues/PRs and the docs available and haven't been able to find any instructions on whether this should or shouldn't be accepted.

Furthermore, I'm not sure if this is the right repo to create a PR in, since https://github.com/microsoft/TypeScript-Handbook seems to be the authoritative repo but the `contribute to this page` links on the handbook point to this repo, and all translations seem to be getting contributed here.

If someone could shed some light on these two things I'd appreciate it.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -20,15 +20,16 @@ let isDone: boolean = false;
...
+As in JavaScript, all numbers in TypeScript are either floating point values or BigIntegers.
+These floating point numbers get the type `number`, while BigIntegers get the type `bigint`.
 In addition to hexadecimal and decimal literals, TypeScript also supports binary and octal literals introduced in ECMAScript 2015.
 
 ```ts twoslash
 let decimal: number = 6;
 let hex: number = 0xf00d;
 let binary: number = 0b1010;
 let octal: number = 0o744;
+let big: bigint = 100n;
```
orta, line 32:
```suggestion
// @target: ES2020
let decimal: number = 6;
let hex: number = 0xf00d;
let binary: number = 0b1010;
let octal: number = 0o744;
let big: bigint = 100n;
```

Should fix your compiler issue I'd expect 👍

corollari:
I see, thanks!

What confused me the most was the following phrase in the readme:
> The existing docs for TypeScript as a direct import the repo of microsoft/TypeScript-Handbook

I would suggest changing that to reflect the current status.

Also, kudos for the quick response, this has been one of my best open source contribution experiences and I will definitely contribute something more in the future if the occasion arises.

orta:
Great, perfect, let's get this in and I have a README update coming along after 👍

PR #602, opened 2020-05-24
Fixes broken redirect url

19shubham11:
this is broken - https://www.typescriptlang.org/handbook/literal-types.html


orta:
Thanks!

PR #603, opened 2020-05-25
fix release notes navigation wrong link

holynewbie:
ts

fix navigation wrong link


orta:
Thanks!

orta:
Thanks!

robinvw1:
When will this fix for the broken link be deployed? According to the README, deployment is automatic on merging with `master`. When looking at GitHub actions, I can see it did deploy for staging but not yet for production.

orta:
Both sites are deployed with the fix

PR #604, opened 2020-05-25
Limit tooltip width in gatsby-remark-shiki-twoslash

pveyes:
As promised, part 1 of https://twitter.com/orta/status/1262782512260136962

The implementation can definitely be optimized as we're executing `getBoundingClientRect` to the parent node every time even though it probably never changes. But it works for now and I didn't notice any performance issue.

**Before**
Kapture 2020-05-25 at 14 22 58

**After**
Kapture 2020-05-25 at 14 20 59


orta:
Nice, yeah, this looks good!

PR #605, opened 2020-05-25
Add syntax highlight support in lsp tooltip

pveyes:
Part 2 of https://twitter.com/orta/status/1262782512260136962

This change is independent from #604 so there should be no merge conflict

One notable change is that I have to extract highlighter into separate module so it can be used to highlight both the code (in `index.ts`) and LSP (in `renderer.ts`)


packages/gatsby-remark-shiki-twoslash/src/index.ts
```diff
@@ -1,43 +1,11 @@
...
-    } catch (error) {
-      throw new Error("Unable to load theme: " + theme + " - " + error.message)
-    }
-  }
-
-  return getHighlighter({ theme: shikiTheme, langs: languages }).then(newHighlighter => {
-    highlighter = newHighlighter
-    return highlighter
-  })
-}
```
pveyes, line 40: Moved to `highlighter.ts`

packages/gatsby-remark-shiki-twoslash/src/highlighter.ts
```diff
@@ -0,0 +1,35 @@
...
+    } catch (error) {
+      throw new Error("Unable to load theme: " + theme + " - " + error.message)
+    }
+  }
+
+  return getHighlighter({ theme: shikiTheme, langs: languages }).then(newHighlighter => {
+    highlighter = newHighlighter
+    return highlighter
+  })
+}
```
pveyes, line 35: Moved with no change (other than comment) from `index.ts`

packages/gatsby-remark-shiki-twoslash/src/dom.ts
```diff
@@ -63,8 +63,10 @@ export const setupTwoslashHovers = () => {
 
     // Use a textarea to un-htmlencode for presenting to the user
     var txt = document.createElement('textarea')
-    txt.innerHTML = message
-    tooltip.textContent = txt.value
+    txt.innerHTML = message;
+    tooltip.innerHTML = txt.value
+    // remove shiki wrapper element
+    tooltip.innerHTML = tooltip.querySelector('code')?.innerHTML!;
```
pveyes, line 69: Because we're using the same highlighter logic, the lsp value will be wrapped with `pre` and additional `div` element.

```html
<pre class="shiki twoslash">
  <div class="language-id">
  <code>actual definition here</code>
</pre>
```

Because we only need to content of the `code` element I'm assuming this is safe to do

pveyes (commented):
I've added some comments in case it's not clear enough

orta:
Great, this all looks good - I'll handle getting it merged etc 👍

orta:
Now that https://github.com/microsoft/TypeScript-Website/pull/655 is in, I can look at some of the more complex PRs - this looks good on principal. I've started work on migrating it in but it'll need a second run at tests etc, it is working in the site though 👍

orta:
Alright @pveyes - I'm going to close this! 

In part the world has really moved on a lot, but I have https://github.com/shikijs/twoslash/issues/42 which has a fresh debate on the trade-offs and now shiki-twoslash is a"real" library which lives outside of this repo - this work could inspire the new version in external version 👍🏻 so thanks, sorry I took so long and didn't make a call earlier

PR #609, opened 2020-05-26
Updated broken link

guettli:
This does not exit any more:
http://blogs.msdn.com/b/typescript/archive/2015/11/03/what-about-async-await.aspx


orta:
Perfect, thank you!

PR #611, opened 2020-05-26
Upgrade gatsby-plugin-typegen to v2.2.0

cometkim:
Hi :wave: 

I'm the maintainer of gatsby-plugin-typegen,

There have been many improvements and bugfixes since v1.1.2.

https://github.com/cometkim/gatsby-plugin-typegen#changelog

After this upgrade, types are exposed via the specified namespace (default is `GatsbyTypes`, it can be customized), so you don't need to import it from `__generated__` path!

----

And it generates `SitePageContext` type, so it can replace `any`s in pageContext type.

```ts
type Props = {
  pageContext: GatsbyTypes.SitePageContext, // automatically inferred all possible context by gatsby's query compiler
  data: GatsbyTypes.IndexPageQuery
}

// strict-usage
if (!props.pageContext.lang) {
  throw new Error('lang is required');
}
```

If you know the page context shape correctly, checking the schema type for every page component feels verbose. For the future release, I am looking for a way to provide the exact pageContext type.

### DIFF
--- a/packages/typescriptlang-org/package.json
+++ b/packages/typescriptlang-org/package.json
@@ -39,7 +39,7 @@
     "gatsby-plugin-sass": "^2.1.28",
     "gatsby-plugin-sharp": "^2.4.5",
     "gatsby-plugin-sitemap": "^2.2.27",
-    "gatsby-plugin-typegen": "^1.1.1",
+    "gatsby-plugin-typegen": "^2.2.0",
     "gatsby-remark-autolink-headers": "^2.1.24",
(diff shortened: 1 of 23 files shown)


orta:
Heh, yeah - it has been a while! 

I've wanted to have a solid re-think about the queries for a while (their perf was getting serious because of localization) and I've cleaned up the need for every page to make a query. With that done, I've updated to typegen 2.2 in #1292 - this PR made upgrading a breeze, thanks!

PR #646, opened 2020-05-29
tutorials/TS for the New Programmer: Misc improvements

elibarzilay:
Note: there's a bunch of "issues" as comments in here, since it's nearly impossible to create an issue on a line in an md file since GH does not have an option to show the md source.  These comments should be considered separate issues, which I'll remove if rejected, or write it if accepted.

### DIFF
--- a/packages/handbook-v1/en/tutorials/TS for the New Programmer.md
+++ b/packages/handbook-v1/en/tutorials/TS for the New Programmer.md
@@ -5,57 +5,59 @@ permalink: /docs/handbook/typescript-from-scratch.html
 oneline: Learn TypeScript from scratch
 ---
 
-Congratulations choosing TypeScript as one of your first languages - you're already making good decisions.
+Congratulations on choosing TypeScript as one of your first languages — you're already making good decisions!
 
 You've probably already heard that TypeScript is a "flavor" or "variant" of JavaScript.
-The relationship between TypeScript and JavaScript is rather unique among modern programming languages, so learning more about this relationship will help you understand how TypeScript adds to JavaScript.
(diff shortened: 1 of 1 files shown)


packages/handbook-v1/en/tutorials/TS for the New Programmer.md
```diff
@@ -5,57 +5,59 @@ permalink: /docs/handbook/typescript-from-scratch.html
...
-  // It is! But why??
-}
-```
+  ```js
+  if ("" == 0) {
+    // It is! But why??
+  }
+  if (1 < x < 3) {
+    // True for *any* value of x!
+  }
```
orta, line 39: hah, yeah, what a wild example

packages/handbook-v1/en/tutorials/TS for the New Programmer.md
```diff
@@ -101,55 +103,69 @@ TypeScript, though, considers division of number by an array to be a nonsensical
...
 
 Finally, TypeScript doesn't provide any additional runtime libraries.
 Your programs will use the same standard library (or external libraries) as JavaScript programs, so there's no additional TypeScript-specific framework to learn.
+<!--
+Should extend this paragraph to say that there's an exception of
+allowing you to use newer JS features and transpile the code to an older
+JS, and this might add small stubs of functionality when needed.  (Maybe
+with an example --- something like `?.` would be good in showing readers
+that this document is maintained.)
+-->
```
orta, line 146: Yeah, I think a section on 'transpiling for older runtimes' is a good call here, it's likely someone with zero experience in programming would have no concept of this at all

PR #649, opened 2020-05-29
Summary line improvements

### LINKED ISSUE #628: Idea: Remove (or revise+trim) short descriptions on handbook page?
I generally don't like the short descriptions below each box heading, since they don't contribute any useful information.  For example, "Get Started", plus the link texts makes it obvious that it's an overview for the listed starting points.  I think that with the other suggestion for improving the titles, most or all of these could go away.

elibarzilay:
Fixes #628 -- with a few changes from the suggestions there.


orta:
👍

PR #650, opened 2020-05-29
Fix external links

### LINKED ISSUE #629: Docs: avoid external language-specific links
There are many places with a reference to `https://developer.mozilla.org/en-US/docs` --- the `en-US` part should be removed.  Same goes for the `en` in: `https://en.wikipedia.org/`. (Some of these were even kept in translated texts...)

(I'll be happy to do it if agreed.)

### LINKED ISSUE #636: Docs: packages/handbook-v1/en/tutorials/DOM Manipulation.md -- link URLs
`packages/handbook-v1/en/tutorials/DOM Manipulation.md`, bottom:

Why is this showing the URL text instead of making it just a link?
(And if it's intended, then md has the `<https://...>` syntax.)

elibarzilay:
* Drop `en`/`en-US` from mozilla and WP links.

* Change links with the url showing to text links.

* Note: Does this on `aka.ms/types` -- but maybe it was intended to show
  the url?

* Note: Changes links in release notes too -- but maybe these files
  should be unchanged?

* Note: random fix -- adds backticks around a `tsconfig.json` title

Fixes #636
Fixes #629

### DIFF
--- a/packages/handbook-v1/en/Advanced Types.md
+++ b/packages/handbook-v1/en/Advanced Types.md
@@ -189,7 +189,7 @@ We mentioned these briefly in [the Basic Types section](docs/handbook/basic-type
 By default, the type checker considers `null` and `undefined` assignable to anything.
 Effectively, `null` and `undefined` are valid values of every type.
 That means it's not possible to _stop_ them from being assigned to any type, even when you would like to prevent it.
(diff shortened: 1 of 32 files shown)


orta:
Cool, yeah, looks good

PR #651, opened 2020-05-30
Adds support for dev builds of TS

orta:
Adds the basic support we'd need on the playground side for hooking into a built copy of the compiler.


orta:
"4.0.0-dev"

Screen Shot 2020-05-30 at 7 12 00 AM

orta:
This isn't 100% complete, because this setting needs to be propagated to the d.ts lookup in twoslash, but it's good enough for now

PR #654, opened 2020-05-31
Fix mistake in TS for Functional Programmers doc

tom7:
There is no additional `.` in this syntax. I think the sentence was intended to refer to the parentheses (otherwise 1. looks like the beginning of a floating-point number literal)?


orta:
I agree - I think this has better intent 👍

PR #660, opened 2020-06-02
Bug in roolDice function.

Pierre-Thibault:
Math.random always return a value less than 1 so it was not possible to have a 6.


orta:
You're right! Thanks!

PR #667, opened 2020-06-05
fix(handbook): Use correct import syntax

FlorianDr:
I believe this is a copy and paste error from the above line.

### DIFF
--- a/packages/handbook-v1/en/declaration files/Library Structures.md
+++ b/packages/handbook-v1/en/declaration files/Library Structures.md
@@ -103,7 +103,7 @@ var fs = require("fs");
 In TypeScript or ES6, the `import` keyword serves the same purpose:
 
 ```ts
-import fs = require("fs");
+import fs from "fs";
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
Oddly enough, this is not wrong: http://www.staging-typescript.org/play/?module=1#code/JYWwDg9gTgLgBAMwM5wLxygUwI4FdhYAUARMsQJQDcAUNcgHRYCGAJgGLAA2mAygJ4A7akA ( if you give it a second to download the node d.ts files you'll see it correctly fail )

orta:
Whether it's something we should encourage anymore is a different question, which I'm not sure of yet, will ask around

FlorianDr:
@orta thank you for getting back to me so fast. As a reader, I was confused to find this syntax, because it's referencing ES6 in the line above and I was expecting ES6 modules import syntax.

FlorianDr:
@orta do you have any updates regarding this one?

FlorianDr:
@orta I will close this PR for now. Let me know, if there are any changes.

PR #673, opened 2020-06-09
fix: quick fix index path on spanish docs

KingDarBoja:
This solves the issue of spanish text on website index not being displayed. @orta


orta:
EZ!

PR #685, opened 2020-06-18
Fix "applyMixin" type issue

reggi:
There's a type issue I see in my editor when I use `applyMixin`

I'm not sure the best way to fix this, the naive approach is to check if this value is undefined. 

It this something other people are seeing as well, or is it just my `tsconfig.json` settings?

```
Argument of type 'PropertyDescriptor | undefined' is not assignable to parameter of type 'PropertyDescriptor & ThisType<any>'.
  Type 'undefined' is not assignable to type 'PropertyDescriptor & ThisType<any>'.
    Type 'undefined' is not assignable to type 'PropertyDescriptor'.ts(2345)
```

Screen Shot 2020-06-18 at 6 44 21 PM


orta:
Cool, let's do this - the mixin doc is currently getting a complete rewrite in https://github.com/microsoft/TypeScript-Website/tree/new_mixin so it might not be around for too long

PR #687, opened 2020-06-20
Multi-file playground

orta:
WIP


orta:
Might be able to get a sidebar 'for free' - https://github.com/BlueMagnificent/monaco-tree/blob/master/src/lib/monaco-utils.js#L8-L15

orta:
Unlikely to carry on with this branch if/when multi-file playgrounds exist

PR #690, opened 2020-06-21
Try the new algolia search

orta:
WIP, but it means I can get the a11y bug on search passed 

Screen Shot 2020-06-21 at 2 11 50 PM


orta:
The feel is nicer for sure, will try on mobile when I get a PR build made

Trade-offs: 
 - This uses a bunch of new react things I've never heard of
 - Introduces CSS vars into the app ( I had to do this eventually anyway for code samples)
 - Used to ship the JS/CSS for search async, it isn't anymore ish

TODO: 

 - Need to see how having all this bundled affects DLs
 - Need to verify if it affects running without JS
 - Need to see what it does re:dark mode

orta:
I've been exploring #928 instead

PR #699, opened 2020-06-23
Adds a doc for babel + tsc in a project

orta:
Based on https://twitter.com/sebastienlorber/status/1275067106254393347


orta:
I'm happen enough with this, at least right now I can't think of how to extend it

PR #702, opened 2020-06-24
Add links to JSDoc @type explanation.

sandersn:
Should encourage people to think of using the full range of Typescript types in JSDoc.


orta:
agree!

PR #706, opened 2020-06-26
Dispatch is no longer part of the react-redux library

alexanderkjall:
The Dispatch symbol is no longer exported from react-redux, instead import it directly from redux.


orta:
👍

PR #707, opened 2020-06-26
make the tutorial compatible with redux 4

alexanderkjall:
The createStore function have been expanded to take four template arguments, and the initial state injection have been moved to the reducer.

See this answer for reference https://stackoverflow.com/a/60884430/131697


alexanderkjall:
There is some things here that could really need some more explaining, but that isn't my strongest skill.

orta:
We should probably drop this demo entirely and provide links out to something like https://redux.js.org/recipes/usage-with-typescript/

orta:
We should take this anyway though!

PR #708, opened 2020-06-27
fix: Remove debugger calls

Gerrit0:
The playground `debugger` gets in my way frequently since I like to keep devtools open while in the playground and this triggers on almost every character typed when importing modules. The twoslash call was removed for completeness.


orta:
Yeah, let's do this

PR #712, opened 2020-06-28
adds switch from light to dark mode

### LINKED ISSUE #401: Feature: Add support for changing the colour of code blocks
Hi, the new colour scheme is unreadable. White on grey just blurs into a big blob for me, can you at least bring in a toggle to the old style so I could continue my studies?

dandelionadia:
## Motivation:

Page do not have switcher to the dark/light mode.

## GitHub:

Fixes #401


dandelionadia:
Switcher changes theme of the page to the light/dark, but it doesn't work when it is dark mode by default, just when it is light mode.

orta:
Cool, this is a good move forwards

orta:
👋 - I've picked this top and got it merged in #784 - thanks for the base to work from!

PR #725, opened 2020-07-02
Fix --jsxFragmentFactory description

vlad-zhukov:


orta:
Cool, this checks out - thanks!

PR #742, opened 2020-07-06
Adds unknown as a basic type in place of any and cautions against using any.

MicahZoltu:
`any` should really only be used as a last resort or when porting JS code.  The docs prior to this PR held `any` in a pretty positive light and suggested some uses for it that really weren't appropriate (`unknown` would have been better).  This PR swaps in `unknown` in the spot `any` previously held in the basic types section, moves `any` to the bottom of the list, and includes language that discourages the usage of `any` in favor of other options like `unknown`, Mapped Types, and Index Types.  Fixed up some references to `any` as well, redirecting them to `unknown`.

Also adjusted the wording on one of the `never` comments while I was in here.

### DIFF
--- a/packages/handbook-v1/en/Basic Types.md
+++ b/packages/handbook-v1/en/Basic Types.md
@@ -175,45 +175,36 @@ let colorName: string = Color[2];
 console.log(colorName); // Displays 'Green' as its value is 2 above
 ```
 
-# Any
+# Unknown
 
 We may need to describe the type of variables that we do not know when we are writing an application.
 These values may come from dynamic content, e.g. from the user or a 3rd party library.
-In these cases, we want to opt-out of type checking and let the values pass through compile-time checks.
-To do so, we label these with the `any` type:
+In these cases, we want to provide a type that tells the compiler and future readers that this variable could be anything so we give it the `unknown` type.
 
 ```ts twoslash
-let notSure: any = 4;
+let notSure: unknown = 4;
 notSure = "maybe a string instead";
 notSure = false; // okay, definitely a boolean
 ```
 
-The `any` type is a powerful way to work with existing JavaScript, allowing you to gradually opt-in and opt-out of type checking during compilation.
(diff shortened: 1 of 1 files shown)


MicahZoltu:
The CI failures appear to be unrelated to my PR?  I'm not sure what they mean.

JojOatXGME:
@MicahZoltu, I think the first build failed because [one of your samples](https://github.com/microsoft/TypeScript-Website/pull/742/files#diff-6be1fe681f467b5410c8487059b22df5R193) does not compile. It looks like you need to add

    // @errors: 2322 2322 2322

to the top of the code block. To my understanding, it should tell the build that this example is expected to contain this three errors.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -244,7 +235,7 @@ let n: null = null;
 By default `null` and `undefined` are subtypes of all other types.
 That means you can assign `null` and `undefined` to something like `number`.
 
-However, when using the `--strictNullChecks` flag, `null` and `undefined` are only assignable to `any` and their respective types (the one exception being that `undefined` is also assignable to `void`).
+However, when using the `--strictNullChecks` flag, `null` and `undefined` are only assignable to `unknown` and their respective types (the one exception being that `undefined` is also assignable to `void`).
```
JojOatXGME, line 238: Both types – `any` and `unknown` – should be mentioned here.

MicahZoltu:
Checks passing, thanks @JojOatXGME!  I also applied the feedback you provided.

JojOatXGME:
I created an alternative suggestion in #755 which is based on this one. It combines and rearranges some parts of the orginal and this updated version. (I also added some text of my own.) I wasn't sure if creating another pull request is the right way but I couldn't see another option. What do you think?

orta:
Let's call this one deprecated by https://github.com/microsoft/TypeScript-Website/pull/755

PR #746, opened 2020-07-06
chore: fix bootstrap on windows

wardpeet:
I had to do 2 fixes to make `yarn bootstrap` work

1. Git should have files saved as LF so `os.EOL` doesn't work. Switching to just "\n" works.
1. cp does not exists on windows, cpy-cli can fix this. (happy to use any other package)

### DIFF
--- a/packages/gatsby-remark-shiki-twoslash/package.json
+++ b/packages/gatsby-remark-shiki-twoslash/package.json
@@ -19,7 +19,7 @@
     "lint": "tsdx lint"
   },
   "dependencies": {
-    "@typescript/twoslash": "0.5.1",
+    "@typescript/twoslash": "0.5.2",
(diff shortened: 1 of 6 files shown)


orta:
Cool! I wonder what's up with the `yarn install` - re-running

wardpeet:
@orta, I see that I committed a change with my prettier config. Should I revert? Are you ok with adding husky, prettier perhaps?

I'm happy to do the work

orta:
Yeah, it's fine - that's the right prettier config, I think sometimes my editor gets confused and ships the default

orta:
( I already have that stuff set up, but it only triggers for me )

orta:
There we go, working fine - it was mainly just me pushing to master during work on https://github.com/microsoft/TypeScript-Website/issues/130#issuecomment-653931944

PR #754, opened 2020-07-08
add "createFSBackedSystem"

hipstersmoothie:


packages/typescript-vfs/test/index.test.ts
```diff
@@ -41,6 +44,27 @@ it("runs a virtual environment and gets the right results from the LSP", () => {
...
+  const fsMap = createDefaultMapFromNodeModules(compilerOpts)
+  addAllFilesFromFolder(fsMap, path.resolve(path.join(__dirname, "../../../node_modules/@types")))
+
+  const content = `import * as path from 'path';\npath.`
+  fsMap.set("index.ts", content)
+
+  const system = createFSBackedSystem(fsMap)
+  const env = createVirtualTypeScriptEnvironment(
+    system,
+    ["/node_modules/@types/node/path.d.ts", "index.ts"],
```
hipstersmoothie, line 58: @orta I really want to be able to not support the `@types` file. IDK if this is possible with how the languageService works. It seems you would need to import every file manually, gather it's imports, and then tell the language service to use those files.

hipstersmoothie: >  It seems you would need to import every file manually, gather its imports, and then tell the language service to use those files. 

So i've done this now locally but it doesn't work with how typescript handles the `@node/types`

packages/typescript-vfs/src/index.ts
```diff
@@ -371,7 +375,44 @@ export function createSystem(files: Map<string, string>): System {
...
+      fileName => files.has(fileName) || files.has(libize(fileName)) || fs.existsSync(fileName)
+    ),
+    getCurrentDirectory: () => "/",
+    getDirectories: () => [],
+    getExecutingFilePath: () => notImplemented("getExecutingFilePath"),
+    readDirectory: audit("readDirectory", directory => (directory === "/" ? Array.from(files.keys()) : [])),
+    readFile: audit(
+      "readFile",
+      fileName => files.get(fileName) || files.get(libize(fileName) || fs.readFileSync(fileName, { encoding: "utf-8" }))
+    ),
```
orta, line 408: I think this might be a bit too naive, as it looks like the current dir is "/" and I'm not sure it can resolve to your projects node modules, will take a look

orta: Perhaps the answer is to provide a FS root to work from

hipstersmoothie:
Closing for #757

PR #755, opened 2020-07-08
Adds unknown as a basic type

JojOatXGME:
I created a new suggestion for adding `unknown` based on #742. It adds `unknown` as basic type to the handbook. I combined parts of the original version and parts of #742 and rearranged them. I moved the section about `unknown` and `any` next to each other to provide a direct comparison.

PS: While editing the document, I was wondering if `unknown` has the same properties as `Object | null | undefined`.

### DIFF
--- a/packages/handbook-v1/en/Basic Types.md
+++ b/packages/handbook-v1/en/Basic Types.md
@@ -107,8 +107,10 @@ When accessing an element with a known index, the correct type is retrieved:
 let x: [string, number];
 x = ["hello", 10]; // OK
 /// ---cut---
-console.log(x[0].substring(1)); // OK
-console.log(x[1].substring(1)); // Error, 'number' does not have 'substring'
+// OK
+console.log(x[0].substring(1));
+
+console.log(x[1].substring(1));
 ```
 
 Accessing an element outside the set of known indices fails with an error:
@@ -118,9 +120,9 @@ Accessing an element outside the set of known indices fails with an error:
 let x: [string, number];
 x = ["hello", 10]; // OK
 /// ---cut---
-x[3] = "world"; // Error, Property '3' does not exist on type '[string, number]'.
+x[3] = "world";
 
-console.log(x[5].toString()); // Error, Property '5' does not exist on type '[string, number]'.
+console.log(x[5].toString());
 ```
 
 # Enum
@@ -172,27 +174,22 @@ enum Color {
 }
 let colorName: string = Color[2];
 
-console.log(colorName); // Displays 'Green' as its value is 2 above
+// Displays 'Green'
+console.log(colorName);
 ```
 
 # Unknown
 
 We may need to describe the type of variables that we do not know when we are writing an application.
-These values may come from dynamic content &ndash; e.g. from the user &ndash; or we may want to intentionally accept all value in our API.
+These values may come from dynamic content &ndash; e.g. from the user &ndash; or we may want to intentionally accept all values in our API.
 In these cases, we want to provide a type that tells the compiler and future readers that this variable could be anything, so we give it the `unknown` type.
 
 ```ts twoslash
 let notSure: unknown = 4;
 notSure = "maybe a string instead";
-notSure = false; // okay, definitely a boolean
-```
 
-You might expect `Object` to play a similar role, as it does in other languages.
-However, variables of type `Object` do not allow you to assign `null` or `undefined` when `--strictNullChecks` is specified.
-
-```ts twoslash
-// @errors: 2322
-const notSure: Object = null; // error, type 'null' is not assignable to type 'Object'
+// OK, definitely a boolean
+notSure = false;
 ```
 
 If you have a variable with an unknown type, you can narrow it to something more specific by doing `typeof` checks, comparison checks, or more advanced type guards that will be discussed in a later chapter:
@@ -200,14 +197,21 @@ If you have a variable with an unknown type, you can narrow it to something more
 ```ts twoslash
 // @errors: 2322 2322 2322
 declare const maybe: unknown;
-const aNumber: number = maybe; // error, 'maybe' could be a string, object, boolean, undefined, or an array
+// 'maybe' could be a string, object, boolean, undefined, or an array
+const aNumber: number = maybe;
+
 if (maybe === true) {
-  const aBoolean: boolean = maybe; // ok
-  const aString: string = maybe; // error, 'maybe' is definitely a boolean, so definitely NOT a string
+  // TypeScript knows that maybe is a boolean now
+  const aBoolean: boolean = maybe;
+  // So, it cannot be a string
+  const aString: string = maybe;
 }
+
 if (typeof maybe === "string") {
-  const aString: string = maybe; // ok
-  const aBoolean: boolean = maybe; // error, 'maybe' is definitely a string, so definitely NOT a boolean
+  // TypeScript knows that maybe is a string
+  const aString: string = maybe;
+  // So, it cannot be a boolean
+  const aBoolean: boolean = maybe;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/handbook-v1/en/Basic Types.md
```diff
@@ -175,42 +175,79 @@ let colorName: string = Color[2];
...
+However, variables of type `Object` do not allow you to assign `null` or `undefined` when `--strictNullChecks` is specified.
 
 ```ts twoslash
-// @errors: 2339
-let notSure: any = 4;
-notSure.ifItExists(); // okay, ifItExists might exist at runtime
-notSure.toFixed(); // okay, toFixed exists (but the compiler doesn't check)
+// @errors: 2322
+const notSure: Object = null; // error, type 'null' is not assignable to type 'Object'
+```
```
MicahZoltu, line 196: I think this section should be moved to `any`.  I don't think there is confusion around `Object` vs `unknown` as they fulfill quite different purposes and have relatively little overlap.

MicahZoltu: Actually, I wonder if we should get rid of it entirely since there is a lot more difference between `any` and `Object` than just the ability to assign null/undefined.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -175,42 +175,79 @@ let colorName: string = Color[2];
...
+To do so, we label these values with the `any` type:
+
+```ts twoslash
+declare function getValue(key: string): any;
+const str: string = getValue("myString"); // ok, return value of 'getValue' is not checked
+```
+
+The `any` type is a powerful way to work with existing JavaScript, allowing you to gradually opt-in and opt-out of type checking during compilation.
+Just like `unknown`, variables of type `any` allow you to assign any value to them.
+Unlike `unknown`, TypeScript let you can call arbitrary methods on them, even ones that don't exist:
```
MicahZoltu, line 228: This line is more meant to be a comparison against `Object` rather than `unknown`.  `unknown` is **not** assignable to any type other than `unknown`, while `any` **is** assignable to any type (type checking is disabled).  Consider removing this line all together since there are a lot more differences between `any` and `Object` than just the ability to call arbitrary methods.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -175,42 +175,79 @@ let colorName: string = Color[2];
...
 ```
 
 The `any` type is also handy if you know some part of the type, but perhaps not all of it.
 For example, you may have an array but the array has a mix of different types:
 
 ```ts twoslash
 let list: any[] = [1, true, "free"];
-
 list[1] = 100;
 ```
```
MicahZoltu, line 246: I recommend removing this section, when this was originally written `any` was a solution to this problem, but the language now has `tuples` and union types which are a much better solution to this problem.

orta:
I've addressed @MicahZoltu's feedback and applied my own feedback to this, it feels pretty good to me now 👍

packages/handbook-v1/en/Basic Types.md
orta: Yeah, I couldn't think of a good example like this which doesn't rely on using an interface/type, so I dropped it

packages/handbook-v1/en/Basic Types.md
orta: Agree

packages/handbook-v1/en/Basic Types.md
orta: Agree, I expanded this a bit in a different direction in a paragraph after after

packages/handbook-v1/en/Basic Types.md
```diff
@@ -172,45 +174,87 @@ enum Color {
...
+// @errors: 2322 2322 2322
+declare const maybe: unknown;
+// 'maybe' could be a string, object, boolean, undefined, or an array
+const aNumber: number = maybe;
+
+if (maybe === true) {
+  // TypeScript knows that maybe is a boolean now
+  const aBoolean: boolean = maybe;
+  // So, it cannot be a string
+  const aString: string = maybe;
```
MicahZoltu, line 207: It is not immediately clear to me as a reader that one of these lines is an error, while the other is not an error.  Consider adding something to the comment to indicate that one of the lines will result in a compiler error while the other will not.

Side note: Would be totally cool if the TS docs page actually put the intellisense squiggles on these inline code samples so people could see the actual TSC errors right in the docs!

packages/handbook-v1/en/Basic Types.md
```diff
@@ -172,45 +174,87 @@ enum Color {
...
+}
 
-let prettySure: Object = 4;
-prettySure.toFixed(); // Error: Property 'toFixed' doesn't exist on type 'Object'.
+if (typeof maybe === "string") {
+  // TypeScript knows that maybe is a string
+  const aString: string = maybe;
+  // So, it cannot be a boolean
+  const aBoolean: boolean = maybe;
```
MicahZoltu, line 214: Same issue as above, the way this is currently laid out it is not immediately obvious to me that one of these lines is an error, and if so which one.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -172,45 +174,87 @@ enum Color {
...
 ```ts twoslash
-let list: any[] = [1, true, "free"];
+declare function getValue(key: string): any;
+// OK, return value of 'getValue' is not checked
+const str: string = getValue("myString");
+```
+
+The `any` type is a powerful way to work with existing JavaScript, allowing you to gradually opt-in and opt-out of type checking during compilation.
+
+Unlike `unknown`, TypeScript lets you call arbitrary methods, even ones that don't exist:
```
MicahZoltu, line 233: Consider changing:
> TypeScript lets you call arbitrary methods

to
> TypeScript lets you call arbitrary methods on an `any` variable

Also consider maybe clarifying that all possible properties on an `any` are also `any`, not just functions.  It feels weird to me that functions are being singled out here, when they aren't really a special difference between `unknown` and `any`.  Its just that an `any` object has all possible properties on it, and they all have the `any` type.  I see you mention this briefly below, but I think that brief mention below should take the center stage, rather than the current focus on function properties in particular.

packages/handbook-v1/en/Basic Types.md
orta: It does, that's why I removed them from the comments: https://www.staging-typescript.org/docs/handbook/basic-types.html#tuple

MicahZoltu: Oh man, so cool!

packages/handbook-v1/en/Basic Types.md
JojOatXGME: > I think this section should be moved to `any`. I don't think there is confusion around `Object` vs `unknown` as they fulfill quite different purposes and have relatively little overlap.

I don't know that much about the details how TypeScript handles `Object`, so I might be wrong. To me, `Object` looks much more similar to `unknown` than to `any`. For example in Java:

```java
Object str = "some string";
str.length(); // Error because length() does not exist on Object. With any, that would not be an error.
```

Anyway, I think this section doesn't have much value anymore. Previously, this section highlighting a specific difference between `Object` and `any`. This difference is gone with `unknown` and we already have a comparision between `unknown` and `any` below. I agree that removing the section is the right decision.

packages/handbook-v1/en/Basic Types.md
JojOatXGME: Agree. I think the paragraph is still true in rare cases but that is nothing a new user should worry about and probably to complex to draft here.

packages/handbook-v1/en/Basic Types.md
```diff
@@ -172,42 +174,44 @@ enum Color {
...
+notSure = false;
 ```
 
 If you have a variable with an unknown type, you can narrow it to something more specific by doing `typeof` checks, comparison checks, or more advanced type guards that will be discussed in a later chapter:
 
 ```ts twoslash
 // @errors: 2322 2322 2322
 declare const maybe: unknown;
-const aNumber: number = maybe; // error, 'maybe' could be a string, object, boolean, undefined, or an array
+// 'maybe' could be a string, object, boolean, undefined, or an array
```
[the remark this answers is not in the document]
MicahZoltu: I would remove mention of `function` entirely and adjust the example to include accessing a few different types of properties (including functions and nested objects).  However, I'm also fine with either of the suggestions provided.

orta:
Alright, I've addressed feedback and rebased - 👍

PR #758, opened 2020-07-09
Take another shot at using Berry

orta:
Replaces #451 - This time it leaves node_modules around, and at least seems to build and run locally.


orta:
Alright, let's do it

PR #767, opened 2020-07-13
Make it possible to set noErrors and showEmit to true

d-fischer:
In twoslash, you can only set `noErrors` and `showEmit` to `false` in `defaultOptions`. I changed that to `boolean` and also fixed a small typo in the doc comment.


orta:
Thanks! 👍  looks good

PR #769, opened 2020-07-14
Adds custom config for GitHub CodeSpaces

khaosdoctor:
GH Codespaces is out for a lot of people (including me) and one of the cool features it has is too apply custom configs to different repos. 

I thought this could make the onboarding process easier, so I created the `.devcontainer` folder following [the default TS container template from MS](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/typescript-node-14) to integrate in codespaces.

This container:

- Comes with Node 14 and TS
- Install yarn globally
- Installs Code Spell Checker extension for translating purposes
  - It'd be nice to suggest other extensions based on the language the person is translating/using
- Runs `yarn install` and `yarn bootstrap` after start
- Forwards port `8000` to the client

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,27 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:14
+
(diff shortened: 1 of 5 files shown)


khaosdoctor:
CC @orta

orta:
Cool, this looks good to me 👍

orta:
Does this need some kind of thing to run `yarn start`? or would it be normal to expect someone to run that themselves?

orta:
Eh, let's just see

khaosdoctor:
> Does this need some kind of thing to run `yarn start`? or would it be normal to expect someone to run that themselves?

There's a `postCreateCommand` that's run once the codespace is created, so it _should_ create the whole space automagically

PR #793, opened 2020-07-21
Update TS for the New Programmer.md

ScarboroughCoral:
The result of `4/[]` is `Infinity`,at least in Chrome.


orta:
Agree, same in Safari - thanks

PR #798, opened 2020-07-24
add twoslash to library structure and module template file

Ethan-Arrowood:
Adds twoslash to the Library Structure and module template files. Had to use a lot of `// @noError: true` since a lot of the blocks are around import/export/require stuff.


orta:
I've dropped the bits which were failing and moved it to point at v2

Ethan-Arrowood:
are we thinking of shipping the @no-errors or should I go back through and add the necessary json files to each of the code examples? Does the vfs resolve the package.json stuff correctly now?

orta:
Yeah, it should resolve the package, I was happy enough with the no-errors can improve in the future

orta:
Going to close this as some features were added in different places and I think we're in a good spot for the rest

PR #801, opened 2020-07-25
Fix lighthouse results

orta:
Looks like the files changed positions


orta:
Not worth keeping around

PR #802, opened 2020-07-25
Adds a 'this page in your lang' popover

orta:
Design is WIP

Screen Shot 2020-07-25 at 1 59 05 PM


orta:
The top nav item is always there, but the bottom can be dismissed forever.

Screen Shot 2020-07-26 at 12 08 38 PM

Issue #805, opened 2020-07-26
Translation Request: This Page in Your Language


orta:
Hello!

ja: @bobstrange @DuGlaser @sasurau4 @Quramy
es: @KingDarBoja @fanyCaz
zh: @Kingwl @laosb 
pt: @khaosdoctor @alvarocamillont  @danilofuchs 

I have just added a new feature which kinda _requires_ translating. Normally I **do not** want to make requests about translations, which is why English fallback is totally OK in my mind. For this feature though, I really do need to have it translated for anything to make sense. Sorry!

In #802 I added a popup when you are on a page which is not your browsers language, and there is a native version. 

![](https://user-images.githubusercontent.com/49038/88483879-baf5a380-cf38-11ea-8372-cc4a48e1d19e.png)

There are two design elements:

- "In En" in the top nav
- "This page is available in English" at the bottom

So, I am on a Chinese page, but my browser is in English, so it offers the English page.

However, if I were on a Spanish page on a Japanese browser, the link would redirect correctly but the text would say "Available in English" right now! 😸 This makes it very weird. 

Can you add the equivalent to this file: [`/packages/typescriptlang-org/src/copy/inYourLanguage.ts`](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/copy/inYourLanguage.ts)

So, for Portugese it would look something like:

```diff
export const inYourLanguage: Record<
  string,
  { body: string; open: string; cancel: string; shorthand: string }
 > = {
  en: {
    shorthand: "In En",
    body: "This page is available in English",
    open: "Go",
    cancel: "Don't ask again",
  },
+  pt: {
+    shorthand: "Em Pt",
+    body: "Isso pagina tem em Portugues",
+    open: "Avancar",
+    cancel: "Nao pergunte novamente",
+  },
}
```

You should be able to see it in action at:

- https://www.staging-typescript.org/zh
-  https://www.staging-typescript.org/pt

KingDarBoja:
Sure, glad to help with this task 🚀

danilofuchs:
Sure! I will create a PR for the Portuguese version

danilofuchs:
Just a note, @orta: The `don't ask again` button may overflow depending on the translation:
![image](https://user-images.githubusercontent.com/17657014/88490050-0c208a00-cf6f-11ea-8003-f530ce229619.png)

orta:
[...]

PR #812, opened 2020-07-27
Propose new translation to go

khaosdoctor:
Just proposing a new translation to the button: "Go" in the `in your language` pop-up (as refered in https://github.com/microsoft/TypeScript-Website/issues/805). 

In my opinion, "Go" in the meaning of just "Ir" in Portuguese seems kinda vague. I believe the term "Visitar" (or "Visit") might be a better option.

What do you think @orta @danilofuchs?

### DIFF
--- a/packages/typescriptlang-org/src/copy/inYourLanguage.ts
+++ b/packages/typescriptlang-org/src/copy/inYourLanguage.ts
@@ -17,7 +17,7 @@ export const inYourLanguage: Record<
   pt: {
     shorthand: "Em Pt",
     body: "Esta página está disponível em Português",
-    open: "Ir",
+    open: "Visitar",
     cancel: "Não perguntar novamente",
   },
   ja: {
--- a/packages/typescriptlang-org/src/copy/inYourLanguage.ts
+++ b/packages/typescriptlang-org/src/copy/inYourLanguage.ts
@@ -17,7 +17,7 @@ export const inYourLanguage: Record<
   pt: {
     shorthand: "Em Pt",
     body: "Esta página está disponível em Português",
-    open: "Trocar Língua",
+    open: "Trocar língua",
     cancel: "Não perguntar novamente",
   },
   ja: {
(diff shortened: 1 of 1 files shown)


danilofuchs:
Honestly, I don't recall any real life examples of this kind of CTA. I didn't really like `Ir` but it seemed to be the simplest option. In the original issue, @orta suggested to use `Avançar`, but I think it would not be clear enough.

I can see many alternatives:

- Ir (Go)
- Visitar (Visit)
- Acessar (Access)
- Avançar (Go)
- Mudar (Change)
- Trocar (Change)
- Mudar língua (Change language)
- Trocar língua (Change language)
- Mudar para Pt (Change to Pt)
- Traduzir (Translate) _not sure about this one_

I really like `Mudar língua` or `Trocar língua`, and it seems like most users will understand it easily. What do you think?

orta:
Change language was one of my original explorations too

khaosdoctor:
I like "Traduzir" and "Trocar Língua", but I was thinking on the button size in the site. Do you think it is going to fit?

orta:
I think both buttons will need to be able to be 2 lines elegantly anyway (German localization is nearly always English chars * 2) so you have the space to make it a bit verbose. I would expect most people to maybe read it once or twice and then never show it again

packages/typescriptlang-org/src/copy/inYourLanguage.ts
```diff
@@ -17,7 +17,7 @@ export const inYourLanguage: Record<
   pt: {
     shorthand: "Em Pt",
     body: "Esta página está disponível em Português",
-    open: "Ir",
+    open: "Visitar",
```
khaosdoctor:
```suggestion
    open: "Trocar Língua",
```

khaosdoctor:
@orta that's true... I updated the translation to what we have discussed here. 

See what you guys think!

packages/typescriptlang-org/src/copy/inYourLanguage.ts
```diff
@@ -17,7 +17,7 @@ export const inYourLanguage: Record<
   pt: {
     shorthand: "Em Pt",
     body: "Esta página está disponível em Português",
-    open: "Ir",
+    open: "Trocar Língua",
```
danilofuchs:
```suggestion
    open: "Trocar língua",
```
image
image

Looks better IMO

orta:
👍 cool, thanks folks

PR #835, opened 2020-08-02
fix The definition for "super-greeter"

B2D1:


orta:
great, thanks!

B2D1:
> great, thanks!

Hi, maybe I found another typo on this page, it should be `declare module "super-greeter"`, not `export`

orta:
Not in this case, you use that for global declarations but for modules we assume that you're replicating the file system too

PR #847, opened 2020-08-04
Consistent analogy of polymorphic nature of types as sets.

soloham:
Because types are sets, which can have subsets it's better to describe polymorphic nature of types in terms of these subsets.


orta:
I think I'm going to skip on this, I'm not sure it adds enough to extra clarity to the idea vs the original to the point-ness. Thanks for trying to improve our docs though!

PR #849, opened 2020-08-05
Fixed typo

Christoph-Wagner:


packages/typescriptlang-org/src/copy/en/index.ts
```diff
@@ -38,7 +38,7 @@ export const indexCopy = {
 
   index_dts: "Types On Every Desk",
   index_dts_copy: `
-<p>Most of the worlds JavaScript’s is un-typed, and inference can only go so far. To address this, the TypeScript team helps maintain <dt>Definitely Typed</dt> - a community project to provide types and inline documentation to existing JavaScript.
+<p>Most of the worlds JavaScript is un-typed, and inference can only go so far. To address this, the TypeScript team helps maintain <dt>Definitely Typed</dt> - a community project to provide types and inline documentation to existing JavaScript.
```
DanielRosenwasser:
```suggestion
<p>Most of the world's JavaScript is un-typed, and inference can only go so far. To address this, the TypeScript team helps maintain <dt>Definitely Typed</dt> - a community project to provide types and inline documentation to existing JavaScript.
```

DanielRosenwasser: @orta, does the site use smart or simple apostrophes throughout?

orta:
Thanks - good catch

packages/typescriptlang-org/src/copy/en/index.ts
orta: simple

PR #858, opened 2020-08-06
fix Chinese typo

Kingwl:


Kingwl:
@laosb @stkevintan PTAL

PR #872, opened 2020-08-07
Fixes Union Exhaustiveness checking description

ericchernuka:
The previous description referred to an older example than the current example. This updates the references to refer to the code snippet presented below it.


orta:
Thanks!

PR #888, opened 2020-08-11
Fix typo in module template name

luke-john:
Templates directory seems to only have a module-plugin file which I'm assuming is the one this is meant to reference

https://github.com/microsoft/TypeScript-Website/tree/v2/packages/documentation/copy/en/declaration-files/templates


orta:
it is, thanks!

PR #890, opened 2020-08-12
Fix link to `object` type information in "Do's and Don'ts"

cherryblossom000:
Fix link to `object` type information in the declaration files "Do's and Don'ts".


orta:
thanks!

PR #894, opened 2020-08-12
Chore: Format var keyword as code

JonathanWbn:
The `var` keyword is also formatted as code earlier in the same paragraph.


orta:
👍🏻 thanks

PR #893, opened 2020-08-12
Translate nav into ja

bobstrange:
Part of #220

Navigation bar translation.

- Original
	- [`nav`](https://github.com/microsoft/TypeScript-Website/blob/v2/packages//typescriptlang-org/src/copy/en/nav.ts?rgh-link-date=2020-02-06T15%3A46%3A01Z)
- Diff
	- https://github.com/microsoft/TypeScript-Website/commit/88539cfbebf444d4ee16360791db5cddd1bb2844

PR #898, opened 2020-08-13
Remove beta navigation translation key

bobstrange:
Remove `nav_beta` related things.

https://github.com/microsoft/TypeScript-Website/pull/893#discussion_r469907813


orta:
Thank you!

PR #905, opened 2020-08-14
Rewrite TS for Programmers

rafpaf:
I've tried to make the writing clearer and more active.


orta:
Heh, when I see a 'I've re-written this entire doc' PR - its normally something a PR I tend to not look forwards to.

However, this doc was written by me and so I know all my motivations for building it the way it is. I like your changes, basically all of them, I think you quite often make the right point terser and with a bit more punch. It's good work, thank you.

rafpaf:
Glad I could help!

PR #918, opened 2020-08-17
small docs update

orta:
I wrote this for a chat in the discord


orta:
Unsure what's up with windows, maybe that file can't be represented? Gonna assume a false-negative

Issue #901, opened 2020-08-13
Playground: Not obvious how to share links


DanielRosenwasser:
The playground lets you share using the typical "save" key combo, but there's nowhere I can click to share a plain link, which is kind of a pain if you're new to the site.

Issue #900, opened 2020-08-13
Playground: Ctrl+S to share the playground session doesn't work if you try too soon


DanielRosenwasser:
If I open up the playground and immediately hit `Ctrl+S`, I'll get a pop-up with `undefined` in it.

Try it with a nightly to make sure you're getting something not cached.

PR #920, opened 2020-08-17
Playground fixes

### LINKED ISSUE #908: Playground: Saving playground doesn't capture current code on document load
Type some stuff into

https://www.typescriptlang.org/play

Hit <kbd>Ctrl</kbd>+<kbd>S</kbd> or <kbd>Cmd</kbd>+<kbd>S</kbd> to save the URL.

Paste this URL somewhere.

Then navigate back to

https://www.typescriptlang.org/play

Don't type anything. Hit <kbd>Ctrl</kbd>+<kbd>S</kbd> or <kbd>Cmd</kbd>+<kbd>S</kbd> to save the URL.

The URLs end up not being the same; instead you'll get `https://www.typescriptlang.org/play`

orta:
Fixes #908 #901, and I think #900 is a dupe of #908 so fixes #908 too

### DIFF
--- a/packages/typescriptlang-org/src/templates/pages/community.tsx
+++ b/packages/typescriptlang-org/src/templates/pages/community.tsx
@@ -41,17 +41,6 @@ export const Comm: React.FC<Props> = props => {
   const intl = useIntl()
   const i = createInternational<typeof comCopy>(intl)
 
-  // Make an array of all the meetups which share a rough location
-  const meetupByArea = {}
-  meetups.forEach(meetup => {
-    const area = meetup.meetup.continentish
-    if (meetupByArea[area]) {
-      meetupByArea[area].push(meetup.meetup)
-    } else {
-      meetupByArea[area] = [meetup.meetup]
-    }
-  });
-
   return (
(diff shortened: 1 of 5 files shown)


packages/typescriptlang-org/src/templates/pages/community.tsx
```diff
@@ -193,23 +182,21 @@ export const Comm: React.FC<Props> = props => {
...
-                      <img src={require("../../assets/community/meetup-logos/" + meetup.image)} className="icon img-square" alt={"logo of " + meetup.title} />
-                      <div>
-                        <h4 className="community-callout-headline">{meetup.title}</h4>
-                        <div className="text">{meetup.country}<br /><a rel="noopener" target="blank" href={meetup.url}>Website</a>{meetup.twitter ? <a rel="noopener" target="blank" href={meetup.twitter}>Twitter</a> : null}</div>
-                      </div>
-                    </Col>
-                  ))}
+
+          <div className="callouts">
+            {meetups.sort().map(m => m.meetup).map(meetup => (
```
DanielRosenwasser, line 187: 1. Is it okay that this mutates `meetups`?
2. If these are objects, what is this sorting by?
3. You can turn this into 1 `map` call

PR #921, opened 2020-08-17
Allow translation codeowners merging of root copy TS files

danilofuchs:
These root `pt.ts`, `es.ts`, etc. files are always changed when new copy files are added


orta:
Thanks! Makes sense

PR #933, opened 2020-08-18
Static web app upload for forks

orta:


orta:
OK, so that's the uploading side - I have to merge to work on the downloading side

Issue #839, opened 2020-08-03
Idea: Dark mode transition back not applying light theme


orta:
**Improvement Idea:**

- Be in OS dark mode
- Switch to light
- Refresh
- Switch to dark 

It won't change live, but it will change when you refresh

orta:
I figured this

PR #948, opened 2020-08-20
Fixed dark mode switching bug

ZibanPirate:
This PR fixes the bug #839

The bug happens only on staging and production, locally it doesn't happen. the cause of this bug is that, the server-side rendered HTML is having light-them by default, thus the switch is checked
```html
<div class="switch-wrap">
    <input type="checkbox" checked>
    <div class="switch"></div>
</div>
```
and then, when the JS loads, due to the locally stored dark-them value, React will render this (removes the `checked` attribute)

```html
<div class="switch-wrap">
    <input type="checkbox">
    <div class="switch"></div>
</div>
```
which will cause the `onChange` event to fire for the input element (the switch), and here where the bug happens, because it's updating the`isDarkMode` like this:
```ts
setDarkMode(!isDarkMode)
```
instead of taking the input (the switch) value
```ts
setDarkMode(!event.currentTarget.checked)
```
this ensures that, whatever you select in the switch, will set the value of the `isDarkMode`

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/SiteFooter.tsx
+++ b/packages/typescriptlang-org/src/components/layout/SiteFooter.tsx
@@ -221,8 +221,8 @@ export const SiteFooter = (props: Props) => {
   const useDark = !customThemeOverride && systemIsDark ? true : customThemeOverride === "dark-theme"
   const [isDarkMode, setDarkMode] = useState(useDark)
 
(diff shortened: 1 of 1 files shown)


orta:
What a great explanation, thanks for the PR - solid work.

khalilmansouri:
well done

PR #951, opened 2020-08-21
Swap mapped types example order

bennettdams:
The description of mapped types starts with optional properties and `readonly` after that, but the examples start with `readonly` and optional after that.  This PR swaps the order of the examples.

image


orta:
👍🏻 thanks

PR #956, opened 2020-08-22
Splits the renderer from the Gatsby 

### LINKED ISSUE #750: Module: Split gatsby-shiki-remark-shiki-twoslash into two modules
**Improvement Idea:**

I tried to make a markdown-it plugin for the module and it didn't work, I can separate out the remark specific bits and have a renderer be separate

orta:
Fixes #750 - Deprecates #956 - but should allow for building fourslash renderers for other tools like 11ty etc


orta:
Coool

PR #957, opened 2020-08-23
fix(playground): scroll to bottom does not work in Logs tab

86:
Scroll to bottom does not work in Logs tab of playground. 
It might be that the scroll target element has been changed without noticing.

I made the following changes
- Change scroll target from log-container to tabpanel
- Extract scroll to bottom logic as func and use it


ghost:
[CLA assistant check](https://cla.opensource.microsoft.com/microsoft/TypeScript-Website?pullRequest=957) <br/>All CLA requirements met.

orta:
Thanks 86, this looks good to me! It'll get deployed on monday

PR #958, opened 2020-08-23
Fix: Remove additional newline

davidbernegger:
Changing example 3 in accord to example 1 and 2 where the call to the printPoint function are after one new line and not two.


orta:
cool, consistency is a win

PR #967, opened 2020-08-24
Add declare keyword in module.d.ts.md, fixes #966

### LINKED ISSUE #966: Docs: following .d.ts default export documentation raises compiler error
**Page URL:** https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

**Issue:** Using `const` declarations at the top level of a `.d.ts` file raises a compile error. Such declarations should start with `declare`.

image

**Recommended Fix:** Include the `declare` keyword before the `const` variable declarations within two code blocks in the `Default Exports` section.

**Browser:** Edge

Zenulous:
Without the `declare` keyword in these code blocks, the TS compiler throws the following error: `Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier.` Fixes #966.

### DIFF
--- a/packages/documentation/copy/en/declaration-files/templates/module.d.ts.md
+++ b/packages/documentation/copy/en/declaration-files/templates/module.d.ts.md
@@ -58,7 +58,7 @@ module.exports = /hello( world)?/;
 Which can be described by the following .d.ts:
 
 ```ts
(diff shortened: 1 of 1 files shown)


orta:
Great, thanks!

PR #969, opened 2020-08-25
Two object of same shape are considered to be of same type

pragya91:


orta:
Hi, thanks!

I think I prefer the original version, can you explain your inspiration for making the change?

pragya91:
> Hi, thanks!
> 
> I think I prefer the original version, can you explain your inspiration for making the change?

Hi,  Thanks for reviewing the PR. 
Two objects are said to be same if they both have the same reference. Event though I understood what the documentation meant when I read it, I think it would still be a good idea to be more explicit.

orta:
Yeah, alright, let's do it 👍🏻

PR #971, opened 2020-08-25
fix: ensure strict checks are all recommended and related to

agilgur5:
## Description

- strictPropertyInitialization and alwaysStrict were missing from
  recommended even though the other Strict Checks were in there and all
  of them are enabled if strict is enabled

- alwaysStrict, strictNullChecks, noImplicitAny, and noImplicitThis were
  missing from strict's relatedTo even though all are enabled if strict
  is enabled

## Tags

Found while working on #970.
Related to another PR I had on the topic of Strict Checks, #500 

## Review Notes

They're all listed as "recommended", but `strict` enables all of them and is the only one turned on [in `@tsconfig/recommend`](https://github.com/tsconfig/bases/blob/4c3e845b83b8485ffbad371540598cbf0bfa5890/bases/recommended.json#L5) and by `tsc --init`, so maybe all of them should be removed from "recommended" except for `strict` instead? Thoughts?

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -28,9 +28,11 @@ export const internal: CompilerOptionName[] = ["preserveWatchOutput", "stripInte
 export const recommended: CompilerOptionName[] = [
   "strict",
   "forceConsistentCasingInFileNames",
+  "alwaysStrict",
   "strictNullChecks",
   "strictBindCallApply",
   "strictFunctionTypes",
+  "strictPropertyInitialization",
   "noImplicitThis",
   "noImplicitAny",
   "esModuleInterop",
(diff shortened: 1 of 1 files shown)


orta:
I'm in favor of this change, I agree that it could encourage redundant tsconfig entries ( which does make me wonder if we could make a page which helps you trim / understand your tsconfig visually #972  ) but I think it helps folks who aren't going to jump straight to strict

agilgur5:
> but I think it helps folks who aren't going to jump straight to strict

Doesn't the `relatedTo` help with that or the "Default `false`, unless strict enabled" handle that case?



> I'm in favor of this change, I agree that it could encourage redundant tsconfig entries

TSDX itself had that too until I fixed it in https://github.com/formium/tsdx/pull/475. And I didn't notice the templates had them too until a user pointed it out in https://github.com/formium/tsdx/issues/619 and I fixed in https://github.com/formium/tsdx/pull/673, https://github.com/formium/tsdx/pull/690. The draft PR for monorepo template has redundancies too https://github.com/formium/tsdx/pull/778#discussion_r457492085.. And I just noticed recently that we have `declaration: true` set redundantly too in the templates when it is the default

That is to say, it seems to be super common with TS users and a lot don't seem to know how all the options work (e.g. `paths` and `baseUrl` too). Though I'm sure you've seen that more than me. I only know after having read through the docs and now reference (which is a great addition!) several times as well as some issues in the TS repo.

> ( which does make me wonder if we could make a page which helps you trim / understand your tsconfig visually #972 )

Yea that could be pretty helpful if enough folks know of it and use it. I was thinking a `tsc --trim` command or something. Or other ways of recommending config -- I think the bases repo definitely helps as a reference point.

PR #985, opened 2020-08-29
Type assertion section improvement

ghost:
I propose to replace `let someValue: any ` with `let someValue: unknown`, because with `any` the type assertion on the next line is unnecessary `(someValue as string).length;` , in other words it can be replaced with `someValue.length` (without any error or warning). While with `unknown` the only way to proceed without an error is to use the type assertion.


orta:
Agreed, thanks!

PR #994, opened 2020-08-31
feat: translate Docs to ZH-Hans

LynnCheeses:
Update:   packages/documentation/copy/zh-hans/get-started/TS for the New Programmer.md
Update:   packages/documentation/copy/zh-hans/handbook-v1/The Handbook.md


orta:
Is there a difference between zh and zh-hans?

( I ask because all of the other docs are in zh, and it's likely this probably should )

LynnCheeses:
> Is there a difference between zh and zh-hans?
> 
> ( I ask because all of the other docs are in zh, and it's likely this probably should )

*zh-hans* means  simplified Chinese. There are also Traditional Chinese(zh-hant).
Simplified Chinese is officially used in mainland China, and is often also used by the Chinese community in Malaysia. Traditional Chinese is used in Taiwan, Hong Kong, and Macau.

orta:
I guess the question that matters is this: are you trying to add a _new_ language to the TypeScript website?

I _think_ we use zh for simplified Chinese here: https://www.typescriptlang.org/zh/ - which would mean `zh` can be used everywhere

PR #996, opened 2020-08-31
feat: add files in PT-BR

khaosdoctor:
Creating another PR from #922 since that one seems to have stalled. Already added the proposed changes

@danilofuchs @orta


danilofuchs:
LGTM

PR #924, opened 2020-08-17
feat: translate tscofing infos into portuguese/br 

developerfred:
#233
 * - [x] incremental.md
 * - [x] inlineSourceMap.md
 * - [x] inlineSources.md
 * - [x] isolatedModules.md
 * - [x] jsxFactory.md
 * - [x] keyofStringsOnly.md
 * - [x] lib.md
 * - [x] listEmittedFiles.md
 * - [x] listFiles.md
 * - [x] locale.md
 * - [x] mapRoot.md
 * - [x] maxNodeModuleJsDepth.md
 * - [x] module.md
 * - [x] newLine.md
 * - [x] noEmit.md
 * - [x] noEmitHelpers.md
 * - [x] noEmitOnError.md
 * - [x] noErrorTruncation.md
 * - [x] noFallthroughCasesInSwitch.md

PR #997, opened 2020-08-31
feat: translate tscofing infos into portuguese/br

khaosdoctor:
Creating another PR so we can move on with a set of files from #924 

Since there are a LOT of files in this PR, having it stalled was keeping a lot of good translations without being published.

@danilofuchs @orta


danilofuchs:
LGTM

PR #1008, opened 2020-09-01
Fix typo

sharils:
I've tested on tsc Version 3.9.7 and I can confirm that, the default is ES3 not ES5, because getter syntax is not supporter by default. And I have to set target to ES5 to support getter syntax. This also matches the document at https://www.typescriptlang.org/tsconfig#target where it says that ES3 is the default in the Allowed section.


orta:
Great! Thanks

PR #1010, opened 2020-09-02
Modify the number of built-in types from 7 to 8

chayeoi:
With the addition of the bigint type, the number of built-in types is now 8 not 7.


orta:
Thanks

PR #1013, opened 2020-09-02
Remove text that's sensitive to chapter order

sharils:
From this screenshot, we can see that we've already discussed union types when we are discussing about Enums. The best way to make the text not to depend on the order of chapters.

image


orta:
Agree! Thanks

PR #1020, opened 2020-09-03
Makes twoslash annotations work in the playground, and improve the bug workbench

orta:
Lots of folks wanting to try some advance stuff with WIP PRs, this allows it to work https://github.com/microsoft/TypeScript/pull/39560#issuecomment-686368195

This improves the speed and accuracy of real-time twoslash annotations, and allows them to be used in the playground 
Screen Shot 2020-09-03 at 11 07 09 AM


Kingwl:
Cool!

PR #1039, opened 2020-09-04
chore(doc): fix error in classes/accessors section

### LINKED ISSUE #1037: Docs: classes/accessors property not initialized
**Page URL:** https://www.typescriptlang.org/docs/handbook/classes.html#accessors

**Issue:**
When I launched the playground from the documentation and saw the error message **"Property '_fullName' has no initializer and is not definitely assigned in the constructor."** I thought "Oh, it doesn't work!". I just realized I can run the code and it works as described by the documentation. So, the issue new learners can be confused by the lack of property initialization.

**Recommended Fix:** Initialize the property.

**Browser:** _All I guess_

Capture d’écran 2020-09-04 à 11 42 50

masonlouchart:
Before this commit, when the playground to try class instance accessors
was launched the "errors" tab was saying the "_fulName" property was not
properly initialized. This commit fixes this issue.

fixes #1037

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Classes.md
+++ b/packages/documentation/copy/en/handbook-v1/Classes.md
@@ -373,7 +373,7 @@ To preserve existing functionality, we also add a simple getter that retrieves `
 const fullNameMaxLength = 10;
 
 class Employee {
-  private _fullName: string;
+  private _fullName: string = "";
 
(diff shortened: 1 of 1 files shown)


orta:
Thanks! Looks good to me, it's great to see little docs improvement PRs

PR #1041, opened 2020-09-04
Explain `keyof any` === `string | number | symbol`

sharils:
I was wondering what `keyof any` mean, then I found that the doc for [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) says `Record<K extends string | number | symbol, T>` and assumes it a typo.

Then I found that both https://github.com/microsoft/TypeScript/blob/master/src/lib/es5.d.ts#L1464-L1469 and https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1484-L1489 says it's `Record<K extends keyof any, T>`. 

So I was confused, until I found https://stackoverflow.com/questions/55535598/why-does-keyof-any-have-type-of-string-number-symbol-in-typescript that says they are equal.


orta:
Thanks for the detailed explanation 👍🏻

PR #1047, opened 2020-09-05
Fixing logic error in example code

freedoms33ker:
The point of the examples are described clearly, 
Though there is a minor logic error in the createSquare function. 

area: config.width || 20 should be area: config.width ? config.width*config.width : 20


orta:
thanks - great spot

PR #1050, opened 2020-09-06
fixing ts errors in class expressions example code

freedoms33ker:
When trying out the examples while reading the handbook, noticed some TS errors in the example code where types were not being defined in the interfaces, so TS was defaulting to the any type causing the example to break. Resolving these errors to assist future readers in playing with example code.


orta:
Nice spot - thanks!

PR #1051, opened 2020-09-07
Readonly properties are for the immutable props

mberneti:
Readonly properties are for the immutable properties and aren't for the modifiable properties.


orta:
Thanks for the PR, but I don't think this change is correct, the original statement re-worded is "you can only set `readonly` properties in the constructor"

mberneti:
@orta You are right.

PR #1057, opened 2020-09-09
[PT-BR] Classes 101.ts

dheysonalves:
Add PT-BR translations to:

**- Classes 101.ts file from Working with Classes**

@khaosdoctor @alvarocamillont @danilofuchs @orta


danilofuchs:
Nice one!
LGTM

PR #1063, opened 2020-09-10
Improve wording for @ts-expect-error usage

JulioC:
The original text reads 'prefixed', which might indicate that the comment should go in the suppressed line.

Changing the phrase to "when a line is preceded by a ..." might improve the understanding.


orta:
Thanks, yeah, I like this

PR #1073, opened 2020-09-11
Correct link to outFile option

datenreisender:
The 'F' in '#outFile' was wrongly not capitalised, breaking the link in most browsers.


orta:
Thanks!

PR #1078, opened 2020-09-12
Fix twoslash error caused by incorrect spacing

jsrath:
Under the 'Record' section of the [Utility Types page](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) there is a twoslash spacing error which results in the error `//      ^ = Could not get LSP result: bou>t<;` being shown:

ts-error

The error is caused by a few errant tabs before the caret.

Issue fixed by removing the tabs.


orta:
Ha, great spot! Thanks

PR #1083, opened 2020-09-13
Switch KeyboardEvent attribute from keyCode to key

koji:
In this code base, I saw that the code is using `key` instead of `keyCode`.
I guess unifying could be good.

Keycodes list
http://www.javascripter.net/faq/keycodes.htm

Did double-check on MDN
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

### DIFF
--- a/packages/playground/src/createElements.ts
+++ b/packages/playground/src/createElements.ts
@@ -142,16 +142,16 @@ export const createTabBar = () => {
   tabBar.addEventListener("keydown", e => {
     const tabs = document.querySelectorAll('.playground-plugin-tabview [role="tab"]')
     // Move right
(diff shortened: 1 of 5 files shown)


orta:
Nice, looks _close_! 

Just this to handle:

> `src/components/layout/Sidebar-keyboard.tsx(148,43): error TS2339: Property 'code' does not exist on type 'KeyboardEvent<Element>'.`

koji:
> Nice, looks _close_!
> 
> Just this to handle:
> 
> > `src/components/layout/Sidebar-keyboard.tsx(148,43): error TS2339: Property 'code' does not exist on type 'KeyboardEvent<Element>'.`

@orta, thank you for pointing out!
Seems that I forgot to remove a couple of `KebordEvent.code`.
Passed all checks!

orta:
Thanks!

PR #1088, opened 2020-09-14
Fix logic on Union and Intersection Types and position of popover on TypeScript with Web

richardspacheco:
This PR contains fixes for the following Playground examples:

- TypeScript with Web
- Union and Intersection Types

**Union and Intersection Types**

Currently, type CreateArtistBioRequest is either
`(CreateArtistBioBase & { html: string })`
or just
`{ markdown: string }`

I believe the expected type should be (as described on lines 72-73):

`CreateArtistBioBase`
and either 
`{ markdown: string }` or `{ html: string }`

This PR implements the expected behavior.

**TypeScript with Web**

Just changes the position of the popover to the left, which is the expected behavior according to lines 68-69:

```
// If you hit "Run" above, then a popup should appear
// in the bottom left, which you can close by clicking
```

The current position also renders the popover under the sidebar, hiding it.

Should I also make the changes to the localized versions of these examples?


orta:
Agreed - thanks!

PR #1093, opened 2020-09-16
typo: extraneous word in link to TSConfig Reference

agilgur5:
## Description

- seems like one was thinking "Check out the" and "There's also the",
  which came out as "There's also out the" 😄

## Tags

Discovered while reading through Compiler Options doc for discrepancies like those in #1092


orta:
Ha, thanks!

PR #1094, opened 2020-09-16
fix: alphabetize useDefineForClassFields in Compiler Options

agilgur5:
## Description

- the rest seem to be alphabetized, but this one was out of place

## Tags

Noticed this while checking for discrepancies between Compiler Options and TSConfig Reference after #1092


orta:
👍🏻

PR #163, opened 2019-12-26
V2 Playground

orta:
The start of the two libraries:
 
 - Sandbox - a wrapper around monaco optimized for working with TS, with the goal as being useful for anyone to embed on their own sites

 - Playground - the same thing as https://www.typescriptlang.org/play/ (eventually, as typeScript and not one massive file)

Screen Shot 2019-12-27 at 1 16 05 AM

Screen Shot 2019-12-27 at 1 16 05 AM

Screen Shot 2019-12-31 at 10 44 40 AM

Screen Shot 2020-01-02 at 2 30 23 PM

Screen Shot 2020-01-04 at 12 43 57 PM

- https://typescript-v2-163.ortam.now.sh/en/play/
- https://typescript-v2-163.ortam.now.sh/leet/play/
- https://typescript-v2-163.ortam.now.sh/sandbx-dev/

PR #1095, opened 2020-09-16
fix: remove extraneous declaration in relatedTo

agilgur5:
## Description

- it was specified twice, once grouped with types and typeRoots for some
  reason and missing declarationDir
  - so remove that one, leave in the one that has both declarationDir
    _and_ emitDeclarationOnly
 - second one that remains is ~5 lines below this: https://github.com/microsoft/TypeScript-Website/blob/2ab62adbfaf69e3e74c0a90dcdd7ea48fb21f17e/packages/tsconfig-reference/scripts/tsconfigRules.ts#L80

## Tags

Stumbled upon this after writing #971 

This seems to have been erroneously added in #163 [here](https://github.com/microsoft/TypeScript-Website/pull/163/files#diff-a912c6af3a16bf4288093c1264955bc6R77)

## Screenshot

Currently the first one supersedes the second one, so `declarationDir` isn't mentioned as related on the website:

Screen Shot 2020-09-16 at 12 04 44 AM

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -72,7 +72,6 @@ export const relatedTo: [AnOption, AnOption[]][] = [
 
   ["types", ["typeRoots"]],
   ["typeRoots", ["types"]],
-  ["declaration", ["emitDeclarationOnly"]],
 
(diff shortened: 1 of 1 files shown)


orta:
👍🏻 thanks

PR #1099, opened 2020-09-17
fix: add missing inverses in relatedTo

agilgur5:
## Description

- one side of the relationship existed but not the other side, ensure
  both sides of the relationship are relatedTo each other for proper
  backlinking etc

- add a new line after all the strict options for proper grouping same
  as the other groups

## Tags

Found while writing #1095 / #971 

## Review Notes

I omitted `allowJs` and `checkJs` from `emitDeclarationOnly`'s `relatedTo` because of #1098 , but can add them if that PR is rejected


orta:
Great, thanks!

PR #1100, opened 2020-09-17
fix: remove importHelpers' erroneous relatedTo self

agilgur5:
## Description

- it listed itself relatedTo which would link to itself when clicked

## Tags

Found while writing #1099 

## Screenshot

Here's how it looks before this change 😅 :

Screen Shot 2020-09-17 at 2 03 19 AM


orta:
Ha, good spot!

PR #1101, opened 2020-09-17
fix: allowUnreachableCode default should be undefined

agilgur5:
## Description 

- it was recently changed to undefined in the TSConfig Reference in
  ea8407c28fee7555446a869c095f2a4d2bab1ed4
  - this carries over that change to the Compiler Options docs as well

- and it being undefined seems to align with my past experience with
  unreachable code in TS, but have not tested directly recently

## Tags

Found while investigating discrepancies between the TSConfig Reference and Compiler Options after finding #1092 

This goes along with the changes to the Reference in #776


orta:
Agree - thanks!

PR #1102, opened 2020-09-17
fix: `jsx` is undefined by default, not preserve or react

agilgur5:
## Description

- one gets an error when using JSX without configuring `jsx`:
  `error TS17004: Cannot use JSX unless the '--jsx' flag is provided.`

- TSConfig Reference said default was `react`, which is inaccurate
- Compiler Options said default was `preserve`, which is also inaccurate

- Add a `undefined` as the default in TSConfig Reference, same as
  allowUnrechableCode
- Remove the default in Compiler Options, same as all the other options
  without defaults

#### Reference:
- `preserve` was written as the default in https://github.com/microsoft/TypeScript-Website/commit/eae5cc3ab38ddce1c0e70e36c223cb07bc734199#diff-444415c73770eafd0f9db7d708d74131R5  /  #103 [here](https://github.com/microsoft/TypeScript-Website/pull/103/files#diff-a912c6af3a16bf4288093c1264955bc6R54)
- But then was written as `react` in the body in https://github.com/microsoft/TypeScript-Website/commit/eae5cc3ab38ddce1c0e70e36c223cb07bc734199#diff-444415c73770eafd0f9db7d708d74131R5 / #119 [here](https://github.com/microsoft/TypeScript-Website/pull/119/files#diff-444415c73770eafd0f9db7d708d74131R5)
- It was then switched to `react` in https://github.com/microsoft/TypeScript-Website/commit/450ae964411f5180e4c3f7e6e974ccba03eb4e59#diff-a912c6af3a16bf4288093c1264955bc6R120 / #861 [here](https://github.com/microsoft/TypeScript-Website/pull/861/files#diff-a912c6af3a16bf4288093c1264955bc6R120)
- And then removed in https://github.com/microsoft/TypeScript-Website/commit/76b7d435fdbec8a4b88ac0b01b88e3fb72a829d1#diff-a912c6af3a16bf4288093c1264955bc6L120 / #864 [here](https://github.com/microsoft/TypeScript-Website/pull/864/files#diff-a912c6af3a16bf4288093c1264955bc6L120)

#### Compiler Options:
- `preserve` was written as the default in https://github.com/microsoft/TypeScript-Website/commit/66f17ad4d0fa6b2b9fee791b0b7a729710aeb907#diff-bfc78b96c15b0ee906c1ab6176546dceR41

## Tags

Similar to #1101 , found while investigating discrepancies between the TSConfig Reference and Compiler Options after finding #1092 

See historical references above as well

## Review Notes

Can remove the `undefined` default and just have nothing instead in the Reference. Wasn't sure if I should match `allowUnreachableCode` or if that were an exception since `undefined` means something else there.


orta:
I think wiping the string is 👍🏻  - this one also hit the CLA ( maybe it's on the size of the PR?) - good to go if you accept it

agilgur5:
CLA accepted. 🤷  didn't really see a pattern to why it chose this and #1111 to ask, sounds like a potential bug in the CLA checker 😬

orta:
Yeah, I'm not sure I want to look behind the curtain of the wizard of CLA - thanks, this is super useful

PR #1103, opened 2020-09-17
fix: resolve broken links in Compiler Options

agilgur5:
## Description

- basically all due to simple, 1 char mistakes
  - missing slash, extra dash, extra 'd'

## Tags

Found while writing #1102 

## Upcoming Merge Conflicts

This will conflict with #1102 since they change the same line

## Screenshot Diffs

Before:

Screen Shot 2020-09-17 at 1 17 07 AM

After:

Screen Shot 2020-09-17 at 1 17 40 AM

PR #1104, opened 2020-09-17
fix: resolve broken link to preserveConstEnums

agilgur5:
## Description

- spec.md was renamed to spec-ARCHIVED.md in the TypeScript repo, so
  link started 404'ing
  - as it's now archived, perhaps this should point somewhere else?
    e.g. the [const enums section in the Enums doc](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) in the Handbook or
    the [preserveConstEnums description in the TSConfig Reference](https://www.typescriptlang.org/tsconfig#preserveConstEnums)?
    - they both have different text than the spec, so unsure

## Tags

Found while writing #1103 

## Review Notes

Not sure if that's the best link per description, but unsure which it _should_ link to


packages/documentation/copy/en/project-config/Compiler Options.md
```diff
@@ -97,7 +97,7 @@ Option                                         | Type      | Default
...
 `paths`<sup>[2]</sup>                          | `Object`  |                                | List of path mapping entries for module names to locations relative to the `baseUrl`. See [Module Resolution documentation](/docs/handbook/module-resolution.html#path-mapping) for more details.
-`--preserveConstEnums`                         | `boolean` | `false`                        | Do not erase const enum declarations in generated code. See [const enums documentation](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#94-constant-enum-declarations) for more details.
+`--preserveConstEnums`                         | `boolean` | `false`                        | Do not erase const enum declarations in generated code. See [const enums documentation](https://github.com/microsoft/TypeScript/blob/master/doc/spec-ARCHIVED.md#94-constant-enum-declarations) for more details.
```
orta: There's a good section in the handbook on this:

```suggestion
`--preserveConstEnums`                         | `boolean` | `false`                        | Do not erase const enum declarations in generated code. See [const enums documentation](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) for more details.
```

orta (approved):

packages/documentation/copy/en/project-config/Compiler Options.md
agilgur5: ok, yea I mentioned that in my PR description but all the `const enums` docs are different so I wasn't sure which would be preferable

orta: Yeah, even this has a reasonably limited shelf life because we're re-writing the handbook for v2 - but at least it's not definitely 2+ years out of date

PR #1105, opened 2020-09-17
feat: add relatedTo for `jsx` in TSConfig Reference

agilgur5:
## Description

- jsxFactory and jsxFragmentFactory are relatedTo each other, so would
  think jsx itself should be as well

## Tags

Found that it was missing during #1102 and #1099 

## Review Notes

Feel free to close this if they shouldn't be `relatedTo`


orta:
Agree 👍🏻

PR #1110, opened 2020-09-17
fix: incremental is default true only if composite is

agilgur5:
## Description

- Compiler Options docs specify this behavior as such
  - which I assume means if composite is false, incremental is also
    false unless set explicitly, but I have never used it myself

## Tags

Similar to #1102 and #1101 , found while investigating discrepancies between the TSConfig Reference and Compiler Options after finding #1092 

## Review Notes

I've not actually used this option per the description, so I haven't tested to see which is accurate. Perhaps the TSConfig Reference was correct and the Compiler Options docs were wrong and so this alignment change should be inverted


orta:
I read the compiler to try and figure this out, it generally looks like this is true and I think the 3.4 release notes also back up this change - thanks

agilgur5:
Thanks for double-checking; yea my read on [the 3.4 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#composite-projects) is similar

PR #1111, opened 2020-09-17
fix: UMD should be specified in moduleResolution default

agilgur5:
## Description

- it's mentioned in the TSConfig Reference, but not in Compiler Options,
  so ensure the two align

- also use similar grammar in the TSConfig Reference as in Compiler
  Options as I thought this was easier to read this way

## Tags

Similar to #1110 , #1102, and #1101 , found while investigating discrepancies between the TSConfig Reference and Compiler Options after finding #1092 

## Review Notes

Similar to #1110, have not tested this myself to know which one is correct.
Also feel free to reject the Reference grammar change or let me know if you want the wording to be different in either.

Phew, this will be my last PR for a bit...


orta (approved):
I'm not sure what started triggering the CLA, but this is looks good to me 👍🏻

agilgur5:
Yea I was like "this isn't my first PR... 🤔  🤔 ". Signed now

orta:
👍🏻

PR #1119, opened 2020-09-21
Update Enums.md

awormasi:
I was trying to find information on the convention for singular vs plural enum naming. I couldn't tell if in this example `enum Directions` was intentionally plural.


orta:
Yeah, I think this is the right way 👍🏻

PR #1133, opened 2020-09-24
Fixed typeof type guard documentation

sodic:
The `typeof` operator works as a type guard with all the values it returns. The compiler successfully infers the types in all these functions:
```typescript
function f1(a: string | { [key: string]: string }): string {
  return typeof a === 'object' ? a['foo']: a;
}

function f2(a: string | (() => string)): string {
  return typeof a === 'function' ? a() : a;
}

function f3(a: string | null): string {
  return typeof a === 'object' ? String(a): a;
}


function f4(a: string | undefined): string {
  return typeof a === 'undefined' ? String(a): a;
}

function f5(a: string | bigint): string {
  return typeof a === 'bigint' ? a.toString(): a;
}
```

### DIFF
--- a/packages/documentation/copy/en/reference/Advanced Types.md
+++ b/packages/documentation/copy/en/reference/Advanced Types.md
@@ -158,7 +158,7 @@ function padLeft(value: string, padding: string | number) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
Great - thanks!

PR #1143, opened 2020-09-26
Fix gramatical error "then" to "than"

Davi-Danger:
"Then" declares order, and "than" declares comparison. Fixed the choice of word.


orta:
Thanks 👍🏻

PR #1169, opened 2020-09-30
Improve inline and google search results 

orta:


packages/typescriptlang-org/src/templates/tsconfigReference.tsx
```diff
@@ -74,7 +74,7 @@ const TSConfigReferenceTemplateComponent = (props) => {
 
   return (
     <Layout title={i("tsconfig_title")} description={i("tsconfig_description")} lang={props.pageContext.locale} allSitePage={props.data.allSitePage}>
-      <div className="tsconfig raised main-content-block">
+      <div className="tsconfig raised main-content-block markdown">
```
orta, line 77: [The algolia indexer looks for `.markdown hX`](https://github.com/algolia/docsearch-configs/blob/34df0a0c50a3395d589ff2df225f3677e09c0250/configs/typescriptlang.json#L42)

packages/typescriptlang-org/lib/bootup/ingestion/createPlaygroundExamplePages.ts
```diff
@@ -64,14 +60,11 @@ export const createPlaygroundExamplePages = async (
...
-    const id = postLangPath
-      .split("/")
-      .slice(-1)[0]
-      .split(".")[0]
+    const id = postLangPath.split("/").slice(-1)[0].split(".")[0]
 
     const { inlineTitle, compilerSettings } = getCompilerDetailsFromCode(code)
     createPage({
-      path: newPagePath,
+      path: newPagePath + ".html",
```
orta, line 67: Prev: `https://www.typescriptlang.org/ja/play/typescript/language-extensions/types-vs-interfaces.ts`
Post: `https://www.typescriptlang.org/ja/play/typescript/language-extensions/types-vs-interfaces.ts.html`

I think it's safe to assume that search engines don't try read `.ts` or `.js` files.

PR #1170, opened 2020-09-30
Improved docs

rahul-bhatt43:
Always being in support of humanity and creativeness
To make this society a better, faster and smarter place to live in.


orta:
Thanks, but this is bit off topic for this repo

PR #1180, opened 2020-10-01
Fix variable name in Advanced Types reference material

crd:
Fix variable name in Advanced Types reference material


orta:
👍🏻 thanks

PR #1182, opened 2020-10-01
fix: Typo: 'synonomous' should be 'synonymous'

bennorth:
I'm using the default (`v2`) branch as the destination.  Hope that's correct.  If not, I can adjust.  Thanks!


orta:
Thanks 👍🏻

PR #1183, opened 2020-10-01
Update Modules.md

RevealedFrom:
According to https://nodejs.org/api/esm.html#esm_differences_between_es_modules_and_commonjs, filename extension is required for ES modules.


orta:
This is a commonjs example, which doesn't need it.

PR #1184, opened 2020-10-01
Children should be typedchecked as React.ReactNode

espipj:
It is recomended to cover all the edge cases.
Reference: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L825

I'm happy to create a new section stating how react props with children should be typed by default. But I think in the rest of the examples it should be encouraged to use ReactNode 😃


RyanCavanaugh:
This phrasing is intentional; the Handbook doesn't use React-specific types here.

orta:
Cool, thanks for the PR @espipj  but I agree with Ryan here - in this context we're just talking JSX, not React specific parts of it.

PR #1187, opened 2020-10-02
Update Modules.md

RevealedFrom:
According to https://nodejs.org/api/esm.html#esm_differences_between_es_modules_and_commonjs, filename extension is required for ES modules.


orta:
See https://github.com/microsoft/TypeScript-Website/pull/1183

RevealedFrom:
@orta, sorry I made a mistake earlier and modified the wrong line in #1183. This is line 361, an ES2015 example.

PR #1188, opened 2020-10-02
Update Errors.ts

prmlimajr:
Translation to brazilian portuguese.


RyanCavanaugh (changes requested):
Thank you for the translation!

You've translated the English copy in-place, but the translation should be placed under the appropriate locale folder instead. https://github.com/microsoft/TypeScript-Website/pull/1154 is a good example of how to do this correctly.

khaosdoctor:
@prmlimajr Can you copy this same file you edited to the PT folder following #233 directives? 

Thanks a lot!

PR #1192, opened 2020-10-02
remove search box outline

johnpapa:
remove the rounded border outline around the search box when it has focus

image

to

image

cc @orta

### DIFF
--- a/packages/typescriptlang-org/src/components/search/SearchArea.scss
+++ b/packages/typescriptlang-org/src/components/search/SearchArea.scss
@@ -22,6 +22,10 @@
   padding-bottom: 0.5rem;
 }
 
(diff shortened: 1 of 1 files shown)


jefrydco:
Imho the rounded border outline is good for a11y, it has uniform styling across clickable thing in ts website

TypeScript Website Search A11Y

orta:
My first reaction was the same as @jefrydco's but I wanted to at least try it locally before solidifying my opinion. After playing with it, I prefer this PR - I think it's pretty obvious that you have it selected given the large colour change and this feels much nicer for mouse users

image


Thanks @johnpapa 👍🏻

johnpapa:
thanks

Issue #1193, opened 2020-10-02
Docs: README.md invalid markdown for links in Docs issue


sharechiwai:
The links in README.md for the Docs section is now shown as hyperlink, this is cause by the document has spaces, which markdown could not treat it as hyperlink

![image](https://user-images.githubusercontent.com/843624/94932226-5daf3d00-04fb-11eb-9843-d77cb23b4187.png)

**Page URL:** https://github.com/microsoft/TypeScript-Website

**Issue:** README.md Docs section, hyperlink does not work 

**Recommended Fix:** fix the spacing issue on the link or avoid having space in filename.

**Browser:** Google Chrome, Edge

soub4i:
I will take this

sharechiwai:
> I will take this

I have created a pull-request already https://github.com/microsoft/TypeScript-Website/pull/1194

sharechiwai:
as #1194 has been merged and links is shown correctly, I will close this issue
![image](https://user-images.githubusercontent.com/843624/95095103-3788d780-075d-11eb-9e3a-9f2650601c00.png)

PR #1194, opened 2020-10-02
fixed links issue in Docs section, which cause by the target document filename contain spaces

sharechiwai:
This fix is relative with #1193


orta:
Great spot - thanks!

PR #1209, opened 2020-10-05
add noop catch for duplicate module paths

### LINKED ISSUE #1208: Playground: Fails to load node types
When trying to use `http` module types i noticed things were failing in the ATA.

Screen Shot 2020-10-05 at 3 37 53 PM

**Repro:**
https://www.typescriptlang.org/play?jsx=0&ts=4.1.0-dev.20201005#code/JYWwDg9gTgLgBACxjMcBmUIjgciSnAKFElkWTACZ1Ns8LKiTp58wBnGrXN9nIA

Go here and try to get http types to work. Open console for error in screenshot

Ethan-Arrowood:
fixes #1208 

catching and noop seems to work fine for this error; not sure if there is a way to check for an existing model; will spend a little more time combing through monaco api but this solution works for now

### DIFF
--- a/packages/sandbox/src/index.ts
+++ b/packages/sandbox/src/index.ts
@@ -160,12 +160,9 @@ export const createTypeScriptSandbox = (
   // In the future it'd be good to add support for an 'add many files'
   const addLibraryToRuntime = (code: string, path: string) => {
     defaults.addExtraLib(code, path)
-    try {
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood:
Found a `getModel` check we can safely use. Pushing commit now for that instead

orta:
👍🏻 great!

PR #1211, opened 2020-10-06
fix: setupTwoslashHovers doesn't work with Gatsby HMR

Jack-Works:


orta:
:wave: - very cool, this looks right to me!

orta:
I'll make a version bump for the module

PR #1213, opened 2020-10-06
Subnav scrolling issue got solved

### LINKED ISSUE #949: Nav can get extra scrollbar
Now that I'm using a scroll wheel

Screen Shot 2020-08-20 at 6 16 08 PM

we should never allow a scroll bar inside the subnav

SMAKSS:
Fixes #949
There was a rule in `sidebar.scss` which cause all `nav#sidebar`, `ul` element `max-height` to be `calc(100vh - 10px)`. This rule changed and it is only applied to direct `ul` elements to prevent scrolling in `subnav` menus.


orta:
Ace, looks good to me - thanks!

PR #1237, opened 2020-10-13
Playground multi file

orta:
This has a looooong way to go, but the biggest blocker, the file browser UI is now wrapped.


ggoodman:
This is amazing! 🎉

orta:
I don't think this approach can work in the long run, the monaco tree API has already changed once or twice since I made this PR and I would have to maintain shims for each and every version inside the playground and do runtime checks to see how I can open it. 

I think I'll just have to make my own file tree.

PR #1248, opened 2020-10-16
Throws with a useful error message if you try access a lib file which isn't in the fileMap

orta:
I hit this, Nathan hit this, anyone can hit this.

### DIFF
--- a/packages/typescript-vfs/src/index.ts
+++ b/packages/typescript-vfs/src/index.ts
@@ -441,7 +441,7 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
         if (!result) {
           const libs = nodeSys.readDirectory(tsLib)
           throw new Error(
-            `TSVFS: A request was made for ${tsLibName} but there wasn't a file found in the file map. You likely have a mismatching in the compiler options for the CDN download vs the compiler program. Existing Libs: ${libs}.`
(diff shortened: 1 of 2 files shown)


packages/typescript-vfs/src/index.ts
```diff
@@ -437,7 +437,14 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
       if (files.has(fileName)) return files.get(fileName)
       if (fileName.startsWith("/lib")) {
         const tsLibName = `${tsLib}/${fileName.replace("/", "")}`
-        return nodeSys.readFile(tsLibName)
+        const result = nodeSys.readFile(tsLibName)
+        if (!result) {
+          const libs = nodeSys.readDirectory(tsLib)
+          throw new Error(
+            `TSVFS: A request was made for ${tsLibName} but there wasn't a file found in the file map. You likely have a mismatching in the compiler options for the CDN download vs the compiler program. Existing Libs: ${libs}.`
```
sandersn:
```suggestion
            `TSVFS: A request was made for ${tsLibName} but there wasn't a file found in the file map. You likely have a mismatch in the compiler options for the CDN download vs the compiler program. Existing Libs: ${libs}.`
```

sandersn (approved):
Would also be nice to throw in the general case, with a less useful, but still indicative, error message.

orta:
Do we always `fileExists` before reading? That used to be considered an anti-pattern in Mac/iOS native code, but if we do it consistently in tsc then I'm not against a general casing

sandersn:
Never mind, I misread the code -- I didn't see the initial `.has` check or the fallback to `nodeSys`.

PR #1261, opened 2020-10-20
Add return type void documentation

Ethan-Arrowood:
Wasn't sure if this deserved to be in the `void` part of this document. It more has to do with function assignability but I'll see what maintainers have to say about it. In addition, wasn't sure the best way to link to other things so let me know how to change that if necessary.

Rel: https://github.com/microsoft/TypeScript/issues/41138 (already closed by bot)


orta:
Yeah, this is great, and a perfect starter for the assignability section. Solid work 👍🏻

PR #1274, opened 2020-10-23
Handle auto playgrounds

orta:
I've got a system set up for automatically generating playground versions when there's a new TS release on npm, this means we should keep the playground selector clean by only showing the latest patch for every min/maj.


packages/sandbox/src/releases.ts
```diff
@@ -15,7 +15,22 @@ export const supportedReleases = [
   "2.8.1",
   "2.7.2",
   "2.4.1",
-] as const;
+] as const
+
+export const supportedReleases = [
+  "4.0.2",
+  "3.9.7",
```
orta, line 23: In above `allReleases`: 

```
  "3.9.7",
   "3.9.2",
```

packages/sandbox/src/releases.ts
```diff
@@ -17,6 +17,23 @@ export const supportedReleases = [
   "2.4.1",
 ] as const;
 
+/** The latest major.min version **/
+export const supportedReleases = [
+  "4.0.2",
```
Gerrit0, line 22: Should this be 4.0.3 now? There's a couple bugfixes that would be nice to have.

orta: I've triggered the deploy for that, should appear on the next prod deploy and automatically get added

Screen Shot 2020-10-23 at 11 56 22 AM

PR #1275, opened 2020-10-23
TS 4.1 & Gatsby & Yarn update

orta:
Figured I may as well bite  the bullet and just do a big dep update all over the show.

PR #1278, opened 2020-10-25
MDX

orta:
Contains #1275


orta:
This is close, just need to figure out:

 - How to get markdown plugins working 
 - What it looks like in the epubs

orta:
This could be unlocked by the new twoslash remark plugin

orta:
I think I'm going to drop the idea of mdx. I can't think of a safe way to do localization with it.

PR #1309, opened 2020-10-31
Update TS for Functional Programmers.md

molaux:
As the comment suggests, `snd` should be just a more precise declaration of the preceding statement `fst`, isn't it ?

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
@@ -87,9 +87,11 @@ Notes:
 1. Function syntax includes parameter names. This is pretty hard to get used to!
 
    ```ts
-   let fst: (a: any, d: any) => any = (a, d) => a;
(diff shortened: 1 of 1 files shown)


orta:
I agree /cc @sandersn just to triple check

sandersn (changes requested):
The code is written as intended.

However, maybe my intention was confusing. My assumption was that `fst` and `snd` are such a linked pair that I could use them as a single example: showing the imprecise type for `fst` and the precise one for `snd` would flow pretty much the same as rewriting the type of `fst` to be more precise.

Maybe it would be clearer to rewrite `fst`. The example will produce an error then, but I'm not sure we're running the compiler on these documents anyway. Are we, @orta?

Edit: Oops, I meant to say: rewrite `snd` &rarr; `fst`.

orta:
We can!

DanielRosenwasser:
"more precisely" implies you're not going to switch to a different function. The change + switching to `fst` sounds good.

PR #1312, opened 2020-11-01
fix: playground actions `clear playground logs`

abdallamourad:
When Run Script `Clear Playground Logs` is appended to actions every single time. they are all doing the same action.

image

**Repro:**
- GO TO [Playground](https://www.typescriptlang.org/play?#code/DYUwLgBAlgzgIgewHYgFwQEYIaAhkiAXggDNdgYQBuIA) 
- Click Run multiple times
- Check actions in editor 
- `Clear background Logs` action is repeated
```ts
// A *self-contained* demonstration of the problem follows...
// Test this by running `tsc` on the command-line, rather than through another build tool such as Gulp, Webpack, etc.
```

**Expected behavior:**
`Clear Background Logs` is appended to actions list one time only
**Actual behavior:**
`Clear background Logs` action is repeated as much as the script runs

### DIFF
--- a/packages/playground/src/sidebar/runtime.ts
+++ b/packages/playground/src/sidebar/runtime.ts
@@ -14,6 +14,7 @@ export const runPlugin: PluginFactory = (i, utils) => {
       if (!addedClearAction) {
         const ui = createUI()
         addClearAction(sandbox, ui, i)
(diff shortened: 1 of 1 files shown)


orta:
Nice spot! Thanks!

PR #1314, opened 2020-11-02
fix: use correct import syntax in esModuleInterop docs

VytautasT:
It seems that an incorrect import syntax is used in the documentation of `esModuleInterop` config option.

This PR fixes all markdown files having this mistake.

The [Japanese](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/tsconfig-reference/copy/ja/options/esModuleInterop.md) and the [Portugese](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/tsconfig-reference/copy/pt/options/esModuleInterop.md) files seem to follow a different template, so no changes there.


SMAKSS (approved):

orta:
great spot -thanks

PR #1315, opened 2020-11-02
Add single quote example

fmachado091:
This change uses 'red' instead of "red" in the color assign, so as to provide examples of both possibilities, as stated in the previous text:

> TypeScript also uses double quotes (`"`) or single quotes (`'`) to surround string data

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Basic Types.md
+++ b/packages/documentation/copy/en/handbook-v1/Basic Types.md
@@ -39,7 +39,7 @@ Just like JavaScript, TypeScript also uses double quotes (`"`) or single quotes 
 
 ```ts twoslash
 let color: string = "blue";
-color = "red";
+color = 'red';
 ```
 
 You can also use _template strings_, which can span multiple lines and have embedded expressions.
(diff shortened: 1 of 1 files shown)


SMAKSS (commented):
Thank you for the PR, but IMO, I think this will lead to confusion because it is against the general code convention to use different quotes to represent a String data type and assign it to a value. If we tend to show the different usage/capability of quotes in the Typescript I think it is better to show them in two different examples, but since the sentence is clear and the use case is pretty obvious I don't see any particular reason to show the different quote usage.

packages/documentation/copy/en/handbook-v1/Basic Types.md
```diff
@@ -39,7 +39,7 @@ Just like JavaScript, TypeScript also uses double quotes (`"`) or single quotes
 
 ```ts twoslash
 let color: string = "blue";
-color = "red";
+color = 'red';
```
orta, line 43:
```suggestion
// prettier-ignore
color = 'red';
```

orta:
Yeah, IMO we should do this (it was [the original intent](https://github.com/microsoft/TypeScript-Handbook/blob/master/pages/Basic%20Types.md#string)) - but right now this would get un-done by prettier which runs across every code sample. I think the change I just pushed should be enough to persist the change.

PR #1317, opened 2020-11-03
Type Guards: how to filter an array

simon04:


SMAKSS (changes requested):
Well, the CI is failing because the snippet that you include constantly redeclaring the same variable as a `const` and using undefined variables to declare the others, so in order to make it pass you need stick with the basic rules that provided [here](https://github.com/microsoft/TypeScript-Website/blob/v2/docs/Converting%20Twoslash%20Code%20Samples.md).

packages/documentation/copy/en/reference/Advanced Types.md
```diff
@@ -99,6 +99,27 @@ if (isFish(pet)) {
...
+declare function getSmallPet(): Fish | Bird;
+function isFish(pet: Fish | Bird): pet is Fish {
+  return (pet as Fish).swim !== undefined;
+}
+// ---cut---
+const zoo: (Fish | Bird)[] = [getSmallPet(), getSmallPet(), getSmallPet()];
+const underWater1: Fish[] = zoo.filter(isFish);
+// or, equivalently
+const underWater2: Fish[] = zoo.filter<Fish>(isFish);
+// error "Signature '(pet: Fish | Bird): boolean' must be a type predicate"
```
orta:
```suggestion

```

The true error message from the compiler is inlined in the HTML 👍🏻

PR #1318, opened 2020-11-04
changed error message with type

decpk:
Changed error message from value to type.
**from**
ts-error
**to**
ts-error-correction


orta:
Sure, I guess why not 👍🏻

PR #1320, opened 2020-11-05
CLI Reference

orta:
Now separates CLI commands like "--build" and "--help" etc from Compiler Flags

Screen Shot 2021-03-24 at 12 10 00 PM

Fixes a design problem with small screens, now description shows under the main row
Screen Shot 2021-03-24 at 12 09 33 PM


orta:
I woke up this morning and figured out the design change needed to make this work, these tables are now much more manageable on smaller screens

PR #1326, opened 2020-11-08
Update module-class.d.ts.md

alper:
Do these changes make sense? It seems like the type file does not correspond to the javascript at the top.


packages/documentation/copy/en/declaration-files/templates/module-class.d.ts.md
```diff
@@ -50,15 +50,15 @@ To handle both importing via UMD and modules:
  *~ loaded outside a module loader environment, declare that global here.
  *~ Otherwise, delete this declaration.
  */
-export as namespace myClassLib;
+export as namespace super-greeter;
```
orta:
```suggestion
export as namespace "super-greeter";
```

orta:
Yep, I agree, this does make sense! Sorry, missed this PR

PR #1342, opened 2020-11-14
Add backticks around readonly in Interfaces.md

fmachado091:
Throughout the whole document, `readonly` is always styled as an inline code, except in this occurrence.
This change adds the backticks around it so it follows the same pattern.


SMAKSS (commented):
Hi there, thanks for the PR, I guess we only put the readonly within a backtick whenever we intend to refer to the `readonly` keyword and I think this sentence is not the case and it is only referring to the read-only definition.

orta:
I think this is probably the right call (it does refer to `readonly` syntax in the code sample above) thanks folks!

PR #1346, opened 2020-11-17
Update @types install instructions

koddsson:
Declaration files from DefinitelyTyped should be installed as a dev dependency.


SMAKSS (approved):

orta:
Agree!

PR #1357, opened 2020-11-20
feat(playground): support `displayName` for plugins

antfu:
image

The package naming converting can't possibly handle all the situations, instead of handling more edge cases, maybe it's better to allow customization. This PR adding a feature to the fetching script, allowing to use `displayName` if it presents.

Similar to VS Code extension manifest: https://code.visualstudio.com/api/references/extension-manifest


orta:
Agree! Thanks @antfu !

PR #1368, opened 2020-11-25
Docs: Correct "Visual Studio Code" typo

Ycmelon:
visual studio and vs code are different editors, from the previous sentence I think we mean vs code instead?


orta:
Agree, thanks

PR #1374, opened 2020-11-26
twoslash: fix list<Map> TS options

d-fischer:
Most option types are parsed in `setOption`, even `Map` options are accounted for, but lists of `Map` are not.

This PR fixes the mentioned problem, which most notably prevents `// @lib:` from working correctly.


orta:
Awesome! Yeah, this looks 💯 

It should automatically deploy when I merge

PR #1375, opened 2020-11-27
Fix generated allPages to use correct paths on Windows

Zlatkovsky:
On a Windows machine, running `yarn start` fails with the following error:

[SITE]
[SITE] File: src\__generated__\allPages.ts:4:7724
[SITE]
[SITE] failed Building development bundle - 14.346s

It appears this is because of Windows using backlash for certain files, leading to incorrect JS

image

This fix ensures that backlashes are converted to forward-slashes, and also changes the generated output to have each entry on its own line rather than all in one line, just for being able to see the output better.


orta:
Thank you! 

I have setup + tests running in windows, maybe I should expand the list of things which run in there to try and nt break things in the future.

Issue #367, opened 2020-03-17
Playground:'Log' on the right side of page miss delayed message when using 'Promise'


asurance:
**Code**
```ts
main()
async function main() {
    console.log(1)
    Promise.resolve().then(() => {
        console.log(2)
    })
    console.log(3)
}

async function main2() {
    console.log(1)
    new Promise((resolve) => {
        console.log(2)
        resolve()
    })
    console.log(3)
}
```
**Logs**
```
1
3
```
We need to open browser debugging tools to get actual result.
```
1
3
2
```
It can be confusing or misleading for people who are not familiar with the 'Promise'

orta:
Makes sense, you're welcome to take a look at this here: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/playground/src/sidebar/runtime.ts

asurance:
> Makes sense, you're welcome to take a look at this here: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/playground/src/sidebar/runtime.ts.  

OK,I know why the problem will happen.Because the delay calls `console.log`, if you just change `console.log`, you really can't output the delayed message。I have a bold idea, but I'm not sure it's feasible and that is to use `namespace` or something like that.
```typescript
const another = console
//random name guarantee no conflicts
namespace a {
    // set context for this use
    const console = {
        log(...argv: any[]) {
            another.log(`under namespace a`)
            another.log(...argv)
        }
    }
    // user‘s first use
    console.log(1)
    Promise.resolve().then(() => {
        console.log(2)
    });
    // end
}
namespace b {
    const console = {
        log(...argv: any[]) {
            another.log(`under namespace b`)
            another.log(...argv)
        }
    }
    // user's second use
    console.log(3)
    // end
}
```
And the output:
```
under namespace a
1
under namespace b
3
under namespace a
2
```
So you can distinguish between the first and the second.

Zlatkovsky:
@asurance , @orta , I just submitted a pull request to fix this issue, and verified that @asurance 's example now works correctly.

Zlatkovsky:
Actually, looks like my fix is incomplete. See more discussion [here](https://github.com/microsoft/TypeScript-Website/pull/1376#issuecomment-734919002)

PR #1376, opened 2020-11-27
Await evaluation of JS code to ensure logs and error-handling shows up

Zlatkovsky:
# Problem statement

As reported by me on [this issue](https://github.com/microsoft/TypeScript/issues/41695), and as noticed & reported by [someone else earlier this year](https://github.com/microsoft/TypeScript-Website/issues/367), the Console on the TypeScript playground only works for synchronous code. If you have code like:

```
console.log("start");
pause(100).then(() => console.log("end"))

function pause(ms: number) {
  return new Promise(resolve =>
    setTimeout(resolve, ms));
}
```

only the word "start" will appears with each run, but never "end". However, the real Dev Tools console does show both being logged:

Works correctly in Dev Tools console

# The fix

The evaluation of user code is done today via

```
try {
      eval(js)
} catch (...) { ... }
```

With my fix, the evaluation of the JS code is `await`-ed (whether or not it's synchronous; if it's synchronous, then the `await` is just a no-op). This ensures both that the `console` variable will remain within the scope of the evaluation, and that any unhandled exceptions would get correctly caught by the `catch` statement.

# Verification
Verified that the TS playground continues to work correctly, and that async code gets logged and error-handled correctly.

### DIFF
--- a/packages/playground/src/sidebar/runtime.ts
+++ b/packages/playground/src/sidebar/runtime.ts
@@ -83,9 +83,11 @@ function rewireLoggingToElement(
   console["oldclear"] = console.clear
   console.clear = clearLogs
 
-  closure.then(js => {
+  closure.then(async js => {
     try {
-      eval(js)
+      // The produced JS might well have asynchronous code in it, so await it.
+      // (Having the await fixes https://github.com/microsoft/TypeScript-Website/issues/367)
+      await eval(js)
     } catch (error) {
       console.error(i("play_run_js_fail"))
       console.error(error)


orta:
Ha, that's a very elegant answer. 

I did not know you could do this, kudos.

Zlatkovsky:
@orta , thanks for merging in my change! That said, I just realized that my fix isn't complete -- it only works if the very last thing being evaluated is a Promise. So it works for something like:

```
console.log("start");
pause(100).then(() => console.log("end"));

function pause(ms: number) {
  return new Promise(resolve =>
    setTimeout(resolve, ms));
}
```

But it does NOT work if you add another statement after the call to `pause`. E.g.,:

```
console.log("start");
pause(100).then(() => console.log("end"));
console.log("something else"); // <== THIS LINE MAKES "end" NOT SHOW UP!

function pause(ms: number) {
  return new Promise(resolve =>
    setTimeout(resolve, ms));
}
```

So I think that's the limit of using the `eval` approach -- `eval` only returns the last statement and so that's the only thing that gets `await`-ed.

I think there are two ways of solving this more thoroughly:
1. Re-wire the actual "window.console" objects to go to the console UI. The biggest issue I see with this is that any other errors that are from the page itself, that have nothing to do with the user's code, would also get logged there... which doesn't feel great.
2. Do the above approach, *but within the context of a sourceless iframe*. This way, only the iframe's own code would log to the console. The disadvantage is that spinning up an iframe -- even a sourceless one -- is slower than just `eval`, and it would be a somewhat more involved change (but I'm happy to do it over the holidays, I have some experience in that space).

What are your thoughts on this? Any other TypeScript website stakeholders want to chime in?

orta:
Yeah, I considered the first and decided against it, it became unwieldy for dev work (e.g. playground plugin dev) and there's legit reasons for the playground to use the console. I also considered the iframe, but I like have examples which include adding to the current DOM which makes an iframe a bit of a bad solution too.

Another alternative is to swap out any references to the identifier `console` in the source code before `eval` with a custom global object like `playConsole` instead - which would always print to the sidebar?

Zlatkovsky:
@orta, could you elaborate on the "example including adding to the current DOM"? I.e., where exactly would that DOM show up, and would it be manipulating the site's own DOM?...

I'm all in support of having user-code be able to impact the DOM, but it feels like it'd be cleaner if it were sandboxed to a portion of the DOM that the user is in control over, rather than the overall site (whose design/layout and CSS could change over time).

So if you want to support the DOM scenario, what if the iframe was actually a visible entity, and showed up on the side bar alongside the tabs for Logs, Errors, etc? In fact, maybe it would even be possible to have it show up *only when the iframe DOM is being manipulated into something interesting*, e.g., attaching a MutationObserver to see if anything changes within the iframe's DOM, and showing up the tab if so. I think such solution would both light up a cleaner version of the DOM-manipulation scenario, and solve the logging-from-only-within-the-iframe scenario.

Regarding the alternative of "swap[ing] out any references to the identifier `console` in the source code before `eval` with a custom global object like `playConsole` instead": yeah, that's an interesting idea, I think it would work. Though if going with the iframe approach, it would be a moot point (and vice-versa) -- so I guess it'd be good to determine for sure the direction for solving this.

orta:
Sorry, been moving house - @asurance just shipped https://github.com/microsoft/TypeScript-Website/pull/1390 which I think is a great answer to this 👍🏻

PR #1390, opened 2020-12-02
fix(playground): Bug #367

### LINKED ISSUE #367: Playground:'Log' on the right side of page miss delayed message when using 'Promise'

asurance:
Fixes #367 
Fixes #1376 
# Problem statement
The previous fix will still have problems in the following code:
```typescript
console.log("start");
pause(1000).then(() => {
    console.log("end")
});
console.log("something else"); // <== THIS LINE MAKES "end" NOT SHOW UP!

function pause(ms: number) {
    return new Promise(resolve =>
        setTimeout(resolve, ms));
}
```
# The fix
The key is to create a new `console` in the `try catch` scope instead of modifying the original one
# After fix
show

### DIFF
--- a/packages/playground/src/sidebar/runtime.ts
+++ b/packages/playground/src/sidebar/runtime.ts
@@ -3,7 +3,9 @@ import { PlaygroundPlugin, PluginFactory } from ".."
 import { createUI, UI } from "../createUI"
 import { localize } from "../localizeWithFallback"
 
-let allLogs = ""
+const allLogs: string[] = []
+let offset = 0
+let curLog = 0
 let addedClearAction = false
 
 export const runPlugin: PluginFactory = (i, utils) => {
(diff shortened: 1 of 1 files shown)


orta:
This is really exceptional work @asurance - thanks a lot!

Having the re-assignment per call is a 10/10 answer

PR #1416, opened 2020-12-11
Adds localization casing etc to the CI

orta:
Looks like reporting is a little off in localized issues (and probably in the site too by proxy) because we never actually validate that the folder structure in a localized folder matches the English version. Now that happens on CI.


khaosdoctor:
This CI appears to be getting a lot of typos in PT, is this correct?

orta:
Yeah, the spellechecker only runs for me and I normally only edit English 👍🏻

khaosdoctor:
I think that's cool! lgtm

khaosdoctor (approved):

PR #1423, opened 2020-12-14
Fix Enum reverse mapping link

tamagosante:
I also checked the other links (3 in total), but the other 2 are as expected

image


orta:
Thanks!

PR #1433, opened 2020-12-17
Add default return value from getOwnPropertyDescriptor to remedy type error

hatched:
Using TS 4.1.2 this example code had a type error. This adds a default return value if `getOwnPropertyDescriptor` returns `undefined` which satisfies the type requirement.

Screen Shot 2020-12-17 at 10 30 29 AM


orta:
👍🏻 thanks

PR #1438, opened 2020-12-22
Remove "boom!" from documentation

laurent22:
I know that's a trend at the moment, but I'm convinced we can write good technical documentation without adding "boom!" all over the place.


orta:
Thanks for the PR, but I think it's fine to have a hint of personality in the release notes. 👍🏻

laurent22:
Yes, about as much personality as a teenager trying to sound cool in front of his friends, but ok 👍

PR #1446, opened 2020-12-26
Update print to log in docs

lalit2005:
This is my **[second](https://github.com/firstcontributions/first-contributions/pull/35578)** pull request, so I don't know if this is the correct way of opening a pull request. 

"print" is from python, so other beginners like me can think the same (Suddenly I thought that typescript is related to python).
Hence I changed "print" to "log" [here](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html) which seems to be more relevant to javascript(and typescript).

Hope my second contribution will be to MICROSOFT😲

[Lalit](https://github.com/Lalit2005)

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for JS Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for JS Programmers.md
@@ -219,13 +219,13 @@ interface Point {
   y: number;
 }
 
-function printPoint(p: Point) {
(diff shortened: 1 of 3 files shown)


orta:
Well, hello there Lalit, glad to be your 2nd pull request!

I agree with your changes, thanks for your contributions!

PR #1447, opened 2020-12-27
Document that index signatures do not include undefined

vobarian:
This has been a source of confusion and much debate. I do not propose to settle the debate but surely it is better to document the current behavior. See https://github.com/microsoft/TypeScript/issues/13778 and https://github.com/microsoft/TypeScript/issues/9235

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Interfaces.md
+++ b/packages/documentation/copy/en/handbook-v1/Interfaces.md
@@ -434,6 +434,29 @@ interface NumberOrStringDictionary {
 }
 ```
 
+Beware, the compiler assumes indexes and properties return the type declared but it cannot actually verify that specific index or
+property names exist. Reading a nonexistent index or property returns `undefined` but is not checked by the type system.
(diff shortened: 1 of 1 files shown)


orta:
I've been punting on reviewing this a few times, I think there's some value in here but I think the tone is off. I'll see if I can get some time to re-write this next week.

vobarian:
Any news on this?

orta:
We've been talking about building a migrating to "strict" doc, and this is probably going to live in that

RyanCavanaugh:
Thanks for taking the time to submit this PR!

While we appreciate the positive edits you've made, we have noticed a number of typographical, grammatical, or technical errors that would need to be addressed before this could be merged. In order to ensure that we can efficiently work through our PR backlog, we have decided to close those PRs which require significant copyediting before they can be merged. Specifically, the phrasing here ("Beware!", linter, etc) is inconsistent in a number of places with the general tone and terminology used throughout the rest of the document.

PR #1450, opened 2020-12-29
Update Utility Types.md

pretentious7:
I found the previous documentation for the record type to be unclear.
This new one is taken from jcalz's answer here : https://stackoverflow.com/questions/51936369/what-is-the-record-type-in-typescript

(is that ok, btw?)


orta:
Yeah, this seems reasonable to me 👍🏻

PR #1457, opened 2021-01-02
Update compilerOptions.ts

YousefED:
config.compilerOptions are not passed to the eventual compileroptions. Only `useJavaScript` is used from the passed in compilerOptions.

The only way for me to set compileroptions from consuming code was via `updateCompilerSetting`, but not via the initial config object. This should fix that


orta:
This makes sense, thanks!

PR #1464, opened 2021-01-05
Add instantiation and usage to class expressions example

mattbaker-digital:


orta:

Issue #1448, opened 2020-12-27
Docs: Bug in Utility Types examples?


mmxw:
**Page URL:** https://www.typescriptlang.org/docs/handbook/utility-types.html
**Issue:** these comments in the examples for Record and Omit seem to be bugs?

in Record example: 
> nav.about;
//   ^ = Could not get LSP result: v.a>b

in Omit example:

> todo;
// ^ = const todo: Pick

Thank you and happy holidays!

PR #1465, opened 2021-01-05
fix: #1448

HomyeeKing:
btw, the util type in `Template Literal Types` for example,  always say that can't find `Uppercase` ... 
any way to fix it ? upgrade typescript version since it's a experimental type or using fsMap  or just leave it as-is


orta:
Thanks!

PR #1470, opened 2021-01-06
Update typeAcquisition.ts: detect dynamic imports

YousefED:
See changes between:
https://regex101.com/r/hdEpzO/6
https://regex101.com/r/hdEpzO/8

This allows for dynamic imports (e.g.: let x.= await import("mod")


orta:
Cool, looks good to me 👍🏻

G-Rath:
Ironically I think this fixes a bug where strings like this would be matched incorrectly as imports:

```
"DELETE /repos/{owner}/{repo}/import": Operation<"/repos/{owner}/{repo}/import", "delete">;
```

image


This string (and many others) come from `@octokit/types` - you can see the errors the old regex is causing in the "Automatic Type Acquisition" example in the playaround, which has these errors in the console: 

image

This bug is also present in Deno.

(this was reported [here](https://github.com/octokit/rest.js/discussions/1708#discussioncomment-271684))

PR #1482, opened 2021-01-08
Add Error comment for accessing method of subclass

xypnox:
Currently an error occurs at the line:
```
department.generateReports();
```

<details><summary>Full example</summary>

```ts twoslash
// @errors: 2511 2339
abstract class Department {
  constructor(public name: string) {}
  printName(): void {
    console.log("Department name: " + this.name);
  }
  abstract printMeeting(): void; // must be implemented in derived classes
}
class AccountingDepartment extends Department {
  constructor() {
    super("Accounting and Auditing"); // constructors in derived classes must call super()
  }
  printMeeting(): void {
    console.log("The Accounting Department meets each Monday at 10am.");
  }
  generateReports(): void {
    console.log("Generating accounting reports...");
  }
}
let department: Department; // ok to create a reference to an abstract type
department = new Department(); // error: cannot create an instance of an abstract class
department = new AccountingDepartment(); // ok to create and assign a non-abstract subclass
department.printName();
department.printMeeting();
department.generateReports();
department.generateReports(); // error: department is not of type AccountingDepartment, cannot access generateReports
```

</details>

This happens because department is not of type `AccountingDepartment` hence can't access the `generateReports` method. This is however harder to understand because:

1. The type is declared in the start a few lines above to be Department and later object is assigned from constructor of `AccountingDepartment`.
2. There is no comment to point out that the error will occur and neither is it described in the text.

This confused me for a while to understand that the type didn't change and to get the method to work, department should be initialized with the correct type: `AccountingDepartment`. Even though the error message displayed in the code is very similar, adding the comment makes it known to the reader that this error is expected as this is not discussed in the text.

This PR adds the comment necessary to note that the error is expected and explains why the error occurred.

 Hopefully this will cause less confusion for someone else.

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Classes.md
+++ b/packages/documentation/copy/en/handbook-v1/Classes.md
@@ -487,7 +487,7 @@ department = new Department(); // error: cannot create an instance of an abstrac
 department = new AccountingDepartment(); // ok to create and assign a non-abstract subclass
 department.printName();
 department.printMeeting();
-department.generateReports();
+department.generateReports(); // error: department is not of type AccountingDepartment, cannot access generateReports
 ```
 
 ## Advanced Techniques


orta:
Looks good, it's consistent with the rest of the error messagings 👍🏻

PR #1484, opened 2021-01-08
Add concrete example of const assertion caveat

vegerot:
When I read the documentation originally, I thought it was obvious

```ts
(Math.random() < 0.5 ? 0 : 1)
```

could not use `as const`, because we don't know what `Math.random()` will produce.

I did not realize this applied to arithmetic as well.  
I think the multiplication example (which I really ran into just now) is a better example of this behavior


orta:
👍🏻

orta:
Agree, thanks

PR #1490, opened 2021-01-10
Summarize compatibility of some abstract types

jablko:
I frequently want to lookup what these types are assignable to/from and to contrast them. Would you consider adding this table to the docs?

https://github.com/jablko/TypeScript-Website/blob/patch-1/packages/documentation/copy/en/reference/Type%20Compatibility.md#any-unknown-object-void-undefined-null-and-never

I included some comparisons between the types that help me rationalize their distinctions.

Location: I think either Type Compatibility or Basic Types? Or not at all if it's not a good fit for the docs. The individual types are documented in Basic Types but I think this is more reference-y, which is why I appended it to Type Compatibility.

Heading: I wouldn't exactly call these types "basic" (at least not without including `boolean`, `number`, and `string`). "Abstract"? because they represent concepts other than `boolean`, `number`, `string`, and structured? But including "abstract" in the heading is itself abstract: I wouldn't look there for this table. "Any, unknown, object, void, undefined, null, and never" at least describes the information I want.

I'm happy to make changes.

FYI I used the following script to experiment with iterations:
```ts
import * as ts from "typescript";
import * as prettier from "prettier";

const types = [
  "any",
  "unknown",
  "object",
  "void",
  "undefined",
  "null",
  "never",
];
// --strict off and on
const cases = [false, true].map(createTable);
// console.table()
for (const table of cases) {
  console.table(
    mapValues(table, (row) =>
      mapValues(row, (diagnostics) => (diagnostics.length ? "❌" : "✔️"))
    )
  );
}
// GFM table
console.log(
  prettier.format(
    `
|   | ${types.join(" | ")}                  |
| - | ${types.map(() => ":-:").join(" | ")} |
` +
      Object.entries(cases[0])
        .map(
          ([source, row]) =>
            `| ${source} → | ${Object.entries(row)
              .map(([target, diagnostics]) =>
                diagnostics.length
                  ? ":x:"
                  : cases[1][source][target].length
                  ? ":white_check_mark:"
                  : source === target
                  ? ""
                  : ":heavy_check_mark:"
              )
              .join(" | ")} |
`
        )
        .join(""),
    { filepath: ".md" }
  )
);
// Graphviz directed graph of --strict off. Black edges for
// bidirectional assignability, red edges errors (uni- and
// bidirectional). Use e.g. `dot -O -Tsvg` to render SVG.
console.log("strict digraph {");
for (const [source, row] of Object.entries(cases[0])) {
  for (const [target, diagnostics] of Object.entries(row).slice(
    types.indexOf(source) + 1
  )) {
    const [forward] = diagnostics;
    const [back] = cases[0][target][source];
    console.log(
      !forward && back
        ? `  ${target} -> ${source} [color="#bf1818",tooltip="${back.messageText}"];`
        : `  ${source} -> ${target} [${[
            forward && 'color="#bf1818"',
            Boolean(forward) === Boolean(back) && "dir=both",
            forward && `tooltip="${forward.messageText}"`,
          ]
            .filter(Boolean)
            .join(",")}];`
    );
  }
}
console.log("}");

// Type check table of synthetic source files, return table of
// diagnostics
function createTable(strict: boolean) {
  const options = ts.getDefaultCompilerOptions();
  options.strict = strict;
  const table: {
    [T in typeof types[number]]: {
      [T in typeof types[number]]: ts.SourceFile;
    };
  } = Object.assign(
    {},
    ...types.map((source) => ({
      [source]: Object.assign(
        {},
        ...types.map((target) => ({
          [target]: ts.createSourceFile(
            `${source}-${target}.ts`,
            `{ const target: ${target} = undefined as unknown as ${source} }`,
            options.target!
          ),
        }))
      ),
    }))
  );
  // Must inject source files w/ ts.CompilerHost,
  // initializeTypeChecker() calls bindSourceFile() on
  // host.getSourceFiles()
  const files = Object.assign(
    {},
    ...Object.values(table).flatMap((row) =>
      Object.values(row).map((file) => ({ [file.fileName]: file }))
    )
  );
  const host = ts.createCompilerHost(options);
  const originalGetSourceFile = host.getSourceFile;
  host.getSourceFile = (fileName) =>
    files[fileName] ||
    originalGetSourceFile.call(host, fileName, options.target!);
  const program = ts.createProgram(Object.keys(files), options, host);
  return mapValues(table, (row) =>
    mapValues(row, (file) => ts.getPreEmitDiagnostics(program, file))
  );
}

// Map values, preserve keys
function mapValues<V, U>(
  obj: { [K in PropertyKey]: V },
  callbackfn: (value: V) => U
): { [K in PropertyKey]: U } {
  return Object.assign(
    {},
    ...Object.entries(obj).map(([key, value]) => ({ [key]: callbackfn(value) }))
  );
}

### DIFF
--- a/packages/documentation/copy/en/reference/Type Compatibility.md
+++ b/packages/documentation/copy/en/reference/Type Compatibility.md
@@ -288,3 +288,30 @@ These differ only in that assignment extends subtype compatibility with rules to
 
 Different places in the language use one of the two compatibility mechanisms, depending on the situation.
 For practical purposes, type compatibility is dictated by assignment compatibility, even in the cases of the `implements` and `extends` clauses.
+
+## Any, unknown, object, void, undefined, null, and never
+
+The following table summarizes assignability between some abstract types.
+Rows indicate what each is assignable to, columns indicate what is assignable to them.
+A ":white_check_mark:" indicates a combination that is compatible only when [`--strictNullChecks`](/tsconfig#strictNullChecks) is off.
+
+|             |        any         |      unknown       |       object       |        void        |     undefined      |        null        | never |
+| ----------- | :----------------: | :----------------: | :----------------: | :----------------: | :----------------: | :----------------: | :---: |
+| any →       |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |  :x:  |
+| unknown →   | :heavy_check_mark: |                    |        :x:         |        :x:         |        :x:         |        :x:         |  :x:  |
+| object →    | :heavy_check_mark: | :heavy_check_mark: |                    |        :x:         |        :x:         |        :x:         |  :x:  |
+| void →      | :heavy_check_mark: | :heavy_check_mark: |        :x:         |                    |        :x:         |        :x:         |  :x:  |
+| undefined → | :heavy_check_mark: | :heavy_check_mark: | :white_check_mark: | :heavy_check_mark: |                    | :white_check_mark: |  :x:  |
+| null →      | :heavy_check_mark: | :heavy_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |                    |  :x:  |
+| never →     | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |       |
+
+Reiterating [Basic Types](/handbook/basic-types.html):
+- Everything is assignable to itself.
+- `any` and `unknown` are the same in terms of what is assignable to them, different in that `unknown` is not assignable to anything except `any`.
+- `unknown` and `never` are like inverses of each other.
+  Everything is assignable to `unknown`, `never` is assignable to everything.
+  Nothing is assignable to `never`, `unknown` is not assignable to anything (except `any`).
+- `void` is not assignable to or from anything, with the following exceptions: `any`, `unknown`, `never`, `undefined`, and `null` (if `--strictNullChecks` is off, see table for details).
+- When `--strictNullChecks` is off, `null` and `undefined` are similar to `never`: assignable to most types, most types are not assignable to them.
+  They are assignable to each other.
+- When `--strictNullChecks` is on, they behave more like `void`: not assignable to or from anything, except for `any`, `unknown`, `never`, and `void` (`undefined` is always assignable to `void`).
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by orta** [comment]
I love this PR, its been my favorite PR of the 2021, I'll give a re-read with your feedback tomorrow but this is a really solid piece of work that makes things very clear. Well done.

File: packages/documentation/copy/en/reference/Type Compatibility.md [hunk 1]
```diff
@@ -288,3 +288,30 @@ These differ only in that assignment extends subtype compatibility with rules to
 
 Different places in the language use one of the two compatibility mechanisms, depending on the situation.
 For practical purposes, type compatibility is dictated by assignment compatibility, even in the cases of the `implements` and `extends` clauses.
+
+## Any, unknown, object, void, undefined, null, and never
```
orta: I thought about this a bit, and I think this is probably the right call. Primitives isn't a great option either, basics is a bit too ambiguous and I think this is the only real improvement I can consider, but I'm fine with it either way:

```suggestion
## `Any`, `unknown`, `object`, `void`, `undefined`, `null`, and `never` assignability
```

packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -288,3 +288,30 @@ These differ only in that assignment extends subtype compatibility with rules to
...
+Reiterating [Basic Types](/handbook/basic-types.html):
+- Everything is assignable to itself.
+- `any` and `unknown` are the same in terms of what is assignable to them, different in that `unknown` is not assignable to anything except `any`.
+- `unknown` and `never` are like inverses of each other.
+  Everything is assignable to `unknown`, `never` is assignable to everything.
+  Nothing is assignable to `never`, `unknown` is not assignable to anything (except `any`).
+- `void` is not assignable to or from anything, with the following exceptions: `any`, `unknown`, `never`, `undefined`, and `null` (if `--strictNullChecks` is off, see table for details).
+- When `--strictNullChecks` is off, `null` and `undefined` are similar to `never`: assignable to most types, most types are not assignable to them.
+  They are assignable to each other.
+- When `--strictNullChecks` is on, they behave more like `void`: not assignable to or from anything, except for `any`, `unknown`, `never`, and `void` (`undefined` is always assignable to `void`).
```
orta:
```suggestion
- When `--strictNullChecks` is on, `null` and `undefined` behave more like `void`: not assignable to or from anything, except for `any`, `unknown`, `never`, and `void` (`undefined` is always assignable to `void`).
```

No need to keep what we are referring to in your head, IMO (in a re-read I had to go back to the previous line to remember what we're talking about)

orta (approved):
I could merge as is, I'm happy with it - got some suggestions, your call on whether to take them of for this to go in as shipped.

jablko:
Thanks for the positive feedback and improvements! I like them.

PR #1492, opened 2021-01-12
Fix `@typescript/vfs` module path

Marcisbee:
In `@typescript/vfs` package.json `module` points to non existing file.

This is package content when loaded from npm:

image

And there's no `./dist/typescript-vsf.esm.js`. Shouldn't it be pointing to `./dist/vsf.esm.js`?


orta:
Yeah, I agree - thanks!

PR #1496, opened 2021-01-14
Adds copy npm instructions on the homepage

orta:
Re-uses the exact same UI metaphors from npm's package page

https://user-images.githubusercontent.com/49038/104592576-4f2e0100-5666-11eb-93d2-b01604fbe31d.mov


DanielRosenwasser:
Is there any reason we can't always make the copy icon visible? It's a nice transition, but I could see that just making it more obvious.

orta:
That is what the B test is going to be 👍🏻

PR #1509, opened 2021-01-20
Fix variables in equality narrowing code snippet

adamochayon:
The explanation refers to the variables as `x` and `y` while the code snippet contained `left` and `right`.


orta:
Cool thanks!

PR #1515, opened 2021-01-23
fix: typo

suica:
Fix one typo about permalink of the "keyof type operator" doc. 

The permalink was wrongly written as "/docs/handbook/2/indexed-access-types.html".


orta:
👍🏻

PR #1516, opened 2021-01-23
Add quote to property keys on tsconfig.json

uraway:
Fix the error `tsconfig.json:13:5 - error TS1327: String literal with double quotes expected.` that raises when tsc runs on tsconfig.json with unquoted property keys.


orta:
I bet this is prettier formatting json5, I'll merge and switch them to jsonc - thanks

PR #1519, opened 2021-01-24
Replace `unrooted` with `relative path`

gurjeet:
I have never seen the word `unrooted` used to refer to relative paths. I
believe that line is referring to relative path so let's use the proper
term.


orta:
Agree, thanks!

PR #1529, opened 2021-01-29
Update Advanced Types.md

goossaert:
Minor change to heading level of "Using type predicates" from #### to ###, so that it becomes consistent with the other headings.


orta:
Agree - thanks 👍🏻

orta:
That's not your fault

PR #1543, opened 2021-02-02
Fix community page

### LINKED ISSUE #1540: Idea: Community page margins are off in Firefox
image

This happens to be the case because of the `.community .callouts .callout a` selector which adds a `margin-right: 8px`. Unfortunately, it applies to the text link (not just the logo link), which pushes the description 8px out.

falinsky:
Fixes #1540.

Also, this fix makes spaces around StackOverflow, GitHub, Blog, etc. images even, just like it's done for the Meetups section images - so now the page is more consistent.

comunity-page-after-changes


orta:
Agree- thanks!

PR #1553, opened 2021-02-07
Update Basic Types.md

YSternlicht:
On line 342, the comment stated  Remember, undefined is a subtype of null.
However, if that is the case, why does it throw an error that "Argument of type 'undefined' is not assignable to parameter of type 'object | null'."
Surely, a subtype is assignable to its base type.

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Basic Types.md
+++ b/packages/documentation/copy/en/handbook-v1/Basic Types.md
@@ -339,7 +339,7 @@ declare function create(o: object | null): void;
 // OK
 create({ prop: 0 });
 create(null);
-create(undefined); // Remember, undefined is a not subtype of null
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v1/Basic Types.md
```diff
@@ -339,7 +339,7 @@ declare function create(o: object | null): void;
 // OK
 create({ prop: 0 });
 create(null);
-create(undefined); // Remember, undefined is a subtype of null
+create(undefined); // Remember, undefined is a not subtype of null
```
falinsky:
```suggestion
create(undefined); // Remember, undefined is not a subtype of null
```

falinsky (approved):
LGTM

orta:
Agree, thanks folks

PR #1558, opened 2021-02-08
Rearrange Utility Types.md

mordv:
Wouldn't it be better to move `Required` type closer to its opposite type `Partial`?

Personally, I've found it only after googling how it's called. Don't see why it's placed between InstanceType and ThisParameterType.


orta:
Agree 👍🏻 happy to merge post CLA

orta:
Thanks

PR #1576, opened 2021-02-17
[Docs] Adding a missing "This" and attempt to add clarity with less pronouns.

evantahler:


orta:
I agree, I like it, thanks

PR #1579, opened 2021-02-19
Fix typo: "it's declaration" -> "its declaration"

stephanerangaya:
I was on the playground and saw a typo: https://www.typescriptlang.org/play?q=252#example/types-vs-interfaces

image


orta:
Nice one, thanks!

PR #1583, opened 2021-02-21
Update Utility Types.md

daiplusplus:
The **reuse** of the page-title names inside the `Record` value's object-literals was very confusing for me, so I'm proposing this alternative based on the explanation on the SO post: https://stackoverflow.com/questions/51936369/what-is-the-record-type-in-typescript


orta:
Yeah, I agree

daiplusplus:
@orta Thanks - but I noticed something unexpected in the published version:

Why is it saying the type of `cats` is `Record` when the tooltip says it's `Record<CatName, CatInfo>` - or have I discovered an XSS vuln?

Normal:

image

Hover:

image

PR #1593, opened 2021-02-24
Update interfaces and variables names more descriptive

sethreidnz:
When reading through this I found it very useful, however I find using variable names like `x` and `y` and interfaces like `Named` to make it so abstract. I feel like people find concrete examples a little easier to understand. I think the whole doc could use this approach and make it more accessible, however I just started with the easiest ones to see if there was any interest. 

Let me know what you think.

### DIFF
--- a/packages/documentation/copy/en/reference/Type Compatibility.md
+++ b/packages/documentation/copy/en/reference/Type Compatibility.md
@@ -12,20 +12,20 @@ This is in contrast with nominal typing.
 Consider the following code:
 
 ```ts
-interface Named {
+interface Pet {
   name: string;
 }
 
-class Person {
+class Dog {
   name: string;
 }
 
-let p: Named;
+let pet: Pet;
 // OK, because of structural typing
-p = new Person();
+pet = new Dog();
 ```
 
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -36,33 +36,39 @@ TypeScript's type system allows certain operations that can't be known at compil
...
-In this case, `y` must have a member called `name` that is a string. It does, so the assignment is allowed.
+To check whether `dog` can be assigned to `pet`, the compiler checks each property of `pet` to find a corresponding compatible property in `dog`.
+In this case, `dog` must have a member called `name` that is a string. It does, so the assignment is allowed.
 
 The same rule for assignment is used when checking function call arguments:
 
 ```ts
-function greet(n: Named) {
-  console.log("Hello, " + n.name);
+interface Pet {
```
sethreidnz, line 58: I also got confused when the y was defined in a code block above, and though repeating that definition here makes it easier to read the example in isolation without skipping back over the previous paragraph.

orta:
I'll take them because the changes are pretty narrow, but these docs are on their last legs as the v2 is almost ready to deprecate them: https://www.typescriptlang.org/docs/handbook/2/basic-types.html

PR #1602, opened 2021-03-01
Sandbox: run script/downloadReleases.js for 4.2.2

tjjfvi:


orta:
Sure, thanks ( these are automatically applied on deploys, so I don't normally do it unless it's a side-effect of doing something else but I appreciate the thought )

PR #1605, opened 2021-03-03
Added "missing" return type declaration

jlecordier:
I have added a "missing" return type declaration (even though it is not in fact mandatory) because the part of the docs explaining that it is not mandatory is just below, and I don't want beginners to be disturbed before the explaination


orta:
Agree, thanks!

PR #1609, opened 2021-03-04
removed unnecessary "heh"

jlecordier:
I removed the "heh", because it is disturbing for non native english speakers (like me), and is not quite reminiscent of the first version of the handbook


orta:
Nah, I'm happy with this one, maybe it could go into italics though to indicate it's meta

orta:
Ah, it already had it, so I'm just going to close 👍🏻

PR #1610, opened 2021-03-04
removed (weird ?) aside

jlecordier:
I propose to remove this `<aside>` as the result is really ugly, more difficult to read, and I don't remember seeing them in the first version of the handbook


orta:
Yeah, looks like this mixes badly with some CSS which also uses aside - we've been using blockquotes everywhere else, I'll switch them over - thanks

PR #1613, opened 2021-03-05
fix "later" by "earlier"

jlecordier:
I think it should be written "earlier", or nothing at all, since the chapter about Type Inference is the one just before this chapter


orta:
Yeah, great point, both the original and these changes might just be inaccurate as this doc lives in the reference section and not in the handbook - I'll take this PR (so you can get the commits), and I'll run over this doc today to make that disconnect more explicit - thanks!

PR #1616, opened 2021-03-06
FIX GH-1578

### LINKED ISSUE #1578: Playground: unable to import submodule from npm package
I'm trying to import a submodule from a package imported from NPM.
But it doesn't seem to work.

The root module works just fine.

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgM2AGxgUynAviqCEOAcigA8ArAZwHoIxsBDGaakgbgCgg

image

```ts
import { Observable } from 'rxjs';       // <-- this works
import { filter } from 'rxjs/operators'; // <-- this doesn't work
```
`rxjs/operators` is included in the rxjs npm package

**Expected behavior:**
Should be able to import from `'rxjs/operators'`

**Actual behavior:**
This error is shown
> Cannot find module 'rxjs/operators' or its corresponding type declarations.(2307)

hahmadia:
An attempt to fix https://github.com/microsoft/TypeScript-Website/issues/1578

### DIFF
--- a/packages/sandbox/src/typeAcquisition.ts
+++ b/packages/sandbox/src/typeAcquisition.ts
@@ -16,8 +16,15 @@ const moduleJSONURL = (name: string) =>
   // prettier-ignore
   `https://ofcncog2cu-dsn.algolia.net/1/indexes/npm-search/${encodeURIComponent(name)}?attributes=types&x-algolia-agent=Algolia%20for%20vanilla%20JavaScript%20(lite)%203.27.1&x-algolia-application-id=OFCNCOG2CU&x-algolia-api-key=f54e21fa3a2a0160595bb058179bfb1e`
 
-const unpkgURL = (name: string, path: string) =>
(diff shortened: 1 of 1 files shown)


hahmadia:
Hey @orta ,

I'm trying to resolve https://github.com/microsoft/TypeScript-Website/issues/1578 . My PR helps get rid of the red underlining but I still get `cannot use import statement outside of module` when I run the code. Any ideas ? Would appreciate any help! 
image


I can't for the life of me figure out what I gotta do to fix this :( Been stuck on it for a while

orta:
Hey @hahmedia - so this _does_ fix the underlaying problem - which is that type acquisition  doesn't resolve the index.d.ts

The playground itself _does not_ load and bundle the JavaScript, and so you cannot hit "Run" and have those libraries imported. So, I think this PR is good to go, and I'll improve the error messaging in the error logs

Issue #1612, opened 2021-03-04
Docs: Undocumented behavior relating to TS2318 errors caused by noLib


RA80533:
**Page URL:**
https://www.typescriptlang.org/tsconfig#noLib

**Issue:**
Running `tsc` with `--noLib` causes the compiler to produce a handful of errors relating globals:

```console
$ tsc --noLib
error TS2318: Cannot find global type 'Array'.
error TS2318: Cannot find global type 'Boolean'.
error TS2318: Cannot find global type 'Function'.
error TS2318: Cannot find global type 'IArguments'.
error TS2318: Cannot find global type 'Number'.
error TS2318: Cannot find global type 'Object'.
error TS2318: Cannot find global type 'RegExp'.
error TS2318: Cannot find global type 'String'.

Found 8 errors.
```

Clarification was provided in https://github.com/microsoft/TypeScript/issues/3930#issuecomment-122953275 that this is the intended behavior; however, the documentation available in the reference guide for the TSConfig file does not provide any hint towards this.

**Recommended Fix:**
Document the purpose behind this behavior.

orta:
Yeah, this makes sense 👍🏻

hahmadia:
@orta

Please close , Thanks.

PR #1618, opened 2021-03-06
Update docs to mention error outs regarding noLib

hahmadia:
Addresses https://github.com/microsoft/TypeScript-Website/issues/1612

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/noLib.md
+++ b/packages/tsconfig-reference/copy/en/options/noLib.md
@@ -5,3 +5,5 @@ oneline: "Disable including any library files, including the default lib.d.ts."
 
 Disables the automatic inclusion of any library files.
 If this option is set, `lib` is ignored.
+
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/copy/en/options/noLib.md
```diff
@@ -5,3 +5,5 @@ oneline: "Disable including any library files, including the default lib.d.ts."
 
 Disables the automatic inclusion of any library files.
 If this option is set, `lib` is ignored.
+
+To avoid any error outputs, you must specify your own library with declarations for Array, Boolean, Function, IArguments, Number, Object, RegExp, and String.
```
orta:
```suggestion
TypeScript _cannot_ compile anything without a set of interfaces for key primitives like: `Array`, `Boolean`,`Function`, `IArguments`, `Number`, `Object`, `RegExp`, and `String`. It is expected that if you use `noLib` you will be including your own type definitions for these.

orta:
Thanks

Issue #1328, opened 2020-11-09
emit() does not actually create js files


akutruff:
program.emit() does not actually create JavaScript files.  I ran into this in my own code and came across this skipped test.

https://github.com/microsoft/TypeScript-Website/blob/1c27887e2537918466b636d5500a98e6af75ed7b/packages/typescript-vfs/test/index.test.ts#L87

orta:
Yeah, I need to look into what's going on there at some point.

PR #1626, opened 2021-03-08
Fix GH-1328

hahmadia:
After many hours of troubleshooting,  I was able to fix the long time broken test in https://github.com/microsoft/TypeScript-Website/issues/1328 .  The test was broken due to the fact that getEmitHost gets the writeFile from writeFileCallback or defines it from `var host = createProgramOptions.host || createCompilerHost(options)`. Passing in the writeFileCallback fixes this test.


orta:
Wow, very impressive

PR #1630, opened 2021-03-08
Adds a mention for this-based type guards in the handbook v2

orta:
This was only mentioned in the 1.8 release notes, now it gets a re-direct in the narrowing section and lives in the classes one.


Retsam:
Perhaps it'd be worth adding a specific example for using this to remove an undefined value?  This does seem to be a common usage of this pattern, and it may not be obvious from the `this is Subtype` examples.

```ts
class Box<T> {
    value?: T;
    hasValue(): this is { value: T } {
      return this.value !== undefined;
    }
}
if(box.hasValue()) {
    console.log(box.value.toUpperCase());
}
```

PR #1631, opened 2021-03-08
Update Narrowing.md

YSternlicht:
On line 435, the last example (const underWater3: Fish[] = zoo.filter<Fish>((pet) => isFish(pet));) throws an error. It does not mention it will throw an error, nor does it say why is the error there.


packages/documentation/copy/en/handbook-v2/Narrowing.md
```diff
@@ -430,6 +430,8 @@ const zoo: (Fish | Bird)[] = [getSmallPet(), getSmallPet(), getSmallPet()];
 const underWater1: Fish[] = zoo.filter(isFish);
 // or, equivalently
 const underWater2: Fish[] = zoo.filter<Fish>(isFish);
+
+However, the following will throw an error.
```
orta:
```suggestion
// However, the following will throw an error.
```

orta:
Thanks - we fixed this in https://github.com/microsoft/TypeScript-Website/pull/1684

PR #1632, opened 2021-03-09
Type Declarations.md: insert missing links

uid11:
[Type Declarations](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html) page is accessible through a search on the https://www.typescriptlang.org/, so I decided to correct the links on it. But if this is too hasty, or if other links were intended, this pull request can be closed.


orta:
Great, thanks! - not sur eon the long term life of that doc but better to be correct either way

PR #1636, opened 2021-03-09
Fix grammar - Soundness.ts playground example

bouscs:


orta:
Thanks, but the original is correct: "provably sound" aka it can be proven to be sound

PR #1639, opened 2021-03-09
Replace remaining <T> with <Type>

pauladams8:
#1586 missed some outside of code blocks, which made the docs confusing


orta:
Perfect, thanks!

PR #1641, opened 2021-03-09
Make parameter names in the description match the code

diegose:
I believe these refer to the function defined in line 133.
They are `Input` and `Output`, not `E` and `O`.


orta:
Yeah, this looks right to me, thanks!

PR #1642, opened 2021-03-10
Fix GH-1640

### LINKED ISSUE #1640: Docs: Object Types / Tuple Types / Optional tuple elements snippet error (handbook v2)
**Page URL:** https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types

**Issue:**

Capture d’écran 2021-03-09 à 21 13 14

hahmadia:
Fix https://github.com/microsoft/TypeScript-Website/issues/1640.
The cause of this error is that `token` is returned as undefined. I guess it is a bigger issue than just this ticket. Typescript isn't able to return definitions for length on self defined types including array like the above example.

Regardless, I still don't see the `token` being required here as I don't quite understand it's purpose and I think checking if we have the `quickInfo` and it's `displayParts` is enough.

Now we get correct display:
image

### DIFF
--- a/packages/ts-twoslasher/src/index.ts
+++ b/packages/ts-twoslasher/src/index.ts
@@ -1,7 +1,7 @@
 let hasLocalStorage = false
 try {
   hasLocalStorage = typeof localStorage !== `undefined`
-} catch (error) {}
+} catch (error) { }
 const hasProcess = typeof process !== `undefined`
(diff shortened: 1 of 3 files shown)


orta:
This is great work, I think `token` was used in an older build of twoslash, but I also can't think of a reason for why it's in there now! 

Very impressive PR, I've pushed a test to make sure we don't break this in the future but I'm really happy with this change (and it's that output so cool? `length: 2 | 3` )

PR #1643, opened 2021-03-10
Classes.md: Perhaps the term needs to be clarified

uid11:
Doesn't _more general type_ mean supertype? In such a case, can the type of the derived class be called a more general type of base class?
It seems that adding a new field narrows the derived type (makes it a subtype of the base type).
Perhaps my edit is not appropriate (maybe it could be called a more specific type), but I just wanted to draw attention to this place.


orta:
Thanks, I agree 👍🏻

PR #1652, opened 2021-03-11
Corrected playground 'jsx' and 'moduleResolution' markdown issue outputs

JoshuaKGoldberg:
I know the issue hasn't been marked as accepting PRs, but it seems like a pretty quick fix. Sending quickly just in case 😄


orta:
Yeah, seems legit to me, I only use that tag when I don't think I'm going to be solving it in the short term 👍🏻

PR #1657, opened 2021-03-11
Update Conditional Types.md

MartinJohns:
Expand section about type inference with the information regarding multiple signatures from the old documentation.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-inference-in-conditional-types


MartinJohns:
The new handbook simply dropped this crucial piece of information.

orta:
Agree, let's take this and I'l have a think about bringing over a lot more of the v1 version 👍🏻

MartinJohns:
@orta I just noticed that I introduced new foos, which you removed a while back: https://github.com/microsoft/TypeScript-Website/commit/925ad49f0049643ff9e5098343cb33642d6864fa

orta:
That's OK, I can handle all that 👍🏻

PR #1667, opened 2021-03-12
Add Omit example, document Keys type

karlhorky:
Hey @orta! 👋 Hope you are well.

A quick PR to:

- document the type of `Keys` (allows for a single string literal but also a union of string literals)
- add an `Omit` example that uses a union of string literals

If this doesn't fit the style correctly or needs to be modified in some way, let me know, happy to do so!


orta:
Yeah, I think this is a really good call 👍🏻

karlhorky:
Great, thanks for the merge!

PR #1668, opened 2021-03-12
Fixes wrong generator reference

Sc4ramouche:
Hey there, seems like the last example in the [`noImplicitAny` Errors Apply to Loose `yield` Expressions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html#noimplicitany-errors-apply-to-loose-yield-expressions) section of 4.2 release notes actually meant to refer to the `g4` generator function, if I understood this snippet correctly.

I'll also use this tiny PR as an opportunity to express my gratitude to the TypeScript team. You rock!


orta:
👍🏻 yep, that fix is right -  thanks!

PR #1671, opened 2021-03-13
Local Doc Search

hahmadia:
No issues but this came out of personal frustration where docsearch would use `typescriptlang.org` instead of my localhost. I kept having to copy the url and paste it in my browser and replace it to localhost to go where I wanted. SO I looked up docsearch docs and made it so if your on localhost, it will use localhost instead of typescriptlang.org.

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/TopNav.tsx
+++ b/packages/typescriptlang-org/src/components/layout/TopNav.tsx
@@ -23,7 +23,7 @@ export const SiteNav = (props: Props) => {
       apiKey: '3c2db2aef0c7ff26e8911267474a9b2c',
       indexName: 'typescriptlang',
       inputSelector: '.search input',
-      handleSelected: function (input, event, suggestion, datasetNumber, context) {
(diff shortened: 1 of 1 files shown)


orta:
I'm hesitant for two reasons:

 - I really need to do the work to get https://github.com/microsoft/TypeScript-Website/pull/928 in which would deprecate these changes
 - I don't want custom code for this running in prod, for example this might not allow ctrl + clicking to do a new tab and could interfere with the keyboard interactions 

If you make it so that this custom handler only occurs in dev, I'm cool with it 👍🏻

hahmadia:
Hey Orta,

So I've maded it such that our custom handler only runs if we are in a dev environment, otherwise undefined is passed and it will fallback on docsearches own handleSelected (https://github.com/algolia/docsearch/blob/7c1d293af9b828cf2553d7ea61c54eb78f4dc882/src/lib/DocSearch.js#L108)

orta (approved):
Thanks!

PR #1672, opened 2021-03-13
Fix GH-1670

### LINKED ISSUE #1670: Anchor symbol is too high relative to header
**Improvement Idea:** 

image

**Why:** It looks weird when it's too high

**Alternative ideas:** 🤷‍♂️

hahmadia:
Fixes https://github.com/microsoft/TypeScript-Website/issues/1670

image

image


orta:
Yeah, let's just do this for now, I'd rather it not be like this another week - I'll have to re-look at dealing with the static nav in another run

PR #1673, opened 2021-03-13
Fixes Non-null Assertion Operation example including | null

StefanoCappellini:
I think there's a small typo in the handbook v2. The undefined should be implied by the ? operator, thus IMO is the null that is missing.


orta:
Thanks, this idea got merged in another PR 👍🏻

PR #1684, opened 2021-03-17
Docs: Fix last example of "Using type predicates"

sliminas:
In the handbook the last example of ["Using type predicates"](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) is described as equivalent to the previous 2 examples but it causes an error making it different to the previous ones:

#
Screenshot_20210317_175206

#
Have a look at the code [before](https://www.typescriptlang.org/play/#code/PTAEAEFMCdoe2gZwFygEwGYAsBWAUAC4CeADpKAGICWiAFqALygDeoiA7lQLaoAUAlIwB8oAG5wqAE1ABfANyFS5AEJVo0pqwBmAGyJ9BDEeKmyFkyAGMdAQ2jktAVwB2lglTjPQAc0gEAylw2OjoACn4CqNR0oAA+oKrqCk6u7p6gNNG0vGQEUTT08YmS-Ki5GYiUBSx4oKD2BI7QXjl+oDaVWfwAdBzcoACEDEwuFlpUzpCSCjJ4IKAAtEuWjgRLC3iWnogEoABecHB8WXEJaiUA2gC6jKAXvgFBIeEEAgA0Pn6BwWER-B8Pb7PP5XBRbZw7UCjGAAdRsBBgAEZ8nRrrcDnBuuMdAjoLxMgV+Ap5ggPpAAI6OKiiYKQZwEPSbba7aHQOG4tAo2hopgYrFUHEwAA8WSE+MQXTBzKhzgsbPhMAwXJ5+0O-MF0BFBTFrQIhhEBLouv4RKAA) and [after](https://www.typescriptlang.org/play?#code/PTAEAEFMCdoe2gZwFygEwGYAsBWAUAC4CeADpKAGICWiAFqALygDeoiA7lQLaoAUAlIwB8oAG5wqAE1ABfANyFS5AEJVo0pqwBmAGyJ9BDEeKmyFkyAGMdAQ2jktAVwB2lglTjPQAc0gEAylw2OjoACn4CqNR0oAA+oKrqCk6u7p6gNNG0vGQEUTT08YmS-Ki5GYiUBSx4oKD2BI7QXjl+oDaVWfwAdBzcoACEDEwuFlpUzpCSCjJ4IKAAtEuWjgRLC3iWnogEoABecHB8WXEJaiUA2gC6jKAXvgFBIeEEAgA0Pn6BwWER-B8Pb7PP5XBRbZw7UCjGAAdRsBBgAEZ8nRrrcDnBuuMdAjoLxMgV+Ap5ggPpAAI6OKiiYKQZwEPSbba7aHQOG4tAo2hopgYrFUHEwAA8WSE+MQXTBzKhzgsbPhMAwXJ5+0O-MF0BFBTFrQIpVA5RoVRiRgqWV1-CJQA) in the Typescript playground.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -430,7 +430,7 @@ const zoo: (Fish | Bird)[] = [getSmallPet(), getSmallPet(), getSmallPet()];
 const underWater1: Fish[] = zoo.filter(isFish);
 // or, equivalently
 const underWater2: Fish[] = zoo.filter<Fish>(isFish);
-const underWater3: Fish[] = zoo.filter<Fish>((pet) => isFish(pet));
+const underWater3: Fish[] = zoo.filter<Fish>((pet): pet is Fish => isFish(pet));
 ```
 
 # Discriminated unions
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -418,9 +418,8 @@ it also knows that in the `else` branch, you _don't_ have a `Fish`, so you must 
 You may use the type guard `isFish` to filter an array of `Fish | Bird` and obtain an array of `Fish`:
 
 ```ts twoslash
-// @errors: 2345
-type Fish = { swim: () => void };
-type Bird = { fly: () => void };
+type Fish = { swim: () => void; name: string };
+type Bird = { fly: () => void; name: string };
 declare function getSmallPet(): Fish | Bird;
 function isFish(pet: Fish | Bird): pet is Fish {
   return (pet as Fish).swim !== undefined;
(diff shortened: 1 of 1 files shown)


hahmadia:
Would it also be correct to do
```ts
const underWater3: Fish[] = zoo.filter<Fish>((pet as Fish) => isFish(pet));
```
Any idea what the difference between the keyword `is` and `as` is?

sliminas:
@hahmadia 

[As you can check in this playground example](https://www.typescriptlang.org/play?#code/PTAEAEFMCdoe2gZwFygEwGYAsBWAUAC4CeADpKAGICWiAFqALygDeoiA7lQLaoAUAlIwB8oAG5wqAE1ABfANyFS5AEJVo0pqwBmAGyJ9BDEeKmyFkyAGMdAQ2jktAVwB2lglTjPQAc0gEAylw2OjoACn4CqNR0oAA+oKrqCk6u7p6gNNG0vGQEUTT08YmS-Ki5GYiUBSx4oKD2BI7QXjl+oDaVWfwAdBzcoACEDEwuFlpUzpCSCjJ4IKAAtEuWjgRLC3iWnogEoABecHB8WXEJaiUA2gC6jKAXvgFBIeEEAgA0Pn6BwWER-B8Pb7PP5XBRbZw7UCjGAAdRsBBgAEZ8nRrrcDnBuuMdAjoLxMgV+Ap5ggPpAAI6OKiiYKQZwEPSbba7aHQOG4tAo2hopgYrFUHEwAA8WSE+MQXTBzKhzgsbPhMAwXJ5+0O-MF0BFBTFrV2HSqdEMIgJdF1-CJQA) that doesn't work.

I think the difference is that 
```(pet as Fish) => isFish(pet)```
 narrows down the input parameter to the Fish type and
 ```(pet): pet is Fish => isFish(pet)```
 narrows down the return value of the function.

So `as` is used to narrow down variables and `is` is only used for `type predicates` to define the return type of a function.

orta:
Yeah, I see what's going on here - tricky. 

TypeScript's narrowing can't _guarantee_ (in a reasonable timeframe) that this is the only possible clause for a return value and doesn't directly move the narrowing across from `isFish` into the `filter`.  I'm a little conflicted about even keeping the line around as is to be honest.

orta:
I've pushed  a change which gives you a sense that you might need to repeat yourself if you're not just directly passing that function in

PR #1685, opened 2021-03-17
Update documentation around `type` error reporting

DuncanWilder:
Based on the linked playground for this point, switching between version 4.2.3 and any version lower yields a different error message. It appears that since version 4.2.3 (at least) the error message is descriptive and the same as an `interface`.


orta:
Agree, I was thinking about this lately and wondered about re-writing this sentence entirely. For now though, this is a good improvement.

PR #1687, opened 2021-03-17
Fixes GH-1682

### LINKED ISSUE #1682: Playground: Error text sometimes not shown
We don't always see error text in the Errors pane, even when errors exist

image

**Repro:**

https://www.typescriptlang.org/play?ts=4.3.0-dev.20210315#code/JYOwLgpgTgZghgYwgAgMJwDYYEaINapgAeAPACoB8yA3gLABQyTyA9iAM4CuCS77AXMgAURQWQCUyALxUQnALbZoAbgbNWIaFBZRBQrYLggAnpJnIAbi2AATVfQC+DBjYgIMcKChicQCMMBsyAiYOPjkFEJgABbAAmihuAgExBHigla29i5uHl7BbOxgwcSC6FhJKaRFUKAA5hTZ9CEV+AB02KA2Qv5E4vYA9APIAHoJrcmE1WC1IA3IccggLMVwfMB1IHDYGChgLONhk6m+eMsA7iAUzs2J7Z0gNiRRsfHlR1UkNfUUZlSZNkivX6DCGo2QAHUdHh2MgYKAUOdohAQMh5EZOKFjMguNgisAwJwAnNkDEUHUUdBgAhScYAA4oFgwUnIuG+fyBEAMIA

**Expected behavior:** We should see a text rendering of the errors in the Error pane

**Actual behavior:** Blank space

hahmadia:
Fixes GH-1682

Also, I've made some modifications.

1. Lets bail out earlier if we have no errors to display. No need to wait until after the playground is found for us to make that decision.
2. Let's avoid using JSON.stringify since the array is an array of numbers (unless I've been mistaken here) and lets compare values individually. From personal experience, we want to use stringify if we are doing a deep compare of objects but since it's just numbers, we can just evaluate every value and avoid the extra conversion.
3. Lets convert prevMarkers to type number for better type checking. If I've understood correctly, there will only ever be numbers stored.

### DIFF
--- a/packages/playground/src/sidebar/showErrors.ts
+++ b/packages/playground/src/sidebar/showErrors.ts
@@ -7,12 +7,19 @@ export const showErrors: PluginFactory = (i, utils) => {
   let container: HTMLElement
   let sandbox: Sandbox
   let ds: ReturnType<typeof utils.createDesignSystem>
-  let prevMarkers: any[] = []
+  let prevMarkers: number[] = []
 
   const updateUI = () => {
     if (!sandbox) return
     const model = sandbox.getModel()
     const markers = sandbox.monaco.editor.getModelMarkers({ resource: model.uri })
+
+    // Bail early if there's nothing to show
+    if (!markers.length) {
+      prevMarkers = []
+      ds.showEmptyScreen(localize("play_sidebar_errors_no_errors", "No errors"))
+      return
+    }
 
     // @ts-ignore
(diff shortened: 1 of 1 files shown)


orta:
These are solid improvements, nice work

I might make `ds.showEmptyScreen` not do DOM work if it's already showing an empty screen ( this could get called a few times ) to make this not mutate the DOM without cause, good work

PR #1688, opened 2021-03-17
fix example in ts4.2 release notes

kaznovac:
fixes example in ts4.2 release notes about `abstract new`

fingercrossed i got the code formatting microsyntax right


orta:
Hey, great work on twoslashing that code sample - looks right to me and the compiler didn't raise so it must be legit 👍🏻

PR #1690, opened 2021-03-18
Handbook fix error line in Modules.md

ghost:
Handbook Error (https://www.typescriptlang.org/docs/handbook/2/modules.html#typescript-specific-es-module-syntax): `'createCatName' cannot be used as a value because it was imported using 'import type'.` should be shown on `app.ts` filename where it's used instead on `animal.ts` where is declared

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -208,10 +208,10 @@ export const createCatName = () => "fluffy";
 
 // @filename: valid.ts
 import type { Cat, Dog } from "./animal.js";
-type Animals = Cat | Dog;
+export type Animals = Cat | Dog;
 
-// @errors: 1361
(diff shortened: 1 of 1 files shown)


orta:
I'm don't _think_ this will fix the underlaying issue, but I've tweaked it anyway. 

I think that might need a TypeScript compiler error message improvement to switch where the error message gets raised from

orta:
Maybe this is a twoslash issue, creating a sample app shows the underlines at the call site

```
$ /Users/ortatherox/dev/typescript/repros/types_error_pos/node_modules/.bin/tsc
myfile.ts:2:13 - error TS1361: 'thing' cannot be used as a value because it was imported using 'import type'.

2 console.log(thing)
              ~~~~~

  myfile.ts:1:14
    1 import type {thing} from "./animal"
                   ~~~~~
    'thing' was imported here.

```

PR #1694, opened 2021-03-18
Clarified Angular blurb in bootstrapping docs

teallarson:
"Be" is out of place here.  "Joyful" implies that the action itself feels joy; "enjoyable" is a more appropriate choice.

### DIFF
--- a/packages/typescriptlang-org/src/copy/en/documentation.ts
+++ b/packages/typescriptlang-org/src/copy/en/documentation.ts
@@ -26,7 +26,7 @@ export const docCopy = {
     "A delightful toolkit for building TypeScript-powered command-line apps",
   doc_frameworks: "Web Frameworks",
   doc_frameworks_angular_blurb:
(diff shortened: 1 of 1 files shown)


orta:
This _is_ (or _was_) their own definition, not ours  and so I can't really take changes to it - however, it looks like they've rebranded since and it seems "The modern web developer's platform" is the new one. I'll replace it with that.

packages/typescriptlang-org/src/copy/en/documentation.ts
```diff
@@ -26,7 +26,7 @@ export const docCopy = {
     "A delightful toolkit for building TypeScript-powered command-line apps",
   doc_frameworks: "Web Frameworks",
   doc_frameworks_angular_blurb:
-    "Makes writing beautiful apps be joyful and fun",
+    "Makes writing beautiful apps enjoyable and fun",
```
orta:
```suggestion
    "The modern web developer's platform",
```

PR #1697, opened 2021-03-19
Upper case file name fails

arafalov:
If link is clicked and is loaded - I am assuming - by Javascript, the upper case makes the page fail to load. Changing the leading character to lower-case makes it work (tested in dev tools). Opening the link in a new window also works, as that uses browser-native HTML fetch which is not case-sensitive.

There is the same problem in the search index (when searching for any Symbol's methods), but I don't know how to fix that.


orta:
Thanks! Maybe the sitemap has an uppercase somewhere too, thanks

PR #1698, opened 2021-03-21
Fixes GH-1693

### LINKED ISSUE #1693: Playground: Suggestion: Strip top-level export keywords when emitting JS
Stripping top-level `export` keywords from emitted JS in the Playground would make it easier to paste and work with code from a real file (which likely has one or more exports).

image

image

image

**Repro:**

Type any code into the Playground that `export`s anything from the top level (not within a `namespace`) and click Run

**Expected behavior:**

The code should run ignoring the `export`s

**Actual behavior:**

The code chokes on the `export`s and will not run until they are removed

hahmadia:
Fixes GH-1693

Create a duplicate model that is only purpose is to wrap code in the model inside a namespace then have it emitted. This way the text displayed to user doesn't change while solving the original problem.

image

### DIFF
--- a/packages/playground/src/sidebar/runtime.ts
+++ b/packages/playground/src/sidebar/runtime.ts
@@ -165,6 +165,11 @@ function rewireLoggingToElement(
       const prefix = `[<span class="log-${name}">${id}</span>]: `
       const eleContainerLog = eleOverflowLocator()
       allLogs.push(`${prefix}${output}<br>`)
+
+      if (output.includes("Unexpected token 'export'")) {
+        allLogs.push(`[<span class="log-warn">WRN</span>]: Please consider changing your Module to \"CommonJS\" in the TS Config settings`)
+      }
+
       eleLog.innerHTML = allLogs.join("<hr />")
       if (autoScroll && eleContainerLog) {
         eleContainerLog.scrollTop = eleContainerLog.scrollHeight
--- a/packages/playground/src/sidebar/runtime.ts
+++ b/packages/playground/src/sidebar/runtime.ts
@@ -155,6 +155,10 @@ function rewireLoggingToElement(
     } catch (error) {
       console.error(i("play_run_js_fail"))
       console.error(error)
+
+      if (error instanceof SyntaxError && /\bexport\b/u.test(error.message)) {
+        console.warn('Tip: Change the Module setting to "CommonJS" in TS Config settings to allow top-level exports to work in the Playground')
+      }
     }
   })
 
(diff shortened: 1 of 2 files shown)


orta:
Lots of discussion about this in [the discord](https://discord.com/channels/508357248330760243/640177465490407425/823230038698360902) but we think the right answer is to recommend people use commonjs in modules when they have an "export" in their source code

hahmadia:
@orta,

I've iterated on the discussion comments and came up with this solution. Let me know if this works.

packages/playground/src/sidebar/runtime.ts
```diff
@@ -165,6 +165,11 @@ function rewireLoggingToElement(
       const prefix = `[<span class="log-${name}">${id}</span>]: `
       const eleContainerLog = eleOverflowLocator()
       allLogs.push(`${prefix}${output}<br>`)
+
+      if (output.includes("Unexpected token 'export'")) {
+        allLogs.push(`[<span class="log-warn">WRN</span>]: Please consider changing your Module to \"CommonJS\" in the TS Config settings`)
```
ada-waffles: It might be good to provide some additional context on why this change would be useful. Also I don't think the template quotes are necessary here?
```suggestion
        allLogs.push('[<span class="log-warn">WRN</span>]: Tip: Change the Module setting to "CommonJS" in TS Config settings to allow top-level exports to work in the Playground')
```

packages/playground/src/sidebar/runtime.ts
```diff
@@ -165,6 +165,11 @@ function rewireLoggingToElement(
       const prefix = `[<span class="log-${name}">${id}</span>]: `
       const eleContainerLog = eleOverflowLocator()
       allLogs.push(`${prefix}${output}<br>`)
+
+      if (output.includes("Unexpected token 'export'")) {
```
ada-waffles: I like the idea of reacting to the actual error rather than the emitted JS, but I wonder how fragile this string is. How is output being captured? Could we potentially catch/intercept the actual `SyntaxError` thrown by the `eval`?

ada-waffles:
We also need to either precede the `eval` call with empty declarations for `module` and `exports`, or switch to `new Function` and pass them in as arguments, to make sure this setting change will _continue_ to make top-level exports work

ada-waffles:
Looks like there's already a `try/catch` around the `eval` call:
https://github.com/microsoft/TypeScript-Website/blob/6cfc8329122494d9cb9396928a701096d9fb1c59/packages/playground/src/sidebar/runtime.ts#L155-L158

Presumably, instead of reacting to the error when the message is logged, we could add
```ts
if (error instanceof SyntaxError && /\bexport\b/u.test(error.message)) {
  //Log our message here
}
```

I feel like that should be a _little_ more resilient to how different JS engines format the error message. Thoughts?

orta:
Nice work, this looks good to me 👍🏻

PR #1699, opened 2021-03-21
Fixes 1680, 1669 - Module: tsvsf - Does not normalize paths

### LINKED ISSUE #1669: Module: tsvsf - FSBackedSystem has a fixed getCurrentDirectory() that breaks relative imports
Im using `@typescript/vfs` and its `createFSBackedSystem` to create an environment to test my transformer. I stumbled on the following problem where I am trying to figure out what the intention of the following line is:
https://github.com/microsoft/TypeScript-Website/blob/4700f41b706108a515c9f40dbcd17e453b7856e4/packages/typescript-vfs/src/index.ts#L432

This breaks relative imports when a file is read from disk but im not sure why. When I override `system.getCurrentDirectory` with `() => process.cwd()` everything works as expected, but I'm not sure why.

For example:
```ts
// test/values.ts

export const value = 42;
```
```ts
// test/objects/interfaces.ts:3:73

import { value } from '../values';
// AssertionError: test/objects/interfaces.ts:3:73 Cannot find module '../values' or its corresponding type declarations
```

I tried setting `host.resolveModulesNames` to see if the compiler would use that instead, but that doesnt seem to be used at all.
I dumped a stack trace to see why `system.getCurrentDirectory` is being used to resolve file relative imports but I'm a bit lost on why the compiler does it like this.

<details>
<summary>
It seems that <a href="https://github.com/microsoft/TypeScript/blob/master/src/compiler/utilities.ts#L4168)"><code>src/compiler/utilities.ts#getPathsBasePath</code></a> is being used to figure out from where relative paths should be resolved
</summary>

```
at Object.getCurrentDirectory (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\test\createEnvironment.ts:51:21)
    at Object.getPathsBasePath (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:17451:192)       
    at tryLoadModuleUsingPathsIfEligible (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39662:36)
    at tryLoadModuleUsingOptionalResolutionSettings (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39643:24)
    at tryResolve (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39788:28)
    at C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39784:69
    at Object.forEach (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:384:30)
    at nodeModuleNameResolverWorker (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39784:25)   
    at nodeModuleNameResolver (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39776:16)
    at Object.resolveModuleName (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:39547:30)       
    at loader_1 (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106565:99)
    at loadWithLocalCache (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106314:42)
    at actualResolveModuleNamesWorker (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106566:129)
    at resolveModuleNamesWorker (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106803:26)      
    at resolveModuleNamesReusingOldState (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106900:24)
    at processImportedModules (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:108351:35)        
    at findSourceFileWorker (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:108153:17)
    at findSourceFile (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:108014:26)
    at C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:107966:85
    at getSourceFileFromReferenceWorker (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:107933:34)
    at processSourceFile (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:107966:13)
    at processRootFile (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:107776:13)
    at C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106658:67
    at Object.forEach (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:384:30)
    at Object.createProgram (C:\Users\kevin\Documents\Projects\ts-transform-runtime-check\node_modules\typescript\lib\typescript.js:106658:16)
```
</details>

Any ideas on what im missing here? I'd like `vfs` to behave as expected in its node environment, but im lost on what is going wrong here.

### LINKED ISSUE #1680: Module: tsvsf - Does not normalize paths
As reported in https://github.com/microsoft/TypeScript-Website/issues/1669#issuecomment-799378345 the `@typescript/vfs` package tests will fail on Windows because of path seperator issues.

The code at https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescript-vfs/test/fsbacked.test.ts#L43-L46

Will produce [different results](https://nodejs.org/api/path.html#path_windows_vs_posix) depending on the operating system and its path seperator. A bandaid fix would be to use `path.posix` in the tests to force `/` to be used as seperator.

Looking at the compiler it states in https://github.com/microsoft/TypeScript/blob/master/src/compiler/path.ts#L3-L7 that internally `/` is used, but it expects the `CompilerHost | LanguageServiceHost` to 'correctly handle paths in our specified format'. Im not sure what means, but I understand it as the host object being responsible for using the internal used seperator. If so I would suggest patching `@typescript/vfs` to normalize the paths on input.

If you agree I'll take a look in how the compiler implements its hosts and see where `vfs` needs to be patched to behave similar.

kevinramharak:
This PR attempts to fix #1680 (and fix #1669) with the following changes:
- adds tests for both the `\` and `/` directory seperators and patches the `ts.System` returned by `createFSBackedSystem` to normalize paths when interacting with the underlying FS `Map<string, string>`.

I tried to figure out how the typescript compiler does this and it seems that the compiler internally uses `/` for every path. It assumes the `host` can deal with those slashes. This requires `vfs` to normalise any `\` to `/` in the `Map` representing the virtual file system to account for this internal behaviour. The normalisation is only applied when interacting with the `Map`. Any calls delegates to `nodeSys` are not touched because both posix and win32 can deal with `/` seperated paths just fine.

~~I haven't touched the interactions that pass around `ts.SourceFile` instances. I currently assume that those are made with the `ts` api and therefore already have normalized paths.~~
After checking the TypeScript source code creating `SourceFile` instances will have their paths normalized. So its best no to touch these. We could add path normalisation when retrieving a source file with `host.getSourceFile`, but its probably better to fail and have the caller be responsible to pass in normalized paths.

It additionally:
- exposes the `ts.LanguageServiceHost` when calling `createVirtualTypeScriptEnvironment`. It is annoying not to have a reference to the `host` being used. It can be used to implement custom module resolution by implementing `resolveModuleNames`.
- fixes `system.readDirectory` on the `system` returned by `createFSBackedSystem`.

### DIFF
--- a/packages/typescript-vfs/src/index.ts
+++ b/packages/typescript-vfs/src/index.ts
@@ -19,7 +19,14 @@ const debugLog = shouldDebug ? console.log : (_message?: any, ..._optionalParams
 const directorySeparator = "/"
 const backslashRegExp = /\\/g
 const normalizeSlashes = (path: string): string => path.replace(backslashRegExp, directorySeparator)
-const normalizeFsMap = (fsMap: Map<string, string>): Map<string, string> => new Map([...fsMap.entries()].map(([filepath, content]) => [normalizeSlashes(filepath), content] as const))
+const normalizeFsMap = (fsMap: Map<string, string>) => {
+  for (const [fileName, content] of fsMap) {
+    const normalizedFileName = normalizeSlashes(fileName);
+    if (fileName !== normalizedFileName) {
+      fsMap.set(fileName, content);
+    }
+  }
+}
 
 export interface VirtualTypeScriptEnvironment {
   sys: System
@@ -416,7 +423,7 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
   const nodeSys = ts.sys
   const tsLib = path.dirname(require.resolve("typescript"))
 
-  files = normalizeFsMap(files)
+  normalizeFsMap(files)
 
   return {
     // @ts-ignore
--- a/packages/typescript-vfs/src/index.ts
+++ b/packages/typescript-vfs/src/index.ts
@@ -15,8 +15,15 @@ const hasProcess = typeof process !== `undefined`
 const shouldDebug = (hasLocalStorage && localStorage.getItem("DEBUG")) || (hasProcess && process.env.DEBUG)
 const debugLog = shouldDebug ? console.log : (_message?: any, ..._optionalParams: any[]) => ""
 
+
+const directorySeparator = "/"
+const backslashRegExp = /\\/g
+const normalizeSlashes = (path: string): string => path.replace(backslashRegExp, directorySeparator)
+const normalizedFsMap = (fsMap: Map<string, string>): Map<string, string> => new Map([...fsMap.entries()].map(([filepath, content]) => [normalizeSlashes(filepath), content] as const))
+
 export interface VirtualTypeScriptEnvironment {
   sys: System
+  host: LanguageServiceHost,
   languageService: import("typescript").LanguageService
   getSourceFile: (fileName: string) => import("typescript").SourceFile | undefined
   createFile: (fileName: string, content: string) => void
@@ -62,6 +69,7 @@ export function createVirtualTypeScriptEnvironment(
     // @ts-ignore
     name: "vfs",
     sys,
+    host: languageServiceHost,
     languageService,
     getSourceFile: fileName => languageService.getProgram()?.getSourceFile(fileName),
 
@@ -401,12 +409,14 @@ export function createSystem(files: Map<string, string>): System {
 export function createFSBackedSystem(files: Map<string, string>, _projectRoot: string, ts: TS): System {
   // We need to make an isolated folder for the tsconfig, but also need to be able to resolve the
   // existing node_modules structures going back through the history
-  const root = _projectRoot + "/vfs"
+  const root = normalizeSlashes(_projectRoot + "/vfs")
   const path = require("path")
 
   // The default System in TypeScript
   const nodeSys = ts.sys
   const tsLib = path.dirname(require.resolve("typescript"))
+
+  files = normalizedFsMap(files)
 
   return {
     // @ts-ignore
@@ -416,11 +426,12 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
     createDirectory: () => notImplemented("createDirectory"),
     // TODO: could make a real file tree
     directoryExists: audit("directoryExists", directory => {
-      return Array.from(files.keys()).some(path => path.startsWith(directory)) || nodeSys.directoryExists(directory)
+      const normalizedDirectory = normalizeSlashes(directory)
+      return Array.from(files.keys()).some(path => path.startsWith(normalizedDirectory)) || nodeSys.directoryExists(directory)
     }),
     exit: nodeSys.exit,
     fileExists: audit("fileExists", fileName => {
-      if (files.has(fileName)) return true
+      if (files.has(normalizeSlashes(fileName))) return true
       // Don't let other tsconfigs end up touching the vfs
       if (fileName.includes("tsconfig.json") || fileName.includes("tsconfig.json")) return false
       if (fileName.startsWith("/lib")) {
@@ -433,14 +444,15 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
     getDirectories: nodeSys.getDirectories,
     getExecutingFilePath: () => notImplemented("getExecutingFilePath"),
     readDirectory: audit("readDirectory", (...args) => {
-      if (args[0] === "/") {
-        return Array.from(files.keys())
+      if (normalizeSlashes(args[0]) === root) {
+        return Array.from(files.keys()).filter(file => file.startsWith(root));
       } else {
         return nodeSys.readDirectory(...args)
       }
     }),
     readFile: audit("readFile", fileName => {
-      if (files.has(fileName)) return files.get(fileName)
+      const normalizedFileName = normalizeSlashes(fileName)
+      if (files.has(normalizedFileName)) return files.get(normalizedFileName)
       if (fileName.startsWith("/lib")) {
         const tsLibName = `${tsLib}/${fileName.replace("/", "")}`
         const result = nodeSys.readFile(tsLibName)
@@ -455,14 +467,15 @@ export function createFSBackedSystem(files: Map<string, string>, _projectRoot: s
       return nodeSys.readFile(fileName)
     }),
     resolvePath: path => {
-      if (files.has(path)) return path
+      const normalizedPath = normalizeSlashes(path)
+      if (files.has(normalizedPath)) return normalizedPath
       return nodeSys.resolvePath(path)
     },
     newLine: "\n",
     useCaseSensitiveFileNames: true,
     write: () => notImplemented("write"),
     writeFile: (fileName, contents) => {
-      files.set(fileName, contents)
+      files.set(normalizeSlashes(fileName), contents)
     },
   }
 }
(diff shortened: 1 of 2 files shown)


packages/typescript-vfs/src/index.ts
```diff
@@ -493,11 +506,12 @@ export function createVirtualCompilerHost(sys: System, compilerOptions: Compiler
...
       getSourceFile: fileName => {
+        const normalizedFileName = normalizeSlashes(fileName)
         return (
-          sourceFiles.get(fileName) ||
+          sourceFiles.get(normalizedFileName) ||
           save(
             ts.createSourceFile(
-              fileName,
+              normalizedFileName,
               sys.readFile(fileName)!,
```
kevinramharak: I might have missed one here.

kevinramharak: Nope,  no normalisation when interacting with SourceFile instances

packages/typescript-vfs/src/index.ts
```diff
@@ -9,14 +9,21 @@ type TS = typeof import("typescript")
...
 
 const hasProcess = typeof process !== `undefined`
 const shouldDebug = (hasLocalStorage && localStorage.getItem("DEBUG")) || (hasProcess && process.env.DEBUG)
 const debugLog = shouldDebug ? console.log : (_message?: any, ..._optionalParams: any[]) => ""
 
+
+const directorySeparator = "/"
+const backslashRegExp = /\\/g
+const normalizeSlashes = (path: string): string => path.replace(backslashRegExp, directorySeparator)
+const normalizeFsMap = (fsMap: Map<string, string>): Map<string, string> => new Map([...fsMap.entries()].map(([filepath, content]) => [normalizeSlashes(filepath), content] as const))
```
kevinramharak: This code typechecks, but transpilation seems to break it? This is really weird

kevinramharak: So this code seems to be transpiled incorrectly without using `--downlevelIteration`. @orta I don't know if this is a project wide problem. Should I rewrite it to work without needing that flag or should the flag be turned on?

See a playground link here:
- [failing](https://www.typescriptlang.org/play?downlevelIteration=false&target=1&ts=4.3.0-dev.20210315#code/LAKAxg9gdgzgLgAgCYEsBOBTMcJoJ4DKGADgIZqk5oIC8CARAPT2iSyIBGpYA1jADakYACwBKGAOYBRAB7FaCRgB0ljCa2jwEUXAFtS-FAC8MBQSIwwFACjJxhALgTw0KKBICUTl24m0AfAh2wgB0mMSCYBjWXLwCQmKSssQANMjoWFSEJOSUuB4a7Np6BsYYAGIwALKk8nTWAGbVtU41xAA8Pu5pXRL+XghtnXCu3c4jvoE0gVAYAO6DtdYA2iFrTW0hGFATltYeALoh+sTWKw0o-Dn2aWxw23AHHgEIyzpo+oYmZgl7F1fBDy3aD3HYHBBCBBseAeDwAblAhS0G1qClmCzaK2WAHJSCpSNi0rjsQc0jiOCoOISENiqQcngjwJoIFcQvwIBJGs1iPCkYh3p8ykg0SUvhVuVy2ryQIxGAg5ih7BAAK6IAC0aqQEDmUCuADcMPwAJL3ChwFDQBD2DDFD6lEzCk4IFBWDC6YhwPBIlkYNkc6wC+0YJDSoA)
- [succeeding](https://www.typescriptlang.org/play?downlevelIteration=true&target=1&ts=4.3.0-dev.20210315#code/LAKAxg9gdgzgLgAgCYEsBOBTMcJoJ4DKGADgIZqk5oIC8CARAPT2iSyIBGpYA1jADakYACwBKGAOYBRAB7FaCRgB0ljCa2jwEUXAFtS-FAC8MBQSIwwFACjJxhALgTw0KKBICUTl24m0AfAh2wgB0mMSCYBjWXLwCQmKSssQANMjoWFSEJOSUuB4a7Np6BsYYAGIwALKk8nTWAGbVtU41xAA8Pu5pXRL+XghtnXCu3c4jvoE0gVAYAO6DtdYA2iFrTW0hGFATltYeALoh+sTWKw0o-Dn2aWxw23AHHgEIyzpo+oYmZgl7F1fBDy3aD3HYHBBCBBseAeDwAblAhS0G1qClmCzaK2WAHJSCpSNi0rjsQc0jiOCoOISENiqQcngjwJoIFcQvwIBJGs1iPCkYh3p8ykg0SUvhVuVy2ryQIxGAg5ih7BAAK6IAC0aqQEDmUCuADcMPwAJL3ChwFDQBD2DDFD6lEzCk4IFBWDC6YhwPBIlkYNkc6wC+0YJDSoA)

kevinramharak: Another problem this introduces is that its a copy of the original `Map` thus invalidating any previous references to the `Map` object. The modification should happen in place. That would also solve the transpilation issue, but I think its still worth looking at.

kevinramharak:
This also fixes #1669. It appears non-normalized slashes was the problem

kevinramharak (commented):
I ended up creating a copy of the given `fsMap` and having those `keys` normalized. This does introduce inconsistant behviour between `createFSBackedSystem` and the other factory functions. Changes to the input `Map` to `createFSBackedSystem` will not propegate to the returned `ts.System`.

kevinramharak:
After thinking this trough I'm convinced that all interactions with `vfs` should work just as it would when interacting with the compiler API. This requires the `Map` to have normalized slashes and all interaction with `System` and `*Host` objects will use `/` slashes in their internal representations and normalize any input paths that need to interact with these representations. This allows any environment to use either `\\` or `/` as directory seperators, but it will only make sense on a windows environment.

This **is** a **breaking** change whenever `vfs` is used in the following pattern:
```ts

const fs = new Map<string, string>();
const system = createSystem(fs);

const path = 'file.ts';
const content = 'const x = 2';
fs.set(path, content);

// before
system.readFile(path) === fs.get(path); // --- true
// after
system.readFile(path) === fs.get(path); // +++ false
```

kevinramharak:
This breaking change breaks tests and is probably to severe to actually push trough. I don't know what the best solution is in this case. The compiler will interact with the `Map` with normalized slashes. Any `\` style paths will cause crashes. Maybe its best to put the burden on the caller of `vfs.create*` functions to use `/` style paths only. This means that `path.*` functions on windows will cause errors, using the `path.posix.*` would be the recommended way of using `vfs` cross platform

kevinramharak:
I'm abandoning this PR because it doesn't really fix the issue, it just introduces more complexity.
A more correct PR would be to change the tests to use `path.posix` explicitly and document (in the `README.md`) that windows paths are not supported.

PR #1702, opened 2021-03-24
Fix broken link to tsconfig reference in Creating DTS files From JS.md

jamisuomalainen:
While exploring the documentation site [here](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html), I noticed a broken link that should point to the correct location (`/tsconfig`). This pull request fixes that.

Please let me know if anything else is required of me in order to have this fix merged.

All the best!


orta:
That's perfect, thanks!

PR #1703, opened 2021-03-24
Fix capitalization of TypeScript

ugultopu:


orta:
I bow my head in shame. Thanks for this one!

PR #1707, opened 2021-03-24
Single CTA for the TS index

orta:
https://user-images.githubusercontent.com/49038/112375133-d4730900-8cda-11eb-960e-46d3a48a1a25.mov


orta:
This is wrapped up now

PR #1715, opened 2021-03-27
Remove 'tense' from the reference to 'Classes' section

navneetkarnani:
The current text refers to the classes section in the 'past tense' when in the handbook flow it is in the future. I am proposing that the sentence be changed to 'tense neutral' to account for the reading sequence changes, if any, that happen, or if the reader follows a different flow.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Generics.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Generics.md
@@ -248,7 +248,7 @@ console.log(stringNumeric.add(stringNumeric.zeroValue, "test"));
 
 Just as with interface, putting the type parameter on the class itself lets us make sure all of the properties of the class are working with the same type.
 
-As we covered in [our section on classes](/docs/handbook/2/classes.html), a class has two sides to its type: the static side and the instance side.
+As cover in [our section on classes](/docs/handbook/2/classes.html), a class has two sides to its type: the static side and the instance side.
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Type Manipulation/Generics.md
```diff
@@ -248,7 +248,7 @@ console.log(stringNumeric.add(stringNumeric.zeroValue, "test"));
 
 Just as with interface, putting the type parameter on the class itself lets us make sure all of the properties of the class are working with the same type.
 
-As we covered in [our section on classes](/docs/handbook/2/classes.html), a class has two sides to its type: the static side and the instance side.
+As cover in [our section on classes](/docs/handbook/2/classes.html), a class has two sides to its type: the static side and the instance side.
```
orta:
```suggestion
As we will cover in [our section on classes](/docs/handbook/2/classes.html), a class has two sides to its type: the static side and the instance side.
```

There isn't really a tense neutral way to say this in English ("see also" maybe) but I think switching the tense is reasonable

navneetkarnani: I missed the "we" between the "As" and "cover". See if that has a different reading impact. Just pushed a change to the branch for that.

orta: I agree, I like your version 👍🏻

orta:
Thanks

PR #1720, opened 2021-03-28
fix the import types example

cesarwbr:


orta:
Interesting, I guess we support a very lax syntax here because it does actually work

Screen Shot 2021-03-29 at 12 01 27 PM

But let's fix this anyway, thanks

PR #1730, opened 2021-04-01
Fixed required and optional

DigitalCop:
I think required and optional is wrong. ? is used for required


orta:
Thanks, but these example aren't wrong - you'd be missing the usage of the utility type in each code sample if we switched it. 

Maybe we should use the highlighter to highlight that line to make it more obvious.

PR #1731, opened 2021-04-01
fix header code tag margin right

jlecordier:
I propose to add a margin on the right of a code element, because otherwise there in no space between the background color of a code element and the following text

Before : 
image

After : 
image


orta:
I agree, thanks

PR #1734, opened 2021-04-03
An example implementation

nxpatterns:
An example implementation is easier to understand than just throwing an "unimplemented".


orta:
Thanks, but I think we're OK with unimplemented here

nxpatterns:
Absolutely no problem, you're welcome. I've cloned the documentation, and I'm creating my own one. 

Maybe just an important note: If you do not implement your type definitions, you won't be aware of [pitfalls like that](https://stackoverflow.com/questions/59904247/).

PR #1741, opened 2021-04-07
Minor grammatical error fix

omasher:
I have added a minor grammatical update that appears to have been an oversight.


orta:
thanks!

PR #1742, opened 2021-04-07
Update to TS 4.3b

orta:
This also includes a yarn update


packages/tsconfig-reference/copy/en/options/noImplicitOverride.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+In this case, TypeScript has provided no warning that `download` on `SharedAlbum` _expected_ to override a function in the base class.
+
+Using `noImplicitOverride` you can ensure that the sub-classes never go out of sync, by ensuring that functions which override include the keyword `override`.
+
+The following example has `noImplicitOverride` enabled, and you can see the error received when `override` is missing:
+
+```ts twoslash
+// @noImplicitOverride
```
orta, line 53:
```suggestion
// @noImplicitOverride
// @errors: 4114
```

PR #1743, opened 2021-04-08
Change playground SEO config

Nixinova:
Expanded the abbreviation "TS" and reiterated "TypeScript" in the desc.


orta:
This is on purpose https://www.reddit.com/r/typescript/comments/mkmzcd/careful_out_there_friends/gtha6yv/?context=3

PR #1746, opened 2021-04-08
Update gatsby

orta:
Looks like I also included the popover in twoslash also, so I'll get that polished in this PR too


orta:
I'd like to see this on staging

PR #1750, opened 2021-04-10
Update Module Resolution.md

nguyenyou:
It should be "typings"

https://github.com/microsoft/TypeScript/blob/131875bb849c0a9c56c26d09881453aace0cbfed/package.json#L23


orta:
TypeScript supports both fields, "types" is generally what we recommend though 👍🏻

PR #1754, opened 2021-04-14
add example on using Interable interface

ycmjason:


orta:
Yeah, this seems good - thanks

PR #1757, opened 2021-04-15
Update Gulp.md

Cheesefoma:
Added a full Gulpfile example and added few comments to the Watchify, Babel, and Uglify "change your gulpfile to the following" section


RyanCavanaugh:
Thanks for taking the time to submit this PR!

While we appreciate the positive edits you've made, we have noticed a number of typographical, grammatical, or technical errors that would need to be addressed before this could be merged. In order to ensure that we can efficiently work through our PR backlog, we have decided to close those PRs which require significant copyediting before they can be merged.

PR #1758, opened 2021-04-15
Add missing word

cdvrooman:
It is not clear what opposing format the enum format is better than.

### DIFF
--- a/packages/documentation/copy/en/reference/Enums.md
+++ b/packages/documentation/copy/en/reference/Enums.md
@@ -416,4 +416,4 @@ walk(EDirection.Left);
 run(ODirection.Right);
 ```
 
-The biggest argument in favour of this format over is that it keeps your codebase aligned with the state of JavaScript, and [when/if](https://github.com/rbuckton/proposal-enum) enums are added to JavaScript then you can move to the additional syntax.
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Enums.md
```diff
@@ -416,4 +416,4 @@ walk(EDirection.Left);
 run(ODirection.Right);
 ```
 
-The biggest argument in favour of this format over is that it keeps your codebase aligned with the state of JavaScript, and [when/if](https://github.com/rbuckton/proposal-enum) enums are added to JavaScript then you can move to the additional syntax.
+The biggest argument in favour of this format over (WHAT?) is that it keeps your codebase aligned with the state of JavaScript, and [when/if](https://github.com/rbuckton/proposal-enum) enums are added to JavaScript then you can move to the additional syntax.
```
orta:
```suggestion
The biggest argument in favour of this format over TypeScript's `enum` is that it keeps your codebase aligned with the state of JavaScript, and [when/if](https://github.com/rbuckton/proposal-enum) enums are added to JavaScript then you can move to the additional syntax.
```

orta:
Thanks

PR #1760, opened 2021-04-16
re: 'as const' - change prefix to suffix

HecticSerenity:
I recommend changing the word prefix to suffix.  'as const' is a suffix and the word prefix creates confusion because, in this example, the word 'const' is a prefix to the line of code.


upupming (approved):

orta:
Agree

PR #1761, opened 2021-04-16
fix: object is not a primitive type

upupming:
According to https://developer.mozilla.org/en-US/docs/Glossary/Primitive and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures, it seems `object` is not a primitive type in JS, doesn't it?

If I am wrong, please feel free to close this PR.


SMAKSS (approved):

orta:
Yeah, I think I agree - thanks

PR #1762, opened 2021-04-17
Docs: Fix navigation for type manipulation section

### LINKED ISSUE #1723: Docs: No previous/next links at the bottom of 'types from types' section
**Page URL:** 
https://www.typescriptlang.org/docs/handbook/2/types-from-types.html

**Issue:** 
No previous/next links at the bottom of the page unlike https://www.typescriptlang.org/docs/handbook/2/objects.html which correctly links to the above URL for its next link. The first page I see after this URL with a next link in it in the Handbook is https://www.typescriptlang.org/docs/handbook/2/classes.html and it doesn't link to its previous page either.

**Recommended Fix:** 
I see navigation data in the codebase for this but I'm guessing it is either declared wrong or not being used fully to build these previous/next links for pages in the handbook that are grouped under a folder.

https://github.com/microsoft/TypeScript-Website/blob/25d856fd80d853734f6f076a37f94d23e0ade094/packages/typescriptlang-org/src/lib/documentationNavigation.ts

mikeesto:
Closes https://github.com/microsoft/TypeScript-Website/issues/1723

The previous/next links aren't currently being generated for Type Manipulation because the subitem doesn't have chronological set to true & it's returning early when trying to compute the links: https://github.com/microsoft/TypeScript-Website/blob/2750cd8bee1f8d8101d13b55da7a1a4e5f5c479b/packages/typescriptlang-org/src/lib/documentationNavigation.ts#L4951 This PR resolves it

### DIFF
--- a/packages/documentation/scripts/generateDocsNavigationPerLanguage.js
+++ b/packages/documentation/scripts/generateDocsNavigationPerLanguage.js
@@ -52,6 +52,7 @@ const handbookPages = [
       { file: "handbook-v2/Object Types.md" },
       {
         title: "Type Manipulation",
+        chronological: true,
         items: [
           { file: "handbook-v2/Type Manipulation/_Creating Types from Types.md" },
           { file: "handbook-v2/Type Manipulation/Generics.md" },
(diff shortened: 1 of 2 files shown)


orta:
oh, that makes a lot of sense! Thanks!

PR #1764, opened 2021-04-21
Added @deprecated tag to jsdoc support

rohit-gohri:
Since this is also a part of JSDoc, I thought this should be mentioned here also.

Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#-deprecated--support


orta:
I'm not wild on including a screenshot, but I agree with the inclusion, we actually have the capability to show auto-complete inline in a code sample, so I'll edit this PR to use that

PR #1767, opened 2021-04-22
Edit info about allowUnreachableCode and allowUnusedLabels flags.

barinbritva:
These flags by default are "true". In this case that means to enable strict for them you need to set "false". This behavior is opposite to flags from "Strict Checks", which are by default "false", but should be set up to "true" to enable strictness


packages/tsconfig-reference/copy/en/options/allowUnreachableCode.md
```diff
@@ -5,7 +5,6 @@ oneline: "Disable error reporting for unreachable code."
 
 When:
 
-- `undefined` (default) provide suggestions as warnings to editors
```
barinbritva, line 8: Actually, there is no such value. It's from obsolete version of TS, I believe.

barinbritva:
I noticed, I was not totally right. Will redo it.

barinbritva:
Here is the new one - #1768.

PR #1770, opened 2021-04-23
Update function parameter bivariance section

metaweta:
Does this solution remove the need for function parameter bivariance in the language?  If not, what other example can't be solved this way?

I'm not sure how to link to the sections on conditional types and function overloading.


RyanCavanaugh:
Thank you for your recent pull request for documentation! We appreciate your time and effort in submitting your work for review.

After reviewing your changes, we've decided to decline this particular edit. While the content is correct, the Handbook in particular requires a certain amount of brevity, and needs to maintain a tight focus on the most foundational aspects of TypeScript. This content might be better suited for an alternate documentation format with fewer scope constraints.

PR #1772, opened 2021-04-25
Update link to first page of the handbook

athongsavath:
The hyperlink text matched the name of the deprecated page, so it was updated to match the current page.


orta:
Thanks!

PR #1773, opened 2021-04-25
fix: remove name of convicted terrorist from the main page; replace with generic name

sreejith-kumar:
The main page shows the name of a convicted Communist terrorist, Angela Davis. 

https://upload.wikimedia.org/wikipedia/commons/2/2d/Angela_Yvonne_Davis_Wanted_Poster.jpg

What purpose does it serve on the main page other than stoke political argument or troll people?

Let's replace it with a generic name.


orta:
Thanks for contributing, but I don't think we want this change

PR #1776, opened 2021-04-26
Add max-width to #handbook-content for wider displays

langscot:
This makes reading the documentation page much easier on wider displays as the content is in the center.

See before & after images.

Before
image

After
image


orta:
Given this a test run locally, I originally was pretty happy with letting it resize, but I think I'm also pretty convinced by how this feels. 

Not gonna promise I won't change my mind again in the future, but I think this is worth having 👍🏻 

Thanks!

PR #1782, opened 2021-04-28
Fix grammar in Everyday Types.md

saurabhjdas786:
Fix a grammatical error in the handbook at the end of the following section:
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#working-with-union-types


orta:
Thanks!

PR #1783, opened 2021-04-28
Update Everyday Types.md

milesnash:
Correct union type in examples regarding checking for null.


orta:
Thanks, cool, this looks good to me

PR #1784, opened 2021-04-28
Remove duplicate example

altearius:
It is exactly identical to the two lines immediately below it.

I am not clear on the meaning of "string string params" in this case.  I have left that as-is because I do not understand it.

### DIFF
--- a/packages/documentation/copy/en/project-config/Compiler Options.md
+++ b/packages/documentation/copy/en/project-config/Compiler Options.md
@@ -29,9 +29,6 @@ tsc index.js --declaration --emitDeclarationOnly
 
 # Emit a single .js file using string string params
 tsc app.ts util.ts --target esnext --outfile index.js
-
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/project-config/Compiler Options.md
```diff
@@ -27,9 +27,6 @@ tsc --project tsconfig.production.json
 # Emit d.ts files for a js file with showing compiler options which are booleans
 tsc index.js --declaration --emitDeclarationOnly
 
-# Emit a single .js file using string string params
-tsc app.ts util.ts --target esnext --outfile index.js
-
 # Emit a single .js file using string string params
```
orta:
```suggestion
# Emit a single .js file from two files via compiler options which take string arguments
```

orta:
Updated, thanks

PR #1789, opened 2021-05-01
Update More on Functions.md

saurabhjdas786:
Section: https://www.typescriptlang.org/docs/handbook/2/functions.html#type-parameters-should-appear-twice

The phrase **"function doesn't need to be generic"** may imply to someone as a **thumb rule**.
As per the topic, changing the sentence to a more **general** scenario.


orta:
Agree, thanks

PR #1792, opened 2021-05-03
Explicitly set cursor: pointer on like/dislike buttons.

### LINKED ISSUE #1791: "Is this page helpful?" buttons have no custom hover cursor
image

If I hover over either button, neither has `cursor: pointer` set on it, so it is unclear whether the buttons are functional.

DanielRosenwasser:
I think this fixes #1791, but I'm not sure because I can't get the site to build.

### DIFF
--- a/packages/typescriptlang-org/src/templates/documentation.scss
+++ b/packages/typescriptlang-org/src/templates/documentation.scss
@@ -118,6 +118,12 @@
         }
       }
     }
+
+    .like-dislike-subnav {
+      #like-button, #dislike-button {
+        cursor: pointer;
+      }
+    }
   }
 
   .preamble a {


orta:
I'd bet that works - and I'm _very_ interested in the site not building ( I have a windows CI running for it )

DanielRosenwasser:
Just realized I didn't do this for `page-helpful-popup` too 🤦‍♂️

orta:
They're styled much more like buttons and have a hover effect, I'm not sure they need it (given that buttons don't normally have the hand cursor) 

Screen Shot 2021-05-05 at 10 22 32 AM


Screen Shot 2021-05-05 at 10 22 35 AM

PR #1794, opened 2021-05-04
Show how to bind `this` in a function

wildlyinaccurate:
The current documentation doesn't actually show how to bind `this` in a function, because it's hidden in the `DB#filterUsers` type signature. While the extra typing in my change is technically redundant, it at least clearly shows in the documentation how to bind `this`.

I also removed the joke-y `this` reference. The sudden change in tone halfway through a technical explanation was confusing.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/More on Functions.md
+++ b/packages/documentation/copy/en/handbook-v2/More on Functions.md
@@ -552,7 +552,7 @@ const user = {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
I'm happy with the code changes, 👍🏻 but I think we'll keep the `this` joke.

wildlyinaccurate:
@orta I'll reinstate the `this` joke, but please consider that it's difficult to read. The rest of this document is extremely technical, and all preceding (and following) uses of `monospace font` are used to indicate to the reader that certain words are related to the code examples. The way it reads to me is (my thoughts in parentheses):

> (The keyword) `this`, _heh_ (is this a typo?), can be enough for a lot of cases (if `this` is enough for "a lot of cases", what do I use instead of `this` for the other case?)

Sorry, I'm not trying to be a killjoy. I just think technical documentation is not the most appropriate place for word play.

orta:
You're not wrong, but we need to keep some whimsy in the docs or else it will end up completely dry which also has its own downsides

PR #1796, opened 2021-05-04
Clarify namespace import behavior with esModuleInterop

khuttun:
Add note about the treatment of own/inherited properties with namespace
import when `esModuleInterop` is `true`.


khuttun:
I think it would be good to explicitly mention this in the documentation as a) enabling `esModuleInterop` is ["highly recommended"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop) b) it might break code when converting module from JS to TS. See https://github.com/microsoft/TypeScript/issues/29687 for discussion about the behavior.

orta:
Cool, I've amended it to make sense for folks who don't know what an owned property is, thanks

PR #1811, opened 2021-05-06
Fix a few small typos

equinox:
Correct a few "it's" to "its" and also adds a missing space.


orta (approved):
Thanks looks good to me

PR #1812, opened 2021-05-07
Invalid argument for `--target`

nadropper:
In the *Downleveling* section, an invalid argument (`es2015`) is passed for the `--target` flag, which results in an error: **`error TS6047: Argument for '--target' option must be 'ES3', 'ES5', or 'ES6'.`**

Also, the file to transpile is wrongly referenced as `input.ts` instead of `hello.ts`

image


orta:
Hi, thanks for the PR - I'd recommend using [yarn over npm](http://orta.io/notes/js/yarn-vs-npm) because you've accidentally been using the tsc npm module [instead of typescript](https://twitter.com/orta/status/1390592772143472643)

PR #1819, opened 2021-05-12
Links to cross-origin destinations are safe

jcabak:
Links to cross-origin destinations are safe (added `rel="noopener”`)


orta:
Thanks, but this isn't needed to my knowledge: https://mathiasbynens.github.io/rel-noopener/

> 2021 update: Browsers now implicitly set rel=noopener for any target=_blank link, following a spec change. If the demo on this page no longer seems scary, congratulations — you’re using a modern browser!

PR #1821, opened 2021-05-12
Add glob examples for excludeDirectories

amcasey:
Turns out globs are allowed.

I don't know whether we want to document it, but the patterns are resolved relative to the current directory.  This is unlikely to affect anyone who isn't forking their own process.


orta:
Agree, I took the current example from the compiler's source code but I think showing the glob is a great call

PR #1823, opened 2021-05-14
Retain custom query URL params when the URL is updated

orta:
When updating the text in a playground, then custom URL params which aren't controlled by the playground/sandbox used to get dropped - this isn't great as installing a plugin may want to read those params after, or you might want to share a URL which would include that info to others.


orta:
These fails look like a GH hiccup

PR #1824, opened 2021-05-14
Update Generics.md

Alekshir:
In order to prevent compiler errors
"error TS2564: Property 'hasMask' has no initializer and is not definitely assigned in the constructor.
Property 'nametag' has no initializer and is not definitely assigned in the constructor.
error TS2564: Property 'numLegs' has no initializer and is not definitely assigned in the constructor.
error TS2564: Property 'keeper' has no initializer and is not definitely assigned in the constructor.
"


orta:
Seems reasonable, we do assume strict everywhere. Can you please add spaces around the `=`s so it matches with the rest of the code 👍🏻

PR #1828, opened 2021-05-15
documentation/handbook-v2: Typo at Anonymous Functions section of Everyday Types v2 documentation

mmkvdev:
In the Everyday Types Documentation Handbook, at the anonymous functions section, the `contextual typing` definition has a grammatical mistake. 

**url**: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#anonymous-functions

**It says** : 
This process is called _contextual typing_ because the _context_ that the function occurred `in` informed what type it should have.

**Fix**:
the `in` highlighted in the above sentence, should be `is`. which will make much sense in the implementation perspective.


orta:
Thanks for the PR, but I think the previous version is more accurate

mmkvdev:
> Thanks for the PR, but I think the previous version is more accurate

:+1:

PR #1829, opened 2021-05-16
Update TypeScript Tooling in 5 minutes.md

salmantec:
Replaced normal string concat with JS template literal in example snippet of TypeScript Tooling in 5 minutes.md file.


orta:
Thanks, but I think we'll keep it as it is, less concepts to learn

PR #1836, opened 2021-05-18
Improve readability with emojis

maedi:
Don't and Do look similar when skimming a page and it's not obvious that each pair of Don't/Do paragraphs are a companion piece. Ideally they would be red and green to enhance readability and differentiate. Since markdown isn't meant for adding colours and inline CSS would be messy, I propose tick and cross emojis. They actually look quite good on the final site, I'll post a screenshot. Additionally I think **bold** is better than _italic_ to emphasise the Do and Don't text itself.

Screen Shot 2021-05-18 at 2 33 40 pm


orta:
I agree, thanks!

maedi:
Cheers :)

PR #1838, opened 2021-05-18
Restore the tuple documentation in the handbook

### LINKED ISSUE #1835: Docs: Tuple documentation dropped from the hand book.
**Page URL:** https://github.com/microsoft/TypeScript-Website/blob/v2/packages/documentation/copy/en/handbook-v2/Everyday%20Types.md#arrays

**Issue:** The current documentation says "Note that [number] is a different thing; refer to the section on tuple types" but there is no tuple section.  

The first version of the handbook has a tuple section: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/documentation/copy/en/handbook-v1/Basic%20Types.md#tuple

**Recommended Fix:** Add a tuple section to the current version of the handbook.

edhager:
Fixes #1835.

The section on tuples was dropped in version 2 of the handbook.  This restores that section.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -32,7 +32,30 @@ To specify the type of an array like `[1, 2, 3]`, you can use the syntax `number
 You may also see this written as `Array<number>`, which means the same thing.
 We'll learn more about the syntax `T<U>` when we cover _generics_.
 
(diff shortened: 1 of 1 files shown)


orta:
Thanks for the PR - we have the docs on Tuples in object types alongside the Array types: https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types - so this would duplicate the tuple docs. I think they're pretty reasonable to keep there.

edhager:
@orta Thanks.  I missed it in the object type section.

PR #1841, opened 2021-05-19
revise greet() examples in Object Types.md

timohaver:
I revised the (first three) function greet() examples to return person.name instead of person.age. It seems more intuitive to "greet" a person by name than age. ("Hello Mary" rather than "Hello 27")

Without knowing for certain, it seems plausible that this was the intent of the author and perhaps person.age was used instead by mistake.

URL
https://www.typescriptlang.org/docs/handbook/2/objects.html


orta:
Agree, this feels better

PR #1843, opened 2021-05-20
docs: Narrowing.md

LeoniePhiline:
Add a note about TypeScript's differences in typing `Boolean("hello")` vs double negation `!!"world"`.

The first is typed as `boolean`, while the latter is typed as `true`.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -133,12 +133,12 @@ Values like
 - `undefined`
 
 all coerce to `false`, and other values get coerced `true`.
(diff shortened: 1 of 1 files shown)


orta:
Yeah, I agree, good work 👍🏻

johnjago (approved):
Good clarification! I didn't know about this until I saw this pull request, probably because it's more often that I type

```ts
const y: boolean = !!"world";
```

instead of 

```ts
const y: true = !!"world";
```

LeoniePhiline:
One could argue this is a TypeScript bug. 
It should be just as easy to infer `true` from `Boolean("hello")`.

orta:
I wonder if that's been explored already (it'd probably take a few overloads on the Boolean constructor to make it work) but the docs should be accurate to what TS does today (and I went back to 3.6 to test which has the same behavior)

PR #1844, opened 2021-05-22
docs: Everyday Types

BagchiMB:
The point of this example was to demonstrate functions that only accept a certain set of known values, not showing some sort of typo mistake, therefore I think it is better to show 'top' rather than 'centre' which resembles general world use cases in a better manner.


orta:
Thanks, but this is a very real use-case, as centre is how it would have been done in British English 

https://www.grammarly.com/blog/center-centre/

PR #1854, opened 2021-05-26
Fix typo in Mapped Types

smellyshovel:
Make the code in the code block correspond to its description below.


orta:
Thanks, but the typo is the one in the paragraph (because the other code blocks also uses `OptionsFlags`

smellyshovel:
Screenshot_20210528_232315

This is what I mean. I think we should either change `OptionsFlags<Type>...` in the code block to be `OptionFlags<Type>...` or change the "In this example, `OptionFlags` will take" text to be "In this example, `OptionsFlags` will take" so that the description of the code block could correspond the code block contents.

PR #1861, opened 2021-05-30
Reduce prominence of keyof

Gerrit0:
Mapped types don't have to use keyof, and implying they do is confusing.


orta:
Yeah, alright, I think that's reasonable enough - given that you could have all sorts of ways to get that union made it's good to show that it's not a thing you can only use with `keyof`

PR #1862, opened 2021-05-30
Remove typeof in Spritable example

Oblosys:
The parameter `T` is the return type of the constructor, so it should be an object instead of a class.


Oblosys:
Relevant StackOverflow question: https://stackoverflow.com/questions/67762619/typescript-mixin-constrained-to-typeof-class

SMAKSS (approved):

orta:
Thanks! Lets get this in

PR #1866, opened 2021-05-31
fix const enum output

fkhadra:
This PR fixes the generated output for `const enum`.


orta:
That code sample has `// @showEmit` which means it will output the JS, so I don't think this PR needs to be merged

fkhadra:
You're right, thanks for your feedback. Don't know how I ended up on a page where it was not giving the right output. I'll close the PR.

PR #1868, opened 2021-06-01
fix content: changed code to represent context

alirezaghey:
text implies that ts code from above is compiled to js, while it's still the same ts code


orta:
Thanks! But this code sample had `showEmit` which means it outputs the emitted JS, we want to always rely on the compiler to give this kind of info - so we don't need this PR.

alirezaghey:
I see. I didn't know the typescript code is compiled on the fly. This is much more consistent than manually doing it.

I don't know why but on my end it was showing the uncompiled typescript, without syntax highlighting. It confused me because the following points didn't make sense. As much as I tried now, I was not able to reproduce it. Anyway, thanks.

PR #1877, opened 2021-06-07
Fix `OddNumbersUnderTen` to `PositiveOddNumbersUnderTen`

tjjfvi:
There are infinitely many `OddNumbersUnderTen` 😉


orta:
Heh, why not

PR #1878, opened 2021-06-08
Adds es5 to the module list

orta:
This was missing


orta:
Or not, should be es6

> Invalid value es5 for module. Allowed values: none,commonjs,amd,system,umd,es6,es2015,es2020,esnext

PR #1881, opened 2021-06-10
Adjust "The in operator narrowing" heading level

AliRoshaniDev:
"The in operator narrowing" header must be one step larger to match the other headers on this page. Otherwise the user thinks it is a subheader.


orta:
Agree! Thanks

PR #1882, opened 2021-06-10
Adjust "Enums" + "Less Common Primitives" heading

Arthie:
On the Everyday Types page the section of "null and undefined" also includes "Enums" + "Less Common Primitives" as sub-sections. Is this intentional? The "null and undefined" section title doesn't really make sense for these sub-sections.

I propose moving the  "Enums" + "Less Common Primitives" to separate sections.


orta:
Agree, thanks!

PR #1886, opened 2021-06-11
Adjust three headings on "The Basics" page

AliRoshaniDev:
These headings ("Emitting with Errors" , "noImplicitAny" , "strictNullChecks") must be one step larger to match the other headers on this page. Otherwise the user thinks these are sub-section.


orta:
Cool, thanks

PR #1887, opened 2021-06-13
Document separate getter/setter types

shalvah:
TypeScript 4.3 now supports separate getter/setter types, so I've updated the handbook on classes to reflect that.

I'm not sure what the convention is for referencing changes introduced in a specific version (eg using a "Since x.x" badge, or linking to the release post, or not mentioning the version at all), so I've just gone with a link to the release announcement.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Classes.md
+++ b/packages/documentation/copy/en/handbook-v2/Classes.md
@@ -248,14 +248,33 @@ class C {
 
 TypeScript has some special inference rules for accessors:
 
-- If no `set` exists, the property is automatically `readonly`
-- The type of the setter parameter is inferred from the return type of the getter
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Classes.md
```diff
@@ -248,14 +248,33 @@ class C {
...
+    _size = 0;
+
+    get size(): number {
+        return this._size;
+    }
+
+    set size(value: string | number | boolean) {
+        let num = Number(value);
+
+        // Don't allow NaN and stuff.
```
orta:
```suggestion
        // Don't allow NaN, Infinity, etc

```

orta:
Thanks, yeah, this is good 👍🏻

PR #1888, opened 2021-06-16
Uses the un-minified versions of typescript on nightly and pr builds

orta:
I jammed with @kingwl on getting better stack traces for PR and Nightly builds of TypeScript in the playground

Before:

Screen Shot 2021-06-16 at 10 09 05 AM

After:
MicrosoftTeams-image (7)


Kingwl:
Thanks for the quick response!

PR #1892, opened 2021-06-21
Update Project References.md

chipit24:
On the _Project References_ page of the documentation (https://www.typescriptlang.org/docs/handbook/project-references.html), I found that I was converting the list of absolute files paths in my mind to a folder structure of the project. I figured that using a tree structure directly to represent the project layout would be more natural for others as well–so here is a PR for it!

I'm sure there's many ways to show a project/directory structure (screenshots, adding some emojis in there, etc.), but through the power of Google and Stack Overflow, I came across https://tree.nathanfriend.io and it generated a pretty nice layout, so I used that.

This is my first PR here so a pre-emptive apology if I've missed some process! 😬


orta:
Yeah, this looks great - I normally use the command [tree](https://en.wikipedia.org/wiki/Tree_(command)) to generate this. Process is fine 👍🏻

PR #1896, opened 2021-06-22
Adjust two headings on "Everyday Types" page

AliRoshaniDev:
These headings ("bigint" , "symbol") must be one step larger


orta:
Thanks!

PR #1898, opened 2021-06-22
Update caveats section for the project references

Andarist:
From my understanding the proposed changes are correct but I'm no expert so please correct me if I'm wrong.

### DIFF
--- a/packages/documentation/copy/en/project-config/Project References.md
+++ b/packages/documentation/copy/en/project-config/Project References.md
@@ -125,7 +125,8 @@ It's important in this situation to not prepend at each reference, because you'l
 Project references have a few trade-offs you should be aware of.
 
 Because dependent projects make use of `.d.ts` files that are built from their dependencies, you'll either have to check in certain build outputs _or_ build a project after cloning it before you can navigate the project in an editor without seeing spurious errors.
-We're working on a behind-the-scenes .d.ts generation process that should be able to mitigate this, but for now we recommend informing developers that they should build after cloning.
+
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/project-config/Project References.md
```diff
@@ -125,7 +125,8 @@ It's important in this situation to not prepend at each reference, because you'l
 Project references have a few trade-offs you should be aware of.
 
 Because dependent projects make use of `.d.ts` files that are built from their dependencies, you'll either have to check in certain build outputs _or_ build a project after cloning it before you can navigate the project in an editor without seeing spurious errors.
-We're working on a behind-the-scenes .d.ts generation process that should be able to mitigate this, but for now we recommend informing developers that they should build after cloning.
+
+When using VScode (since TS 3.7) we have a behind-the-scenes in-memory .d.ts generation process that should be able to mitigate this, but it has some perf implications. For very large composite projects you might want to disable this using [disableSourceOfProjectReferenceRedirect option](https://www.typescriptlang.org/tsconfig#disableSourceOfProjectReferenceRedirect).
```
orta:
```suggestion
When using VS Code (since TS 3.7) we have a behind-the-scenes in-memory .d.ts generation process that should be able to mitigate this, but it has some perf implications. For very large composite projects you might want to disable this using [disableSourceOfProjectReferenceRedirect option](/tsconfig#disableSourceOfProjectReferenceRedirect).
```

orta:
Cool, thanks, that's right - looks good to me

PR #1900, opened 2021-06-23
fix missing keys console error on home page

rpivo:
Fixes the React console error alerting to missing keys for a list of children


RyanCavanaugh:
This is not a correct way to make a key in React.

https://legacy.reactjs.org/docs/reconciliation.html#trade-offs

PR #1903, opened 2021-06-25
update classes caveat documentation with more info on # private fields

rpivo:
Updates documentation in relation to: https://github.com/microsoft/TypeScript/issues/44670


orta:
This is great work, well done @rpivo

PR #1910, opened 2021-06-30
Update Creating DTS files From JS.md

staff0rd:
- fix npx command


orta:
Thanks, yeah that feels right 👍🏻

Issue #1656, opened 2021-03-11
Docs: Wikipedia-like preview popups


orta:
**Page URL:** Any documentation page

**Issue:** A link is context-less, can we provide useful context information on a mouse hover for a link which might be enough for someone to not need to click through?

![Screen Shot 2021-03-11 at 7 15 02 AM](https://user-images.githubusercontent.com/49038/110749837-87525a00-8239-11eb-85f1-dfdfa5e7179b.png)

We don't have images for pretty much all concepts we'd want to showcase, so it's closer to:

![Screen Shot 2021-03-11 at 7 16 13 AM](https://user-images.githubusercontent.com/49038/110749956-afda5400-8239-11eb-84c2-40ff8b357b06.png)

Something like this:

![Screen Shot 2021-03-11 at 7 31 29 AM](https://user-images.githubusercontent.com/49038/110751489-d39e9980-823b-11eb-8af2-2bbc75203368.png)

**Recommended Fix:** 

Take the metaphors used on wikipedia and reuse those.

Today, I can think of three different sources of information I think we could/should pull:

 - TSConfig settings info
 - Documentation pages
 - MDN Links

I think TSConfig info is probably the easiest to get started with, as it already is a well formed json dump in the codebase: `packages/tsconfig-reference/output/en-summary.json`.

**Implementation Notes:** 

For the TSConfig data. This data needs to be looked up at runtime, so we probably need to have the JSON dump  exposed in some format as a faux API (aka a JSON file, or a [JSONP](https://stackoverflow.com/questions/2067472/what-is-jsonp-and-why-was-it-created) script file. 

Then in `documentation.tsx` we need to register hovers on `a` links in an side effect which checks if the link is to /tsconfig/#flag - it pulls out the flag, checks if that's in the API, if so, wait to be sure the mouse is still hovering on the same element after ~0.5s then show the popover

I don't think we _need_ the hover when you tab through anchors, but it's probably a nice to have.

orta:
### MDN implementation ideas

I think we'll need a build step in the `documentation` workspace, which first extracts a list of all developer.mozilla.org URLs which the markdown documentation references. From that list, we use a cloned copy of the https://github.com/mdn/data/ repo and look up each file by their url (it's in their yml)

[...]

PR #1911, opened 2021-06-30
Add wikipedia like popups

alirezaghey:
Adds popups for tsconfig links
Popup positioning is not ideal
Related to #1656


orta:
This is close! I want to auto-generate the JSON object and then we're good

orta:
There we go - thanks @alirezaghey !

alirezaghey:
Thanks @orta, this looks amazing! I'll look into how we can add MDN and internal documentation pages according to #1656 and also how we can make the popups aware of their position on the page so that they move a little bit to the opposite side like how they do on wikipedia.

orta:
I've left some of my ideas on the MDN and docs pages, the tricky thing is making a markdown summary but I've included a pretty reasonable idea for how that could happen.

Re: popup positioning, that's doable - it's just JavaScript and deciding which side is closest to the middle of the page and using that edge to position the popover instead of right now where it's the bottom always. I'd start with maybe doing either above or below first, and then look at left or right

PR #1913, opened 2021-06-30
Update gatsby/react and switch to the lib version of twoslash

orta:
No compelling reason to upgrade Gatsby/React, but I gave a bunch of pages a run through and I trust the build/test infra. 

The Twoslash update I did want to do, there's been a bunch of perf wins over on the old version which used to live in this repo library which I want


orta:
Fixed reds with https://github.com/gatsbyjs/gatsby/pull/32509

PR #1917, opened 2021-07-01
Case adjust

lguima:
According to its own documentation, the `boolean` type should be in lowercase, right?

> The type names `String`, `Number`, and `Boolean` (starting with capital letters) are legal, but refer to some special built-in types that will very rarely appear in your code. _Always_ use `string`, `number`, or `boolean` for types.


orta:
Thanks! Yeah, that's a typo

PR #1921, opened 2021-07-03
Updated ASP.NET Core.md

MahmudX:
Without returning an async completion signal, the Task Runner throws an error like below - 

```
[16:34:30] Starting 'default'...
[16:34:30] The following tasks did not complete: default
[16:34:30] Did you forget to signal async completion?
Process terminated with code 1.
```

We can fix this issue by invoking the default callback function.


orta:
Yeah, this fits with the docs where they use the callback everywhere, likely just gulp moving forwards and our docs not keeping up: https://gulpjs.com/docs/en/api/task/

orta:
Thanks

PR #1923, opened 2021-07-06
Remove unnecessay interface Dog declaration

ledenis:
I think the `interface Dog` is unnecessary in this snippet. It is actually confusing. It is only needed in the next snippet.


orta:
Agree, thanks

PR #1924, opened 2021-07-06
Improve documentation for useUnknownInCatchVariables

iShibi:
This PR fixes the grammar in the description of `useUnknownInCatchVariables` option.

I wanted to fix the rendering issue too **(the backticks in the first line)** but wasn't able to find the cause:

TSConfig Reference - Docs on every TSConfig option

ℹ️ A slightly smaller issue: shouldn't it show `(error: unknown)` without any need to hover over it?

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/useUnknownInCatchVariables.md
+++ b/packages/tsconfig-reference/copy/en/options/useUnknownInCatchVariables.md
@@ -18,4 +18,4 @@ try {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


orta:
Sure, looked reasonable - if you want to show an inline type, use:


```
// ^?
```

To point to `err` on its own line 👍🏻

iShibi:
I see but wouldn't that show the type in a tooltip? Since, we are showing what user had to do prior to `v4.4` to make type of `err` `unknown`, It should show it like this:
```ts
try {
  // ...
} catch (err: unkown) {
  // We have to verify err is an
  // error before using it as one.
  if (err instanceof Error) {
    console.log(err.message);
  }
  ```
  This would clearly convey what the para after that codeblock states: "no need to explicitly write `(: unknown)` if the `useUnknownInCatchVariables` option is `true` in `v4.4`".

PR #1929, opened 2021-07-10
fix: update type acquisition to correctly handle imports nested in index files

### LINKED ISSUE #1926: Playground: type discovery broken for certain re-export patterns
In some cases when doing `export * from ./some-folder` type discovery looks for files in the wrong directory.

Simple playground example: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwIIF84DMoRHA5ACwEMBPA

The module being loaded has 4 files.  Each index file just does a `export * from` for the next file in the list, with the final file exporting a single type `A`.  

* index.ts
* types/index.ts: 
* types/utils/index.ts
* types/utils/a.ts

When the playground is discovering types for this module, it looks for the utils file in  at https://www.unpkg.com/hayes-type-discovery-bug@1.0.1/lib/util.d.ts and /hayes-type-discovery-bug@1.0.1/lib/util/index.d.ts instead of looking in the types directory.  I think this example has 1 more level of nesting than is required to show the issue, but is still a fairly simple reproduction.

Full code for the test package is here: https://github.com/hayes/type-discovery-bug

Files in the imported module:
Screen Shot 2021-07-09 at 10 41 19 PM
playground loading utils file from the wrong directory:
Screen Shot 2021-07-09 at 10 39 59 PM

**Repro:**

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwIIF84DMoRHA5ACwEMBPA

**Expected behavior:**

type discovery should look for /lib/types/util/index.d.ts instead of /lib/util/index.d.ts
**Actual behavior:**

playground attempts to load /lib/util/index.d.ts

hayes:
fixes #1926 

screenshot of fixed repro from #1926
Screen Shot 2021-07-10 at 12 08 58 PM

### DIFF
--- a/packages/sandbox/src/typeAcquisition.ts
+++ b/packages/sandbox/src/typeAcquisition.ts
@@ -169,12 +169,17 @@ const addModuleToRuntime = async (mod: string, path: string, config: ATAConfig) 
   let content = await getCachedDTSString(config, dtsFileURL)
   if (!content) {
     const isDeno = actualPath && actualPath.indexOf("https://") === 0
+    const indexPath = `${actualPath.replace(".d.ts", "")}/index.d.ts`
 
-    const dtsFileURL = isDeno ? actualPath : unpkgURL(actualMod, `${actualPath.replace(".d.ts", "")}/index.d.ts`)
+    const dtsFileURL = isDeno ? actualPath : unpkgURL(actualMod, indexPath)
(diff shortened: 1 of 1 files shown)


orta:
Thanks, sorry, I was out last week - yeah, I've given this a run locally and everything seems to be working fine - looks like we just missed the reassignment for this case. Good spot!

PR #1934, opened 2021-07-13
More on Functions - add bigint to ECMAScript primitives enumeration

Alexsey:
add `bigint` to the list of ECMAScript primitives list


orta:
Thanks!

PR #1935, opened 2021-07-14
Fix Markdown in Object Types

Alexsey:
Because of a [missing empty line after \<summary\> tag](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab), `code` formatting inside of collapsible section was not applying and the text was shown as is, with ` symbols


orta:
Agree, thanks!

Issue #1919, opened 2021-07-03
Playground: `strict=true` in urls doesn't work


DetachHead:
![image](https://user-images.githubusercontent.com/57028336/124347305-5b634600-dc27-11eb-83ec-a4b29701ae99.png)
![image](https://user-images.githubusercontent.com/57028336/124347315-68803500-dc27-11eb-8384-12b3d6399733.png)

**Repro:** https://www.typescriptlang.org/play?strict=true&ts=4.4.0-dev.20210702#code/C4JwngBA3gsAUPAvhAxgQ2CgFgCgKYCU08EpqA9gHYDOwEAltQMqj2UDmEAvBMGAA55yAMwh5uXHgHJaINuykky9UTkYs5HIrDhk9YgHTByAVX6CQAYTTU8OIgHoHEAPIAjAFZ4UdRhBG8AuJSAK6UANaU5ADulIq6ZIhKpCoQOHyCAeKS0rLyUtrJ+nhGpuZ4VjZ2js5RYiAg5CBFSXCIQA

**Expected behavior:**
`strict` flag is enabled when it's in the url
**Actual behavior:**
`strict` flag is not enabled

orta:
strict is default in the playground, this could be a UI bug

orta:
probably related #1922

PR #1940, opened 2021-07-19
Use the compiler options as a backup for finding compiler options from inside the param handling code

orta:
Prior versions of the playground have just relied on a big 'defaults dump' to use as a reference about what compiler options are recognized from the url query. This works well enough _if you are completely on top of all compiler options and updating the website, but I've been doing a lot of work outside of the website and it's a bad call to rely on that in the future. 

https://github.com/microsoft/TypeScript-Website/blob/66d7d8bdf910c5deffc7c39f584addd9a5ec33ce/packages/sandbox/src/compilerOptions.ts#L10-L17

This PR re-uses the same idea as https://github.com/microsoft/TypeScript-Website/pull/1735 which is that we have the TS compiler in the runtime, so we should use that to validate whether a param a legit compiler flag even if it's not available in playground the defaults.


orta:
and includes a fix  for https://github.com/microsoft/TypeScript-Website/issues/1919

PR #1941, opened 2021-07-19
Docs: Handbook v2: Classes: Inheriting Built-in Types

Alexsey:
Extend the notes of the section that it may be skipped not only if the reader is not going to inherit from build-in types but also for compilation target above ES5


orta:
Thanks, makes sense 👍🏻

PR #1946, opened 2021-07-21
Typo in "Cross-hierarchy protected access" section

jesseplymale:
This section discusses access of base members by subclasses. Access via a Derived2 reference is *legal* (no errors in this example), but via Derived1 reference is illegal.


orta:
Agree! Thanks

PR #1954, opened 2021-07-23
Fixed a typo in a public `extractTwoSlashCompilerOptions` method

Andarist:


orta:
Hah, oops! Thanks @Andarist

Andarist:
@orta the CI has failed here but it doesn't seem to be related to this PR

packages/sandbox/src/index.ts
```diff
@@ -409,8 +409,14 @@ export const createTypeScriptSandbox = (
     lzstring,
     /** Returns compiler options found in the params of the current page */
     createURLQueryWithCompilerOptions,
+    /**
+     * @deprecated Use `getTwoSlashCompilerOptions` instead.
```
Andarist, line 413: I have initially thought that `@typescript/sandbox` is a published package so I wasn't sure if you are OK with introducing a "breaking" change here but it seems to only be a local package so I think that removing the "deprecated" variant altogether would be preferred here. Thoughts?

orta: It's exposed to playground plugins and folks using the sandbox via https://www.typescriptlang.org/dev/sandbox/ - so this is the right call 👍🏻

orta:
I fixed CI via another repo this morning, so a re-run should merge 👍🏻

PR #1956, opened 2021-07-23
Drop redundant `detectNewImportsToAcquireTypeFor` call

Andarist:
This is supposed to acquire types for the initial value but the same thing already happens within this call:
https://github.com/microsoft/TypeScript-Website/blob/23cd21aec8ef8284d3cb89fad6607e839c2f97e8/packages/sandbox/src/index.ts#L345


orta:
Cool, thanks!

Andarist:
@orta the CI has failed here but it doesn't seem to be related to this PR

orta:
Re-ran after fixing this morning

PR #1958, opened 2021-07-24
Fix spelling

dcblair:
- Pluralized objects and removed hyphen from rename on Line 459.
previous: "...declare the shapes of object, not re-name primitives"
current: "...declare the shapes of objects, not rename primitives"


orta:
Agree, thanks

PR #1960, opened 2021-07-24
update basic type link

avedetvedea:
Basic type link was broken and further the intended "basic-type" doc page was deprecated. So replaced broken link with the redirected link of mentioned depreciated.


orta:
Thanks!

PR #1972, opened 2021-08-05
Union primary definition may refer values rather than types

alperenbelgic:
With this change the primary definition seemed to me more natural. Because if you see line 147, it specifically refers how you 
 combine different "types". But, since I am a learner and missing important portion of concepts, the change may not be relevant.


orta:
Thanks for the PR but yeah, this isn't right - you're maybe getting confused with how literal types work. `type A = 1 | 2` is still a union of two types.

PR #1977, opened 2021-08-09
Update Compiler Options.md

timthedev07:
Remove one redundant line and fix incorrect table header display.


orta:
Close! This table is generated here: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/tsconfig-reference/scripts/cli/generateMarkdown.ts#L60 (probably just needs to move that inline?

timthedev07:
Thanks, I'll take a look at it.

timthedev07:
Got the RP up, can you take a look.

PR #1980, opened 2021-08-10
Update Namespaces and Modules.md

Qwebeck:


packages/documentation/copy/en/reference/Namespaces and Modules.md
```diff
@@ -70,7 +70,7 @@ This is how the `node.d.ts` file that several of the TypeScript samples use is c
 
 ## Needless Namespacing
 
-If you're converting a program from namespaces to modules, it can be easy to end up with a file that looks like this:
+If you're converting a program from modules to namespaces, it can be easy to end up with a file that looks like this:
```
Qwebeck: There are `namespaces` in the example result file, so I assume, that it is the result of conversion from `modules` to `namespaces`

Qwebeck: Looks like my mistake.
Instead of this sentence, example should be fixed

packages/documentation/copy/en/reference/Namespaces and Modules.md
```diff
@@ -75,7 +75,7 @@ If you're converting a program from namespaces to modules, it can be easy to end
 - `shapes.ts`
 
   ```ts
-  export namespace Shapes {
+  declare module Shapes {
```
Qwebeck: Later in this section, `Shapes` is referenced as a module.
obraz

orta:
Thanks, the typo was underneath!

PR #1994, opened 2021-08-19
Cross-reference utility types -> release notes

jablko:
What do you think about cross-referencing the utility types to the release notes, like how the [TSConfig options are](https://www.typescriptlang.org/tsconfig#extends)?

When I encounter errors like the following, it usually takes me a sec to find the version I should bump `// Minimum TypeScript Version` to (in this case the version that introduced `Omit`):
```Shell
$ dtslint types
Error: Errors in typescript@3.4 for external dependencies:
../../../node_modules/preact/src/jsx.d.ts(310,6): error TS2304: Cannot find name 'Omit'.
```

This information was previously in [the v1 handbook](https://web.archive.org/web/20200803014021/https://www.typescriptlang.org/docs/handbook/advanced-types.html#:~:text=typescript%202.8%20adds%20several%20predefined%20conditional%20types) ... at least for some predefined conditional types.

### DIFF
--- a/packages/documentation/copy/en/reference/Utility Types.md
+++ b/packages/documentation/copy/en/reference/Utility Types.md
@@ -10,6 +10,13 @@ TypeScript provides several utility types to facilitate common type transformati
 
 ## `Partial<Type>`
 
+<blockquote class=bg-reading>
+
+Released:  
+[2.1](/docs/handbook/release-notes/typescript-2-1.html#partial-readonly-record-and-pick)
+
+</blockquote>
+
 Constructs a type with all properties of `Type` set to optional. This utility will return a type that represents all subsets of a given type.
 
 ##### Example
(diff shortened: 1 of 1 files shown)


orta:
I like it 👍🏻 

Style-wise, you could have it like these: https://www.typescriptlang.org/docs/handbook/2/classes.html#constructors

With this syntax instead:

```
<blockquote class='bg-reading'>
  <p>Background Reading:<br /><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes'>Classes (MDN)</a></p>
</blockquote>
```

Which might feel more 'native' to the page

jablko:
@orta Thanks!

> Style-wise, you could have it like these: https://www.typescriptlang.org/docs/handbook/2/classes.html#constructors

Good idea, done. :heavy_check_mark:

orta:
👍🏻

PR #1995, opened 2021-08-20
firstElement examples should take into consideration undefined

jvissers:
`function firstElement<T>(arr: T[]): T `

Started experimenting (n00b alert) with TS (4.3.5) and the above example does not compile. It gives me:

```
Type 'T | undefined' is not assignable to type 'T'.
  'T' could be instantiated with an arbitrary type which could be unrelated to 'T | undefined'. ts(2322)
```

There are two alternatives I think:

1. leave the function result T and let TS inference deal with it
2. specified T | undefined as function result

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/More on Functions.md
+++ b/packages/documentation/copy/en/handbook-v2/More on Functions.md
@@ -103,21 +103,23 @@ In TypeScript, _generics_ are used when we want to describe a correspondence bet
 We do this by declaring a _type parameter_ in the function signature:
 
 ```ts twoslash
-function firstElement<Type>(arr: Type[]): Type {
+function firstElement<Type>(arr: Type[]) {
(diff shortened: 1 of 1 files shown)


orta:
You have https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess turned on, which we don't have on by default in the website code samples - given that this is the first code sample where we introduce generics in the handbook I think we want to keep it how it is (because it is more explicit) vs removing the return type.

jvissers:
Wow - that was quick :)

Your call, again I'm really new at this. For what it is worth, I think (then) the most explicit way to write this would be to **have** `T | undefined`

orta:
Yeah, that I'm game for 👍🏻

orta:
Wonder what was up with CI, but wither way lets get this in

PR #1996, opened 2021-08-20
Link to Tuple Types section of Object Types page

lrowe:
A search for 'tuple' on the website has the deprecated handbook basic types page as the only documentation result. That deprecated page links back to this one so there was is no way to find the current Tuple Types documentation in Object Types from the website.

Unfortunately a search for 'tuple types' shows no documentation results. Perhaps documentation sections (rather than just pages) could be indexed in Algolia.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -32,7 +32,7 @@ To specify the type of an array like `[1, 2, 3]`, you can use the syntax `number
 You may also see this written as `Array<number>`, which means the same thing.
 We'll learn more about the syntax `T<U>` when we cover _generics_.
 
-> Note that `[number]` is a different thing; refer to the section on _tuple types_.
+> Note that `[number]` is a different thing; refer to the section on [Tuple Types](/docs/handbook/2/objects.html#tuple-types).
 
 ## `any`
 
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Everyday Types.md
```diff
@@ -32,7 +32,7 @@ To specify the type of an array like `[1, 2, 3]`, you can use the syntax `number
 You may also see this written as `Array<number>`, which means the same thing.
 We'll learn more about the syntax `T<U>` when we cover _generics_.
 
-> Note that `[number]` is a different thing; refer to the section on _tuple types_.
+> Note that `[number]` is a different thing; refer to the section on [Tuple Types](/docs/handbook/2/objects.html#tuple-types).
```
orta:
```suggestion
> Note that `[number]` is a different thing; refer to the section on [Tuples](/docs/handbook/2/objects.html#tuple-types).

```

orta:
Thanks! Titles in the docs do show up in search "tuple/tuples" works fine, but you're right "tuple types"  doesn't - so I've removed the word "type" from that bit of docs

lrowe:
@orta, thanks for merging the fix. However on both Safari and Chrome I only see the link to the deprecated documentation (https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) when searching for `tuple`. I would have expected https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types to show up too.

image

PR #1998, opened 2021-08-24
Update target.md

imme-emosol:
Attempt to make text about `tsconfig/bases` a bit more clear.


orta:
Love it, let's do it - thanks

PR #2001, opened 2021-08-26
Make the slogan less confusing to beginners

kof:
A common misconception inside the community is that "TypeScript is JavaScript" while typescript doesn't have an exact "is" relationship with JavaScript. The rest of the documentation explains that it's a "superset" but the slogan still doesn't reflect it.

It hurts TypeScript when people expect it to be just JavaScript but then they learn that:

- many things can't be written exactly the same way because some extra checks are needed or some notation should be used for ts to infer the type correctly
- that in a non-trivial application, the types theory knowledge is a significant portion of the total language knowledge, probably much bigger than js itself
- that knowing javascript doesn't let you use typescript efficiently and you will spend a lot of time writing types
- that ts has even language features besides of types which compile to some equivalent in javascript (e.g. enums)

... and as a result they hate the language that promised them to be just JavaScript but it isn't in practice.

Screenshot 2021-08-26 at 15 52 22

### DIFF
--- a/packages/typescriptlang-org/src/copy/en/index2.ts
+++ b/packages/typescriptlang-org/src/copy/en/index2.ts
@@ -1,6 +1,6 @@
 export const indexCopy = {
   index_2_headline:
-    "TypeScript is <bold>JavaScript with syntax for types.</bold>",
+    "TypeScript extends <bold>JavaScript with syntax for types.</bold>",
   index_2_byline: "TypeScript extends JavaScript by adding types.",
   index_2_summary:
     "TypeScript is a strongly typed programming language which builds on JavaScript giving you better tooling at any scale.",


orta:
Changing the tagline isn't trivial because it needs to go through many, many stakeholders inside Microsoft. It's something we've been discussing constantly since before the first re-design 2 years ago, and realistically there isn't a tagline which satisfies everyone.

I'm going to close the PR as I have no idea how long it could take to get it merged, or if it will at all and there's not point in leaving it hanging. Consider the feedback taken though.

kof:
If there are many parties, a good way to go about it is to advertise this issue instead of closing, get some traction, get some public discussion and maybe this will convince people who said initially "TypeScript is JavaScript" it was a good idea initially but now it is simply confusing and maybe they will think "we are ready to change our minds now".

I think keeping discussions like they are fully internal isn't a great way to deal with it especially when there is no consensus and external users are confused.

PR #2003, opened 2021-08-27
Update links: remove en-us from MS Docs url

imba-tjd:
This will let the website redir to localized version.


orta:
Ace, thanks!

PR #2006, opened 2021-08-28
(docs): improved syntax in example

0xflotus:


orta:
Hi, thanks for the PR but I'm not really sure this change adds any value to the code sample (it is legit JavaScript as the [error is optional](https://2ality.com/2017/08/optional-catch-binding.html))

PR #2007, opened 2021-08-31
Should not render </tr> as code block

zartre:
Before: `</tr>` is rendered as a code block

Screen Shot 2564-08-31 at 20 51 29

After: `</tr>` is now part of the table's HTML

Screen Shot 2564-08-31 at 20 51 05


orta:
Thanks! - These files are auto-generated, so I think it needs to be handled in the `tsconfig-reference` package in this monorepo 👍🏻

zartre:
Thanks for pointing out. Will look into it 😀

orta:
Ace! Thanks this looks good to me 👍🏻

zartre:
Thanks!! 😊

PR #2008, opened 2021-08-31
Adds a note in the 4.3 release notes for the useDefineForClassFields default change

orta:
Re: https://github.com/microsoft/TypeScript/issues/45653

### DIFF
--- a/packages/documentation/copy/en/release-notes/TypeScript 4.3.md
+++ b/packages/documentation/copy/en/release-notes/TypeScript 4.3.md
@@ -774,6 +774,12 @@ This is discussed [in some detail here](https://github.com/microsoft/TypeScript-
 
 https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/991
 
+### `useDefineForClassFields` now defaults to true on `es2020` and above
+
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/release-notes/TypeScript 4.3.md
```diff
@@ -774,6 +774,12 @@ This is discussed [in some detail here](https://github.com/microsoft/TypeScript-
 
 https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/991
 
+### `useDefineForClassFields` now defaults to true on `es2020` and above
+
+In ES2020 the class fields feature was added into JavaScript with behavior which differed from how TypeScript had implemented it. In preparation for this, in TypeScript 3.7, a flag was added ([`useDefineForClassFields`](https://www.typescriptlang.org/tsconfig#useDefineForClassFields)) to migrate to emitted JavaScript to match the JavaScript standard behavior.
+
+Now that the feature is in JavaScript, we are switching to the JavaScript behavior for any projects targeting `es2020` and above.
```
KeithHenry: Should this also include that `useDefineForClassFields` is mutually exclusive with `experimentalDecorators`? It's not apparent from this description that setting both to `true` will cause emitted JS to include decorator boilerplate function and property definitions, but that extra code will do nothing.

KeithHenry: Also, the documentation at https://www.typescriptlang.org/tsconfig#useDefineForClassFields specifies `es2022` and above, but this states `es2020` and above, which is it?

orta: Yeah, good point - it is es2022 but that's not an option in TS yet. 

I'm not sure this is the point to cover how decorators works with `useDefineForClassFields`

PR #2009, opened 2021-09-01
improve an in-code comment

avedetvedea:
" // longerString is of type 'string' " statement is at least imprecise, since the type of longerString is actually OR of two string literals 'alice' and 'bob'. At the end, the comment explains a type of 'string' and the type is actually 'alice' | 'bob' hovering over it.


kleinkkk (commented):
packages/documentation/copy/en/handbook-v2/More on Functions.md

orta:
Agree, thanks

kleinkkk:
> " // longerString is of type 'string' " statement is at least imprecise, since the type of longerString is actually OR of two string literals 'alice' and 'bob'. At the end, the comment explains a type of 'string' and the type is actually 'alice' | 'bob' hovering over it.

avedetvedea:patch-2

PR #2010, opened 2021-09-01
Add description for `-readonly` modifier

38elements:
Since I can not find description for `-readonly` modifier, this adds description for `-readonly` modifier.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#readonly-mapped-type-modifiers-and-readonly-arrays


orta:
This is covered in Mapped Types: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html - we could link to that instead?

38elements:
@orta
Thank you for reply. I added links for `mapping modifiers`.

orta (approved):
Nice one!

PR #2012, opened 2021-09-02
delete duplicate code snippet

avedetvedea:
Code snippet was unnecessarily duplicated before moving to another point (reading from properties).


orta:
Agree, thanks!

PR #2013, opened 2021-09-04
fix: Improve indentation

karfau:
- to fix wrong order of `</tr></td>`
- fix `</tr> shown as code blocks (also fixed in #2007)


orta:
Thanks, but this file is auto-generated and any changes would be negated - if you look in `generateMarkdown.ts` for the CLI generator I think you'll find the place for the right change 👍🏻

karfau:
created #2014 , closing this one

PR #2016, opened 2021-09-09
Integrating with build tools: update webpack

### LINKED ISSUE #383: Webpack 4, 5 examples
Here
https://www.typescriptlang.org/docs/handbook/integrating-with-build-tools.html

only examples for Webpack 1 and 2 integration are offered. I would like to suggest that the Typescript documentation should provide examples for Webpack 4 and 5 (maybe 3?), too.

sangxxh:
Closes #383


orta:
Looks great to me - thanks

PR #2023, opened 2021-09-13
added a warning against object literal shap-matching

eugland:
added a warning against object literal shap-matching and provided code example, and error code.


orta:
Thanks for the PR. This is a bit different (it's called excess property checking) and I think it's OK for it to be explained later in the handbook 👍🏻

Issue #2025, opened 2021-09-15
Docs: A popover of Type is incorrect in Keyof Type Operator.


mymactive:
**Page URL:** 
https://www.typescriptlang.org/docs/handbook/2/keyof-types.html

**Issue:** 
The first example of keyof operator is displayed as following: 

![image](https://user-images.githubusercontent.com/60250695/133360431-26fdac5a-babd-49a1-8760-a142cafc393b.png)

It looks like reputation.

**Recommended Fix:** 
Its popover should be displayed as 'type P = "x" | "y"'.
I shall think "twolash" has the cause.

I am a beginner, so comments, reviews are welcome.
We love TypeScript and thanks to members, contributors, communities, etc... of TypeScript!

orta:
Thanks, but this is exactly what you would get in the editor (because the info in the popover comes from typescript, it's not hand-written) - TypeScript keeps around names like this as much as possible before reverting to the actual values

mymactive:
@orta 
Thanks, but I think that this example is added to comments for beginners.
Indeed, I am confused by this example.
I did not understand type P is equal to the union of key types immediately.
I suggest the following:

now:
```markdown
## The `keyof` type operator

The `keyof` operator takes an object type and produces a string or numeric literal union of its keys:

```ts twoslash
type Point = { x: number; y: number };
type P = keyof Point;
//   ^?
```

after:
```markdown
## The `keyof` type operator

The `keyof` operator takes an object type and produces a string or numeric literal union of its keys.
The following type P is the same type as "x" | "y".

```ts twoslash
type Point = { x: number; y: number };
type P = keyof Point;
//   ^?
```
```
(My mother's language is not English, so the above may be unnatural.)

orta:
Yeah, I agree, this is a good idea 👍🏻

mymactive:
Thanks for your comment!
I will do this!

agatsumayuki:
> Thanks for your comment! I will do this!

great work!

@mymactive

PR #2030, opened 2021-09-16
DOCS: Add a comment to the first example of Keyof Type Operator for beginners

mymactive:
It is related to issue #2025.
I add a short comment to Keyof type Operator. 
My mother tongue is not English, so this comment would be unnatural.
Would you point out an expression?


orta:
Thanks, this is reasonable

PR #2034, opened 2021-09-19
Make @typescript/vfs compatible with environments without localStorage

pd4d10:
For example, service worker


orta:
👍🏻

orta:
Reminder to myself that we need to pump the packagejson version to deploy this

PR #2038, opened 2021-09-23
Adds Gists powered multi-file Playgrounds

orta:
Re: https://github.com/microsoft/TypeScript-Playground-Gist-Proxy-API


orta:
Re-read the code and it feels pretty solid to me after giving it a run on a different computer 👍🏻

PR #2039, opened 2021-09-24
Update skipLibCheck.md

trusktr:
Add more explanation as to why someone would want to use skipLibCheck.

This is highly annoying: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/skipLibCheck.md
+++ b/packages/tsconfig-reference/copy/en/options/skipLibCheck.md
@@ -14,6 +14,7 @@ your `node_modules`. In these cases, you should consider using a feature like [y
 to ensure there is only one copy of that dependency in your tree or investigate how to ensure there is
 only one copy by understanding the dependency resolution to fix the issue without additional tooling.
 
-Another all-too-common scenario is during one of TypeScript's in-range breaking releases. When this happens, dependencies
-in your project's node_modules may start to have type errors that are out of your control, and `skipLibCheck` will
-be useful for silencing them.
+Sometimes TypeScript releases in-range breaking changes, and although this can sometimes causes errors in downstream projects, the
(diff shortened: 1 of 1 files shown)


orta:
I like the general idea, though the tone is definitely off

trusktr:
@orta How's that now?

packages/tsconfig-reference/copy/en/options/skipLibCheck.md
```diff
@@ -13,3 +13,8 @@ A common case where you might think to use `skipLibCheck` is when there are two
 your `node_modules`. In these cases, you should consider using a feature like [yarn's resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/)
 to ensure there is only one copy of that dependency in your tree or investigate how to ensure there is
 only one copy by understanding the dependency resolution to fix the issue without additional tooling.
+
+Sometimes TypeScript releases in-range breaking changes, and although this can sometimes causes errors in downstream projects, the
+changes help to progress TypeScript to a better state for everyone. When a breaking change hapepns, dependencies
+in a project's `node_modules` may start to show type errors, andthe `skipLibCheck` option will
+be useful for silencing these errors until an update can be performed to migrate a code base or its dependencies forward.
```
orta:
```suggestion
Another possibility is when you are migrating between TypeScript releases and the changes cause breakages in node_modules and the JS standard libraries which you do not want to deal with during the TypeScript update. 

Note, that if these issues come from the TypeScript standard library you can replace the library using [TypeScript 4.5's lib replacement](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#supporting-lib-from-node_modules) technique.
```

orta:
I've cleaned it up a bit and mentioned that you can use the 4.5 lib replacement technique 👍🏻

PR #2044, opened 2021-09-27
Drop tsdx from the recommendations

orta:
It's been pretty much abandoned at this point.


zeusdeux:
Thanks for the quick turnaround on this @orta!

nyngwang:
No, you didn't drop all of them, here is another pointer: https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-vs-tsc-for-typescript.

nyngwang:
@orta We need your help again. Here is another pointer: https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-vs-tsc-for-typescript.

PR #2049, opened 2021-09-28
Massage tsconfig default and allowed values

jablko:
Mainly what I want is to lookup the default value of [`module`](https://www.typescriptlang.org/tsconfig#module). The [old reference](https://web.archive.org/web/20190102001939/https://www.typescriptlang.org/docs/handbook/compiler-options.html) made clear that it's `es6` unless `target` is `es3` or `es5`.

While I'm at it I'd like to make the descriptions of conditional default values more consistent, instead of using prose in most cases and pseudocode (`module === "system"`) in others:
- I used lists to align the default values (e.g. `commonjs`) on the left and their conditions ("if [`target`](#target) is `es3` or `es5`") on the right.

What do you think?

I noticed some Markdown isn't parsed in the CLI reference: 
Screenshot from 2021-09-28 09-42-36
- I removed Markdown from singular default values and allowed values and updated the condition in `cli/generateMarkdown.ts`/`tsconfig/generateMarkdown.ts` for adding quotes s.t. singular values (`/^[.0-9a-z]+$/i`) are quoted and others aren't.
- This makes the treatment consistent between default values and allowed values (the same values are quoted/not) and between the tsconfig and CLI references (Markdown parsed where appropriate).

Finally some nitpicky edits:
- Consistently punctuated descriptions with ".".
- Removed "n/a" because it wasn't clear (to me) what's difference between these, and options without a default value?

### DIFF
--- a/packages/documentation/copy/en/project-config/Compiler Options.md
+++ b/packages/documentation/copy/en/project-config/Compiler Options.md
@@ -46,7 +46,8 @@ tsc app.ts util.ts --target esnext --outfile index.js
   <tbody>
 <tr class='odd' name='all'>
   <td><code>--all</code></td>
-  <td><code>boolean</code></td>
+  <td><p><code>boolean</code></p>
+</td>
 </tr>
 <tr class="option-description odd"><td colspan="3">
 <p>Show all compiler options.</p>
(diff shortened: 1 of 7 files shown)


orta:
I think this is probably good, will give it a run locally tomorrow 👍🏻

orta:
Thanks, given it a bit of a spit-shine and this is now good to go!

jablko:
Awesome, thanks for doing that!

PR #2050, opened 2021-09-29
Cut after comment

Mike-Dax:
Writing code snippets for complex React examples, we often find ourselves in the position where we're repeating "boilerplate" compositional components or wrapping raw JSX.

```tsx 
<Container>
  <ImportantComponent />
</Container>
```

or

```tsx 
const Page = () => (
  <Container>
    <ImportantComponent />
  </Container>
)
```

The `<Container>` ... `</Container>` or `const Page = () => (` ... `)` parts are unnecessary noise for our documentation in these places, but are syntactically required.

This PR adds a `---cut-after---` comment.

```tsx 
<Container>
// ---cut---
  <ImportantComponent />
// ---cut-after---
</Container>
```

```tsx 
const Page = () => (
// ---cut---
  <Container>
    <ImportantComponent />
  </Container>
// ---cut-after---
)
```

If there are other tests you'd like / a name change, let me know.

### DIFF
--- a/packages/ts-twoslasher/src/index.ts
+++ b/packages/ts-twoslasher/src/index.ts
@@ -817,6 +817,25 @@ export function twoslasher(code: string, extension: string, options: TwoSlashOpt
     tags = tags.filter(q => q.line > -1)
   }
 
+  const cutAfterString = "// ---cut-after---\n"
+
+  if (code.includes(cutAfterString)) {
+    
+    // Get the place it is, then find the end and the start of the next line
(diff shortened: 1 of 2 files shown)


orta:
I think this is reasonable answer (and not a case I'd really considered) - makes total sense. The implementation is 👍🏻 - give me a day or two to let the idea settle in my head but I'm likely to go with this or something just like this

orta:
Yep, I think this is 👍🏻

PR #2052, opened 2021-09-29
Feature/issue 2047

### LINKED ISSUE #2047: Docs: Seemingly-incorrect statement about ESM
**Page URL:**

https://www.typescriptlang.org/docs/handbook/2/modules.html#commonjs-and-es-modules-interop

**Issue:**

It says:

> There is a mis-match in features between CommonJS and ES Module because ES Modules only support having the default export as an object, and never as a function.

That seems to be saying ESM doesn't allow exporting a function as the default, but [it does](https://codesandbox.io/s/function-default-export-fv7md):

```js
// export.js
export default function() {
    console.log("Hi there");
}
// import.js
import fn from "./export.js";
fn();
```

I suspect it's trying to say something else, not that you can't use a function as the default export. But I'm not sure what it *is* trying to say, so I can only flag up the problem, I can't solve it (without help).

**Recommended Fix:**

We should correct or clarify the sentence, but I'm not sure how because it's not clear to me what it's trying to say.

tjcrowder:
Fix #2047 as discussed and update wording in the esModuleInterop option documentation about the spec's "recommendations."

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -306,7 +306,7 @@ squareTwo;
 
 ### CommonJS and ES Modules interop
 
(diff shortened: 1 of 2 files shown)


orta (approved):
Yeah, great, thanks!

PR #2053, opened 2021-09-29
Add some allowed values

jablko:
Add `es2021` `target`, `newLine` allowed values (`crlf` and `lf`), by pulling them from `ts.buildOpts[].type`.

Somewhat related to #2049: The allowed values in `option.type` aren't `` ` `` quoted, but #2049 adds the quotes.

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -248,34 +248,8 @@ export const defaultsForOptions = {
 };
 
 export const allowedValues = {
-  jsx: ["`react`", "`react-jsx`", "`react-jsxdev`", "`react-native`", "`preserve`"],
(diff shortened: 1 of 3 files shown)


packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -248,34 +248,8 @@ export const defaultsForOptions = {
 };
 
 export const allowedValues = {
-  jsx: ["`react`", "`react-jsx`", "`react-jsxdev`", "`react-native`", "`preserve`"],
```
jablko, line 251: These allowed values are currently in sync with `ts.buildOpts["jsx"].type.keys()`, but removing them ensures we'll pick up future updates?

orta: Yah, this seems reasonable to me

PR #2054, opened 2021-09-29
HTMLify tsconfig option summaries

jablko:
Noticed that we're displaying raw Markdown in the [tsconfig popups](https://github.com/microsoft/TypeScript-Website/pull/1911): 
Screenshot from 2021-09-29 11-23-53

This PR renders it to HTML. Likewise affects raw Markdown in the playground tsconfig panel? 
Screenshot from 2021-09-29 11-23-02


orta:
Good call 👍🏻

PR #2055, opened 2021-09-30
Fix the declared types version example to the text description

Andarist:
Given the description:
> On TypeScript 4.0 and above, an import for "package-name" would resolve to ./index.d.ts and for 3.9 and below "./index.v3.d.ts.

and the used filename:
> "index.v3.d.ts"

it seems that the proposed change is correct


orta:
Agree, thanks

PR #2057, opened 2021-10-03
Update The Handbook.md

fishmandev:
Typo fix


orta:
I'm not sure about the change to be honest, at least it would need both 'has' changing to 'have' in that sentence but both versions read fine to me

fishmandev:
> both versions read fine to me

Okay

PR #2061, opened 2021-10-05
Extract parseMarkdown()

jablko:
I'd like to extract some common code from `{cli,tsconfig}/generateMarkdown.ts`:
- the code to format lists of values
- the code to quote singular values
- the `parseMarkdown()` utility

This means that:
- `{cli,tsconfig}/generateMarkdown.ts` will share the same utility
- lists of default values and allowed values will be marked up the same
- marking up lists of default values will be delegated to the utility, simplifying them slightly

The only functional change is that allowed values (and related options) are now marked up in unordered lists, same as default values are already.

While I was at it I revised the code to quote singular values to fix the following: 
Screenshot from 2021-10-05 10-55-26


orta:
Looks reasonable to me 👍🏻 

Thanks!

PR #2064, opened 2021-10-06
Punctuate compound sentences

jablko:
Few commas to help delineate clauses like the following: 
Screenshot from 2021-10-06 10-47-26

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -168,7 +168,7 @@ export const relatedTo: [AnOption, AnOption[]][] = [
 
 function trueIf(name: string) {
   return [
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -168,15 +168,15 @@ export const relatedTo: [AnOption, AnOption[]][] = [
 
 function trueIf(name: string) {
   return [
-    `\`true\` if [\`${name}\`](#${name}) is \`true\``,
```
jablko, line 171: Equally clear but less busy/wordy, I think?

orta: I think this is a reasonable call 👍🏻

orta:
Yeah, this looks solid 💯

PR #2065, opened 2021-10-06
Uniform tsconfig status

jablko:
Currently we display "Recommended: True", "Status: Deprecated" and "Status: internal". What do you think of more uniformly displaying "Recommended", "Deprecated" and "Internal" instead?

||||
|-|-|-|
|Screenshot from 2021-10-06 12-07-17|Screenshot from 2021-10-06 12-08-29|Screenshot from 2021-10-06 12-09-59|
|Screenshot from 2021-10-06 12-10-59|Screenshot from 2021-10-06 12-11-54|Screenshot from 2021-10-06 12-12-41|


orta:
Yeah, I agree, this feels better :+1:

PR #2072, opened 2021-10-08
Make comments match demonstrated code

sgharms:


packages/documentation/copy/en/handbook-v2/Type Manipulation/Template Literal Types.md
```diff
@@ -93,9 +93,10 @@ const person = makeWatchedObject({
 
 person.on("firstNameChanged", () => {});
 
-// It's typo-resistent
+// Prevent human error (using a key instead of a signal)
```
orta: Maybe this? I don't think we use signal elsewhere in this doc

```suggestion
// Prevent easy human error (using the key instead of the event name)
```

orta:
Thanks!

PR #2074, opened 2021-10-08
Revise preserveConstEnums default value

jablko:
I think the `preserveConstEnums` default is `true` if `isolatedModules` is `true`? https://github.com/microsoft/TypeScript/blob/f6c0231f08353cf16d22a54ff33814110e6c9d7c/src/compiler/utilities.ts#L6213-L6215


orta:
TIL'd - yeah, it is 👍🏻 the linked function is used pretty much everywhere

PR #2078, opened 2021-10-10
Fix return type in doSomething(f: Function)

cnydw:
**Description from the handbook**: The global type `Function` describes properties like `bind`, `call`, `apply`, and others present on all function values in JavaScript.
It also has the special property that values of type `Function` can always be called; these calls return `any`:

**What has been done**: Add `return` to make sure that the `Function`'s return type `any` is propagated to `doSomething`. Otherwise the return type is `void`.


orta:
Seems reasonable to me

PR #2081, opened 2021-10-11
Update Literal reference on TS for JS Programmers

JuanCrg90:
The current link to literals sends to a deprecated page. This updates the reference to the right section.

Current deprecated reference: https://www.typescriptlang.org/docs/handbook/literal-types.html
New Reference: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types


orta:
Thanks!

PR #2090, opened 2021-10-15
Update Basics.md

jmcscript:
When first talking about the type `any`, use of the word **any** in the same sentence can be a bit jarring. It caused me a double-take on the first read. I bet we can maintain clarity by simply omitting the first reference.


orta:
agree

PR #2091, opened 2021-10-16
Correct grammar

Untitled-png:


orta:
Thanks for the PR but I don't think this is correct.

Untitled-png:
> Thanks for the PR but I don't think this is correct.

"Which" and "that" aren't interchangeable, even though they are often used that way in speech. There should also be a comma after JavaScript, because it's correct, but it also _technically_ creates ambiguity without it. I just thought it was worth pointing out since it's the lead sentence of the page.

sandersn:
Disclaimer: I am a linguist, so I know a lot of things about language but don't know what counts as "correct".

All 3 changes make the writing more correct in my dialect, but in British English I believe "which" and "that" are much closer to interchangeable.
The comma usage is down to personal style, in my opinion. If we want the web site to seem like "orta wrote this", keep it as-is. If we want it seem like "Microsoft wrote this", probably take this change.

orta:
Yeah, this is a pretty compelling argument - let's do it - thanks @inthmafr

PR #2041, opened 2021-09-27
typo in `classes.html`: toLowercse -> toLowercase

pr-o:
found a typo in `classes.html`

`toLowercse` -> `toLowercase`

PR #2095, opened 2021-10-19
fix type in documentation

jagnani73:
the method `toLowerCase` is typed as `toLowerCse`


HyunsangHan:
I guess it would be duplicated with https://github.com/microsoft/TypeScript-Website/pull/2041 which is closed :)

orta:
Yep, it's a dupe, thanks - this is intentional

PR #2099, opened 2021-10-20
fix receivedFile parameters in readme

YousefED:


orta:
Thanks! I also need to make a deploy with some browser fixes.

( Just triggered them, maybe 20m )

PR #2106, opened 2021-10-24
Update yarn (and some workflows)

tiziodcaio:
Updated the node version used in github actions to 14 (lts)


orta:
Thanks, if you remove the yarn changes I'm good with this PR - but the diffs in the yarn JS aren't something I can review (and thus could be a security hole)

tiziodcaio:
https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md
Have you seriously watched yarn when you've created the repository?

orta:
I don't think you get it. My point is, how did I know you have not included a keyloggers in this diff:
Screen Shot 2021-10-26 at 12 49 12 PM

tiziodcaio:
Oh sure... now I understand sorry for be a bit rude ;-)
if you want to do it better, I suggest to look a moment to [corepack](github.com/nodejs/corepack)... Now I will close the repo! Bye

PR #2107, opened 2021-10-24
Makes padLeft more readable

courtneypattison:
I completely understand if you folks disagree with me on this one, but I think `String.protoype.repeat()` is more readable than `Array.prototype.join()` for `padLeft`.


orta:
Yeah, this looks good to me - thanks

DanielRosenwasser:
I didn't even know `repeat` was a method 😬

DanielRosenwasser (approved):

orta:
Assuming that fixes the compiler errors, this should be good to go

PR #2110, opened 2021-10-25
Add allowed values for list options

jablko:
Extend the current behavior with respect to `option.type` to `option.element.type` as well. This e.g. gets the allowed values for list options, analogous to other options. The only current affected option is `lib`: This change keeps its allowed values up to date.


packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -3,6 +3,24 @@ import * as remark from "remark";
...
+    | "object"
+    | "list"
+    | Map<string, number | string>;
+  defaultValueDescription?: string | number | boolean | ts.DiagnosticMessage;
+  element: CommandLineOption;
+}
+
+declare module "typescript" {
+  const optionDeclarations: CommandLineOption[];
+  const optionsForWatch: CommandLineOption[];
```
jablko, line 21: Declaring these skeleton types means we can drop some runtime checks/assertions, below.

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -173,26 +191,17 @@ function trueIf(name: string) {
   ];
 }
 
-// Number/boolean options without explicit default are 0/false
-const defaultDefaults = { number: "0", boolean: "false" };
```
jablko, line 177: Merged -> https://github.com/microsoft/TypeScript/pull/46280/commits/4981c560171d94e1d7f88d78d2bf41f8ea0e612d

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -238,31 +247,47 @@ export const defaultsForOptions = {
...
-            synonyms.length > 1
-              ? synonyms.map((name) => `\`${name}\``).join("/")
-              : synonyms[0]
-          ),
-        ];
-      })
+    [...ts.optionDeclarations, ...ts.optionsForWatch].map((option) => [
+      option.name,
+      formatAllowedValues(
+        option.type === "list" ? option.element.type : option.type
```
jablko, line 270: Extend current `option.type` behavior -> `option.element.type`.

orta:
Structurally yep I like this, I think I need to do some design work to reduce the line height on it to get it merged 👍🏻 

Screen Shot 2021-10-26 at 12 47 18 PM

orta:
With ES2022 this list gets another 6 values, so maybe the sidebar doesn't scale here - will sit down and have a proper thing

jablko:
:+1: Makes sense, and no rush. Is this mergeable with the latest revision: Restore the `lib.md` and hide the `lib` allowed values? What's left is the refactoring, without actually affecting any current options.

orta:
It is 👍🏻

PR #2120, opened 2021-10-31
Move Node.js + ESM blog post content to reference page

DanielRosenwasser:
Currently I'm thinking of creating a reference page that uses the current blog post contents. These contents aren't comprehensive, but they're good enough for a casual reader and we can update it over time, and we can link to it from the blog post.

I figured a reference page would be the right place, but I'm open to ideas here.

### DIFF
--- a/packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
+++ b/packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
@@ -1,0 +1,240 @@
+---
+title: Experimental Support for ECMAScript Modules in Node.js
+layout: docs
+permalink: /docs/handbook/experimental-esm-node.html
+oneline: Using ECMAScript Modules in Node.js
+---
+
+> Note: this functionality is currently experimental and prone to changing.
+> It can currently only be used on [nightly builds](/docs/handbook/nightly-builds.html) of TypeScript,
+> and we are currently seeking feedback around bugs and pain-points.
+
+For the last few years, Node.js has been working to support running ECMAScript modules (ESM).
+This has been a very difficult feature to support, since the foundation of the Node.js ecosystem is built on a different module system called CommonJS (CJS).
+Interoperating between the two brings large challenges, with many new features to juggle;
+however, support for ESM in Node.js is now largely implemented in Node.js 12 and later, and the dust has begun to settle.
+
+That's why TypeScript brings two new `module` and `moduleResolution` settings: `node12` and `nodenext`.
+
+```json5
+{
+    "compilerOptions": {
+        "module": "nodenext",
+    }
+}
+```
+
+These new modes bring a few high-level features which we'll explore here.
+
+### `type` in `package.json` and New Extensions
+
+Node.js supports [a new setting in `package.json`](https://nodejs.org/api/packages.html#packages_package_json_and_file_extensions) called `type`.
+`"type"` can be set to either `"module"` or `"commonjs"`.
+
+```json5
+{
+    "name": "my-package",
+    "type": "module",
+
+    "//": "...",
+    "dependencies": {
+    }
+}
+```
+
+This setting controls whether `.js` files are interpreted as ES modules or CommonJS modules, and defaults to CommonJS when not set.
+When a file is considered an ES module, a few different rules come into play compared to CommonJS:
+
+* `import`/`export` statements (and top-level `await` in `nodenext`) can be used
+* relative import paths need full extensions (we have to write `import "./foo.js"` instead of `import "./foo"`)
+* imports might resolve differently from dependencies in `node_modules`
+* certain global-like values like `require()` and `process` cannot be used directly
+* CommonJS modules get imported under certain special rules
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
```diff
@@ -0,0 +1,240 @@
...
+    "exports": {
+        ".": {
+            // Entry-point for `import "my-package"` in ESM
+            "import": "./esm/index.js",
+
+            // Entry-point for `require("my-package") in CJS
+            "require": "./commonjs/index.cjs",
+
+            // Entry-point for TypeScript resolution
+            "types": "./types/index.d.ts"
```
andrewbranch: So I think we've since learned that this won't work; either `import` or `require` will be matched first and we will look for a colocated `.d.ts` file there. In this case `types` would need to be the first of these keys.

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
```diff
@@ -0,0 +1,240 @@
...
+There's a lot to this feature, [which you can read more about on the Node.js documentation](https://nodejs.org/api/packages.html).
+Here we'll try to focus on how TypeScript supports it.
+
+With TypeScript's original Node support, it would look for a `"main"` field, and then look for declaration files that corresponded to that entry.
+For example, if `"main"` pointed to `./lib/index.js`, TypeScript would look for a file called `./lib/index.d.ts`.
+A package author could override this by specifying a separate field called `"types"` (e.g. `"types": "./types/index.d.ts"`).
+
+The new support works similarly with [import conditions](https://nodejs.org/api/packages.html).
+By default, TypeScript overlays the same rules with import conditions - if you write an `import` from an ES module, it will look up the `import` field, and from a CommonJS module, it will look at the `require` field.
+If it finds them, it will look for a corresponding declaration file.
```
andrewbranch: I suggested

```suggestion
If it finds them, it will look for a colocated declaration file.
```

in my original review of the blog post and I stand by that suggestion. Any declaration file that describes the JS file in question is “corresponding,” but AFAIK the resolver only looks in one place—right beside the JS file—during this part of resolution.

andrewbranch (changes requested):

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
```diff
@@ -0,0 +1,240 @@
+---
+title: Experimental Support for ECMAScript Modules in Node.js
+layout: docs
+permalink: /docs/handbook/experimental-esm-node.html
```
orta: we _ideally_ don't want to be changing/breaking URLs in the future 

I can add something like `work-in-progress: true` or `experimental` support to rendering a page tomorrow ( like what we do for deprecated pages https://www.typescriptlang.org/docs/handbook/basic-types.html#string )

DanielRosenwasser: Yeah, I think that would work great here - thank you!

orta: Shipped in https://github.com/microsoft/TypeScript-Website/pull/2123 - if you switch this to its permalink  ( maybe `/docs/handbook/esm-node.html` ?) and add `experimental: true` to this yml and it will be 👍🏻

orta:
This post as-is will add the page to the site, but will not add it to the nav, edit `packages/documentation/scripts/generateDocsNavigationPerLanguage.js` to add it to the nav (or I can handle it after merging  👍🏻 )

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
DanielRosenwasser:
```suggestion
            // Entry-point for TypeScript resolution - must occur first!
            "types": "./types/index.d.ts"

            // Entry-point for `import "my-package"` in ESM
            "import": "./esm/index.js",

            // Entry-point for `require("my-package") in CJS
            "require": "./commonjs/index.cjs",
```

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
DanielRosenwasser:
```suggestion
permalink: /docs/handbook/esm-node.html
experimental: true
```

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
DanielRosenwasser: I don't really understand the distinction, but I'll add it for now and we can chat later.

packages/documentation/copy/en/reference/Experimental ESM Support for Node.md
privatenumber: Why should `types` occur first?

I tested in the nightly build and it seems to work in any order.

PR #2124, opened 2021-11-03
Add a reference to `sourcesContent`

joshkel:
Some online discussions about source maps reference `sourcesContent` (e.g., [here](https://scotch.io/tutorials/your-source-maps-are-broken-heres-how-to-fix-them) and [here](https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/#verify-artifacts-are-uploaded)). Referencing that keyword here may help TypeScript users.


orta:
Yeah, this makes sense - thanks

PR #2127, opened 2021-11-05
Document const enum pitfalls and workarounds

### LINKED ISSUE #37179: Pain points with const enum
At the express suggestion of a [package](https://www.npmjs.com/package/chalk) [maintainer](https://github.com/sindresorhus) whose package has over 36 million weekly downloads, I submit the following (original submission/comment is https://github.com/chalk/chalk/pull/373#issuecomment-592269429):

---
[I](https://github.com/michaelsbradleyjr/) wrote:

> [@babel/preset-typescript](https://www.npmjs.com/package/@babel/preset-typescript) gets about 3.7 million weekly downloads; [typescript](https://www.npmjs.com/package/typescript) gets about 7.9 million.
>
> This is not a rigorous analysis... but, because the problem arises with `const enum` being unsupported by `@babel/preset-typescript` (because it only works with code that conforms to `isolatedModules`), that means about half of the typescript projects out there can't use chalk 3.x.
>
> `@babel/preset-typescript` also doesn't fully support TS namespaces. The TS maintainers have publicly said that ES6 modules are the future; on the other hand, they've also publicly said they have zero plans to remove namespaces from the language. I don't know for certain, but I expect they would not be open to removing `const enum` either.
>
> There is a plugin ([babel-plugin-const-enum](https://www.npmjs.com/package/babel-plugin-const-enum)) that can help with the `const enum` situation. However, it's not a wonderful solution because those of us who use `tsc` to typecheck — which is everyone since `babel/preset-typescript` doesn't do any typechecking — need to disable the `isolatedModules` option. But it's important to have that option in place in projects using `@babel/preset-typescript`, especially in larger projects, to ensure that contributors don't commit code that won't work with it.
>
> In summary: many TypeScript projects won't be able to use chalk 3.x if it retains `const enum`. It's not really any one's fault, it's just how things are given the technical constraints of `@babel/preset-typescript` and the practical need to typecheck with `tsc` in a manner that detects code that isn't compatible.
>
> I'm sure none of the above will convince you to change your mind, and there's certainly no resentment on my part. I wasn't sure how many people who encounter this issue understand the core of the problem, so I hope my comment is helpful. Thanks for all your hard work on this wonderful library!

---

The [maintainer](https://github.com/sindresorhus) replied:

> They don't need to remove it, just officially discourage it ... `const enum` is a real pain-point for users. You can fix Chalk here, but there will be lots of other projects using `const enum` and are you going to spend time on this every time you encounter a project using `const enum`?

---

Is there something I'm missing as a consumer of this library running into problems with `isolatedModules` clashing with `const enum`?

Is there something the package maintainer could do to alleviate the problem in a manner that doesn't sacrifice the performance gains of `const enum`?

Is there something the TypeScript maintainers can do to help improve the situation?

cf. https://github.com/microsoft/TypeScript/issues/30590 and https://github.com/facebook/create-react-app/pull/4837#issuecomment-430107471.

Much thanks for your time and consideration!

jablko:
Attempt to recycle these learnings https://github.com/microsoft/TypeScript/issues/37774#issuecomment-961392924 into documentation, to address https://github.com/microsoft/TypeScript/issues/37179#issuecomment-596227119.

Fixes https://github.com/microsoft/TypeScript/issues/37179#issuecomment-596227119

### DIFF
--- a/packages/documentation/copy/en/reference/Enums.md
+++ b/packages/documentation/copy/en/reference/Enums.md
@@ -365,6 +365,33 @@ let directions = [
 ];
 ```
 
+#### Const enum pitfalls
+
+Inlining enum values is straightforward at first, but comes with subtle implications.
+These pitfalls pertain to _ambient_ const enums only (basically const enums in `.d.ts` files) and sharing them between projects, but if you are publishing or consuming `.d.ts` files, these pitfalls likely apply to you, because `tsc --declaration` transforms `.ts` files into `.d.ts` files.
+
+1. For the reasons laid out in the [`isolatedModules` documentation](/tsconfig#references-to-const-enum-members), that mode is fundamentally incompatible with ambient const enums.
+   This means if you publish ambient const enums, downstream consumers will not be able to use [`isolatedModules`](/tsconfig#isolatedModules) and those enum values at the same time.
+2. You can easily inline values from version A of a dependency at compile time, and import version B at runtime.
+   Version A and B's enums can have different values, if you are not very careful, resulting in [surprising bugs](https://github.com/microsoft/TypeScript/issues/5219#issue-110947903), like taking the wrong branches of `if` statements.
+   These bugs are especially pernicious because it is common to run automated tests at roughly the same time as projects are built, with the same dependency versions, which misses these bugs completely.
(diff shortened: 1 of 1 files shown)


orta (approved):
Yep, I agree this is good - I just will find some time to go over the words a bit and tweak but thanks this is a solid idea.

jablko:
:+1: Thanks!

orta:
I think we should take this pretty much as is, it feels solid to me - thanks!

PR #2129, opened 2021-11-06
Added info on keyword `unique` applied to Symbols.

IKoshelev:
Added information about TS keyword `unique` as it applies to `Symbol` declarations. Before adding, checked via search to make sure this keyword is not described anywhere else in documentation.

Information about `unique` keyword taken almost verbatim from release notes: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#unique-symbol

### DIFF
--- a/packages/documentation/copy/en/reference/Symbols.md
+++ b/packages/documentation/copy/en/reference/Symbols.md
@@ -56,11 +56,11 @@ let className = c[getClassNameSymbol](); // "C"
 
 To enable treating symbols as unique literals a special type `unique symbol` is available. `unique symbol` is a subtype of `symbol`, and are produced only from calling `Symbol()` or `Symbol.for()`, or from explicit type annotations. This type is only allowed on `const` declarations and `readonly static` properties, and in order to reference a specific unique symbol, you’ll have to use the `typeof` operator. Each reference to a unique symbol implies a completely unique identity that’s tied to a given declaration.
 
-```ts
-// Works
+```ts twoslash
+// @errors: 1332
 declare const sym1: unique symbol;
 
-// Error! 'sym2' isn't a constant.
+// sym2 can only be a constant reference.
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Symbols.md
```diff
@@ -52,6 +52,38 @@ let c = new C();
...
+let sym2: unique symbol = Symbol();
+
+// Works - refers to a unique symbol, but its identity is tied to 'sym1'.
+let sym3: typeof sym1 = sym1;
+
+// Also works.
+class C {
+  static readonly StaticSymbol: unique symbol = Symbol();
+}
+```
```
orta:
````suggestion
```ts twoslash
// @errors: 1332
declare const sym1: unique symbol;

// sym2 can only be a constant reference.
let sym2: unique symbol = Symbol();

// Works - refers to a unique symbol, but its identity is tied to 'sym1'.
let sym3: typeof sym1 = sym1;

// Also works.
class C {
  static readonly StaticSymbol: unique symbol = Symbol();
}
```
````

File: packages/documentation/copy/en/reference/Symbols.md [hunk 2]
```diff
@@ -52,6 +52,38 @@ let c = new C();
...
+
+```ts
+const sym2 = Symbol();
+const sym3 = Symbol();
+
+// Error: can't compare two unique symbols.
+if (sym2 === sym3) {
+  // ...
+}
+```
```
orta:
````suggestion
```ts twoslash
// @errors: 2367
const sym2 = Symbol();
const sym3 = Symbol();

if (sym2 === sym3) {
  // ...
}
```
````

orta:
Nice work, I just changed your codesamples to use the compiler but this is a great improvement thanks

PR #2133, opened 2021-11-10
Port some templates from strings -> JSX

jablko:
I'm interested in using [hastscript](https://github.com/syntax-tree/hastscript) + JSX for some of the tsconfig templates. This is a cosmetic change: It shouldn't change the website HTML at all.

We're already parsing Markdown -> ASTs and stringifying them -> HTML (remark-html wraps mdast-util-from-markdown and hast-util-to-html). Whereas currently we stringify the ASTs -> HTML and then stitch those strings together, I'd like to reverse those steps: Stitch the ASTs together and then stringify the amalgamation.

I think the advantages are:
1. Unlike strings, the ASTs have some type information, so this might catch some errors at compile/run time. #2014 flipped `</tr>`/`</td>` ...
2. Unlike strings, the AST nodes' meaning isn't reinterpreted depending on their context, e.g.
   - There's no risk of interpreting HTML nodes as code blocks: #2007
   - Leading/trailing newlines don't change the interpretation of Markdown inside HTML. The stringifier emits the appropriate whitespace for nested nodes' intrinsic types.


orta:
Sorry dude, I think this is going a bit too far from my general level of technical complexity in the website tooling - there's going to be a time when I'm not the key person maintaining this repo and I really don't want that next person to end up needing to figure out this section as a unique area with different rules. 

I do think it makes sense in general, but dumb string appending is dumb string appending - it's why I don't write the website as a series of react components because the rest of the team (generally) hasn't written anything but sample repros in React

jablko:
I'm grateful for your reviews! Is the first commit ([Use ESM](https://github.com/microsoft/TypeScript-Website/pull/2133/commits/601cc1f721158bbd2ee38e1135945946c5dced49)) mergeable on its own? (It's a chore that will need doing sometime?)

I like `.tsx` syntax partly because we're already using all of the pieces (almost):
- For building HTML with JavaScript, JSX is common.
- TypeScript is sufficient to compile away the JSX: React isn't involved.
- If `.tsx` syntax is acceptable, we can't get any simpler than compiling to hast, I think, given we're already using hast/mdast/remark.

And it moves opening and closing tags to the same expressions, which helps the next person to edit the markup, I think:

### Base

https://github.com/microsoft/TypeScript-Website/blob/7e4d8016f48c3e233aa9a728e8379550d402500d/packages/tsconfig-reference/scripts/cli/generateMarkdown.ts#L49-L122

### Head

`<tbody>` nice and balanced: https://github.com/microsoft/TypeScript-Website/blob/e02e663edc4245981035189f91abfa28b1e440ac/packages/tsconfig-reference/scripts/cli/generateMarkdown.tsx#L140-L155

But yeah, sticks out like a wart if we continue preferring the string appending pattern everywhere else, and my string appending gripes (subtly significant whitespace, no syntax checking) aren't the biggest of deals. :+1:

orta:
Sorry for the delay, yep the ESM change is cool - I'm a little wary of something like `    "generate:json": "yarn generate:json:tsconfig && yarn generate:json:cli && yarn generate:json:schema && yarn generate:json:msbuild",` causing 4 separate builds, but I think because you're using `--build` it will basically NOOP

jablko:
:+1: Yep that's what I had in mind, that `--build` will only run once.

I'll open a PR for just the ESM commit.

jakebailey:
I think 2 years on, a PR for this would need to be rewritten and reconsidered; the PR now has a lot of unsolved conflicts and we're even _more_ strapped for resources to keep the website running.

I'm going to close this PR, but if you think you're still going to do the work, we can reconsider it separately.

PR #2140, opened 2021-11-12
Remove the use of any from the recommended

ziacik:
The any is discouraged, yet it is used in "Do" at several places. Changed to unknown.


orta:
Yeah, I think this is 👍🏻

PR #2141, opened 2021-11-12
Use latest checkout and setup-node actions

jablko:
I'm seeing the CI `yarn install` step take +/- 5 minutes [currently](https://github.com/microsoft/TypeScript-Website/runs/4189840925?check_suite_focus=true#step:6:1), vs. +/- 1 minute with `actions/setup-node@v2` and a warm cache? I can't explain the difference, but seems like it might be worth updating?

Also, `actions/checkout@v2` does a shallow clone by default. The full clone [currently](https://github.com/microsoft/TypeScript-Website/runs/4189840925?check_suite_focus=true#step:2:1) only takes +/- 7 seconds, but might as well update that one too?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,21 +7,11 @@ jobs:
 
     steps:
       # Check out, and set up the node/ruby infra
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
         with:
           node-version: "13.x"
-
-      - uses: actions/cache@v2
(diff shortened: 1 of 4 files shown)


orta:
I think the deploy ones will need full history, because they search history for commits in the contribution graph - but the other two should be fine!

jablko:
Ah, thanks! For `attribution.json`? https://github.com/microsoft/TypeScript-Website/blob/7a6705dfc22d4918cbcafa939f53ecdcfd12baa0/packages/documentation/scripts/generateAttribution.js#L24

:heavy_check_mark: Done.

orta:
Thanks!

PR #2142, opened 2021-11-12
Correct a comment in the 'Objects vs Enums' sample

orteth01:
In the Objects vs Enums section there is a comment about needing an extra line to pull out keys from an object defined w/ `as const`. As far as I can tell, that line is actually for pulling out the values. This PR updates the comment.

### DIFF
--- a/packages/documentation/copy/en/reference/Enums.md
+++ b/packages/documentation/copy/en/reference/Enums.md
@@ -408,7 +408,7 @@ ODirection.Up;
 // Using the enum as a parameter
 function walk(dir: EDirection) {}
 
-// It requires an extra line to pull out the keys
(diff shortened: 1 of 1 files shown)


orteth01:
[ts playground example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAYgVzAUTAgtjAvDA3gKBiJgGsBTATwEYAuGKgGkOPIoCY62njTKBmOn24tKAFjqimAXxgBDCDFCQoAbnz4oFAA5kYAaUoQAKtt05WIAGYxNOq-CSoMajaZgA1WQBsEZY25xbMntEFDR0AG0LayCQx3CAXTUgA)

orta:
You're right - thanks!

PR #2143, opened 2021-11-13
Narrowing.md: fix comment type

btea:


orta:
Thanks for the PR but the original is correct, TypeScript knows that this can only be true and not ever be false

Screen Shot 2021-11-15 at 4 37 40 PM

btea:
Thanks for reply. But I think the description here should be more detailed,  use `const` the type is `true` and use `let` the type is `boolean`

image
image

btea:
Maybe I misunderstood you. Combined with the context, it should be used in conditional statements like `if`, and there is no need to save variables

orta:
This is due to https://www.typescriptlang.org/play/typescript/language/type-widening-and-narrowing.ts.html

btea:
@orta Thank you!

PR #2144, opened 2021-11-13
Prefer @returns over @return in examples.

nifr:
TSDoc uses `@returns` instead of `@return`. ([Reference](https://tsdoc.org/pages/tags/returns/))

Let's aim for a common standard.


orta:
https://jsdoc.app/tags-returns.html does too, so this looks good to me

PR #2145, opened 2021-11-14
Refine tsconfig schema

jablko:
- Remove nonexistent options.
  - `compilerOptions["fallbackPolling" | "watchDirectory" | "watchFile"]`: `error TS5023: Unknown compiler option 'fallbackPolling'.` (they exist in `watchOptions` instead).
- Remove command-line-only options.
- Support list options other than `lib`.
  - The set of list options with enumerated values is only `lib` currently, so no change.
- Update `pattern` as needed: If the schema rejects any of the enumerated values.


orta:
I've checked the schema diff and everything does look good 👍🏻 

Screen Shot 2021-11-15 at 4 49 21 PM

Clever move on the pattern change, it was definitely out of date but wow that's a regex now. I'll take it and see what that looks like when it gets upstreamed, thanks

PR #2150, opened 2021-11-15
Provide more context to template literal types

sgharms:
I found the flow in introducing Template Literals in types over-onerous. In many places, I wasn't sure what I was supposed to be watching. I've attempted smooth things out.


orta:
Yeah, I agree, I think this feels better - thanks

PR #2154, opened 2021-11-17
Adds a note about `this` in interfaces

orta:


pvinis:
Nice. we could also show something like
```ts twoslash
interface DomainOwner {
  webmaster: {
    name: string;
    email: string;
  };
  contact: this["webmaster"];
  setWebmaster: Partial<this["webmaster"]>
}
```

well 😅, I thought I would make `Partial<this>`, but then i didn't. I guess its easy enough with and without the Partial.

RyanCavanaugh:
Honestly, this is super cursed and not something we want to encourage 😅

PR #2158, opened 2021-11-18
feat: add new copy for internationalization of footer

yuriwithowsky:
adding new copy for internationalization of footer.


orta:
Cool, looks good to me - thanks

PR #2159, opened 2021-11-21
Update Modules.md (typo)

tavoyne:


tavoyne:
Sorry, [my bad](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#top_level_await).

PR #2161, opened 2021-11-24
Update Namespaces.md

priyanshumishra1009:


orta:
Thanks, but I'm not sure we need an extra word here - it doesn't change the sentence in a constructive way

PR #2164, opened 2021-11-25
Change "rather peculiar" to "distinctive"

ghost:
This is supposed to be a welcoming resource to new users of TypeScript, and making snobbish comments about JavaScript will only serve to give them a bad impression of your project before they even begin using it.


RyanCavanaugh:
I don't think this improves the clarity of the documentation. Anyone familiar with JS has encountered an `[object Object]` or `1 == true` from time to time; the fact that JS has quirks is not controversial. Doc authors can express a little opinion as a treat and doing so establishes a human connection between the author and the reader.

ghost:
> Anyone familiar with JS has encountered an [object Object] or 1 == true from time to time

Yes, and if you're familiar with the language it isn't "peculiar" at all.

> Doc authors can express a little opinion as a treat and doing so establishes a human connection between the author and the reader.

Thanks for the patronizing explanation. The only thing that happens here is that it makes the author sound like they have a bad attitude. As the reader in this case, someone looking to learn TypeScript, that plus your response simply tells me that there are other, better places to do it.

PR #2165, opened 2021-11-26
Fix deprecation link for index signature types

webstrand:
The new handbook page for Index Access Types does not discuss Index Signatures at all.


orta:
Thanks!

PR #2170, opened 2021-12-01
Update Creating DTS files From JS.md

ShuiRuTian:
set "declarationMap" to true by default


orta:
Yeah, I think this makes sense 👍🏻

PR #2173, opened 2021-12-01
ATA, always call finished callback after resolving dependencies

jwfwessels:
Currently, there is no way of knowing that ata finished if it doesn't resolve new dependencies. The `finished` callback is only called if there are newly resolved dependencies. 
This is confusing as the callbacks name implies that it would call after running Ata regardless as it's a signal of the runner completing.

PS: Thanks for the great package!

### DIFF
--- a/packages/ata/src/index.ts
+++ b/packages/ata/src/index.ts
@@ -19,7 +19,7 @@ export interface ATABootstrapConfig {
     /** A callback indicating that ATA actually has work to do */
     started?: () => void
     /** The callback when all ATA has finished */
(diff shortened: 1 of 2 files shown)


orta:
I'm pretty sure this is on purpose, because `start` only gets called if there is work to do - are you seeing a case where they don't match?

jwfwessels:
Ooooh 🤔  I did not grok that from the API. thanks for clarifying.

jakebailey:
Per the above, it seems like the current state is intentional?

PR #2174, opened 2021-12-01
Use ESM

jablko:
Use ESM in the tsconfig-reference workspace, to be able to use new ESM-only dependencies. Motivated by https://github.com/microsoft/TypeScript-Website/pull/2133#issuecomment-981768067, but it's a chore that will need doing sometime, independent of that PR. This PR just extracts the first commit from that one.


orta:
👍🏻

PR #2177, opened 2021-12-04
Slightly improve awaited docs

holgergp:
Hi all!

thanks for the awesome new release!

I stumbled over the docs of the new `Awaited`-Type. And to be honest, I do not quite understand those. 
This PR tries to "fix" the second-to-last paragraph, where I fail to understand its intention completely 🤷‍♂️ . But that may just be my lack of english mother tongue.

Cheers

Holger

### DIFF
--- a/packages/documentation/copy/en/release-notes/TypeScript 4.5.md
+++ b/packages/documentation/copy/en/release-notes/TypeScript 4.5.md
@@ -95,7 +95,7 @@ async function doSomething(): Promise<[number, number]> {
 }
 ```
 
-Now `Promise.all` leverages combines certain features with `Awaited` to give much better inference results, and the above example works.
+Now `Promise.all` leverages the combination of certain features with `Awaited` to give much better inference results, and the above example works.
 
 For more information, you [can read about this change on GitHub](https://github.com/microsoft/TypeScript/pull/45350).
 


orta:
Cool, looks reasonable to me 👍🏻

holgergp:
Nice! Thanks!

holgergp:
I think there is still some room for improvement in the docs here. At least I struggle to understand the concept here. 🤷‍♂️ (But that may be sth for a separate PR).
Things I've been stumbling over this far:
- `Awaited` is _not_ the Return type of an async function. That was my first impression. That Return Type still is `Promise<T>`. 
- `Awaited` can be used in (maybe rare) occasions, where I want to explicitly type the returned value of an awaited call.
- `Awaited` then helps me to avoid maybe nested/recursive Promises: e.g `Promise<Promise<string>>`
  - As a sidenote: I even struggled to find an example where I can construct such nested Promise construct. Maybe a hint here may be helpful to others to.
- The second code example does not even mention the `Awaited` type. I would have find it easier to understand if it would have been explicitly stated. Maybe another example would help here.
Maybe something like this (if I have understood it correctly....)

<code>

const pipe = (f, g, h) => (input) => h(g(f(input)));

const thirdAsyncFn = async (): Promise`<string>` => {
    return '3';
};

const secondAsyncFn = async (): Promise`<string>`  => {
    return '2';
};

const firstAsyncFn = async (): Promise`<string>`  => {
    return '1';
};
const pipedFn = pipe(firstAsyncFn, secondAsyncFn, thirdAsyncFn);

const returnedValue1: Promise<Promise<Promise`<string>`>> = pipedFn(2);

const returnedValue2: Awaited<Promise`<string>`> = pipedFn(2);

</code>

holgergp:
I've created a separate Issue for this: https://github.com/microsoft/TypeScript-Website/issues/2178

PR #2179, opened 2021-12-05
Eliminate the error when running the static example

gayratv:
When you press "try" button on Static Property Mixins - you see the error  "Return type of exported function has or is using private name"

To successfully run this example - you need to turn off "declaration" property in TS Config


RyanCavanaugh:
There seem to be two unrelated changes in this PR.

PR #2194, opened 2021-12-23
Typo in input and output

Grubba27:
While I was scrolling trought TS docs i found this realy diferent thing that is the table with a <div> and <div> output for react native. Its been a while since I've not used React Native but I remeber that we used <View> that could transpile to  UIView, <div>, android.view depending on your platform, but you should refer to this block of code as <View> like discussed in this doc (RN View Doc) [https://reactnative.dev/docs/view].
If i'm wrong on this, I would like to say sorry.  This is my first Pull request to making TypeScript better. If there is any things that i could do to make this pr better I gladly would change :) thx and good holidays


orta:
Thanks for giving it a shot, but the goal of this table is to show how the same code transforms to different outputs based on the jsx tag. There isn't a single element which _does_ have all that because of react-natives lack of native lowercase elements so `<div />` is an acceptable trade-off in that the key point is that 'jsx' is kept and it acts the same as preserve. I totally get the PR, but I think it might make it a bit more complex for folks to reasnbou about how the JSX option works

PR #2196, opened 2021-12-26
fix: helloWorld -> HelloWorld

38elements:
Since `helloWorld` is component, it is necessary that first character be capitalized.

https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized


orta:
Agree, thanks

PR #2208, opened 2022-01-05
Swap types on the description

sajadhsm:
According to the TypeScript warning:

> Argument of type 'string | number' is not assignable to parameter of type 'number'.

So I think the description should be "... adding a `number | string` to a `number` ..." instead of "... adding a `number` to a `number | string` ..."


orta:
Seems reasonable, thanks!

PR #2209, opened 2022-01-05
Add declare class fields in the class docs

orta:


packages/documentation/copy/en/handbook-v2/Classes.md
```diff
@@ -482,6 +482,36 @@ const b: Base = new Derived();
 b.greet();
 ```
 
+#### Declare Fields
```
robpalme: Maybe we could be more descriptive.

```suggestion
#### Type-only Field Declarations
```

packages/documentation/copy/en/handbook-v2/Classes.md
```diff
@@ -482,6 +482,36 @@ const b: Base = new Derived();
 b.greet();
 ```
 
+#### Declare Fields
+
+When [`useDefineForClassFields`](/tsconfig#useDefineForClassFields) is `true`, a class extension will overwrite getter/setters of the subclass if a field is declared in the class. This can be an issue when you need to re-declare a more accurate type in a subfield on a class. To handle these cases, you can write `declare` to indicate to TypeScript that there is no runtime change for this field.
```
robpalme:
```suggestion
When `target >= ES2022` or [`useDefineForClassFields`](/tsconfig#useDefineForClassFields) is `true`, class fields are initialized after the parent class constructor completes, overwriting any value set by the parent class. This can be a problem when you only want to re-declare a more accurate type for an inherited field. To handle these cases, you can write `declare` to indicate to TypeScript that there should be no runtime effect for this field declaration.
```

robpalme (commented):
Thanks for updating this.

orta:
Nice one, I knew it was worth waiting for your review- thanks

PR #2212, opened 2022-01-07
Improve scrollbar appearance in dark theme

Aaron00101010:
use [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) in dark theme to improve scrollbar appearance

before:

Screen Shot 2022-01-07 at 11 31 45

after:

Screen Shot 2022-01-07 at 11 32 11


orta:
Cool, never heard of that attribute 👍🏻

PR #2213, opened 2022-01-07
Fix code tags inside details tag

sajadhsm:
Markdown code tags are not being parsed inside the details tag.

before

Replacing ` `` ` with `<code />` fixes the issue, but I am not sure if there is a better solution or not.

after


orta:
Thanks!

PR #2214, opened 2022-01-08
Clarify property renaming example

peashutop:
The discussion here is about how property renaming during a destructure interacts with type definitions. But in this step of the example, the property renaming has dropped out. I think the point would be clearer if the property renaming from a few lines above remained, so the reader's attention is directed to what's meant to be new on this line (the addition of type declarations).


orta:
Agree, thanks!

PR #2215, opened 2022-01-08
Clarify behavior of node12/nodenext tsconfig options

zacharyliu:
I found the explanation of the `module: node12` and `module: nodenext` tsconfig.json options to be confusing for someone unfamiliar with the feature:
- implies that the feature is available in v4.5, but it's currently only in nightly builds
- says that the output is the same as `ES2020`, but it actually depends on whether it's a cjs or mjs file/package
- shows the CommonJS output despite mentioning `ES2020`

In this PR, I've tried to clarify the wording and provide more links to references.

It would be nice to also show the output for both `.cjs` and `.mjs` files. The easy way of doing this (setting `@filename: index.cjs`) doesn't work since `@typescript/twoslasher` doesn't search for these extensions. Fixing this would require changes in multiple repos since `gatsby-remark-shiki-twoslash` pulls in an older version from NPM.

https://github.com/microsoft/TypeScript-Website/blob/b8fb5ab114e75fa42c8b52ee54b2c7a03205c8c4/packages/ts-twoslasher/src/index.ts#L736

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/module.md
+++ b/packages/tsconfig-reference/copy/en/options/module.md
@@ -98,7 +98,9 @@ export const twoPi = valueOfPi * 2;
 
 If you are wondering about the difference between `ES2015` (aka `ES6`) and `ES2020`, `ES2020` adds support for dynamic `import`s, and `import.meta`.
 
-#### `node12`/`nodenext`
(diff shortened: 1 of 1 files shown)


orta:
Agree, with all your docs changes - these are great 👍🏻

PR #2217, opened 2022-01-09
Fix scroll bugs in tsconfig reference

zacharyliu:
Fix a handful of hash scrolling issues in the tsconfig.json reference page:
- Smooth scroll to the top of the target element, rather than to the center of it (caused by `block: "center"` option)
- Take into account the height of the sticky subnav by applying an offset
- Update the hash in the URL so that back navigation works
- Apply these scroll behaviors to all hash links on the page, not just the ones at the top
- Fix the highlighting of the current category in the `compilerOptions` subnav


orta:
Thanks, this looks great to me!

orta:
Gave it a quick check on staging and it's solid - good work

PR #2218, opened 2022-01-09
Fix a Markdown syntax typo

bangseongbeom:


orta:
TIL'd of that syntax - thanks

PR #2222, opened 2022-01-12
Make Working with Union Types clear and concise

sarscode:
This statement can be written in singular to be more  clear and concise.

**Good**
> TypeScript will only allow you to do `things` with the union if `that thing` is valid for _every_ member of the union.

**Better**
> TypeScript will only allow an operation if it is valid for _every_ member of the union.


orta:
Agree, thanks

PR #2225, opened 2022-01-14
Remove `node_modules` from tsconfig `exclude`

lorensr:
Since it's excluded by default, people don't need to put it in tsconfig, so I think the current example is misleading.
 https://github.com/microsoft/TypeScript/blob/70097c490807312faceba75c9e91b9547d3e16b9/src/compiler/utilities.ts#L6418-L6420

image


jakebailey (approved):
LGTM though I feel like we should really describe that we _always_ exclude `node_modules` even if it's not listed in `exclude` in the `tsconfig` reference for the option. But, that's a different page.

PR #2230, opened 2022-01-17
Upgrade gatsby

orta:
Just to keep it all up to date


jakebailey:
Closing in favor of #2717, since that one is more recent and bumps to gatsby 5 instead of 4.

PR #2231, opened 2022-01-18
Update Classes.md

NeonMika:
The docs mentioned that Java has a concept of "static classes", which is not the case. C#, on the other hand, has that feature.


orta:
Thanks!

PR #2234, opened 2022-01-21
Fix broken(?) link on Type Compatibility page

MichaelAllenWarner:
Add `/docs` prefix to link href because without it I'm getting `"The resource you are looking for has been removed, had its name changed, or is temporarily unavailable."`. (I'm assuming this isn't a "temporarily unavailable" situation, but if it is then disregard.)


orta:
Yeah, this change is the right one - thanks!

PR #2236, opened 2022-01-21
Fix grammatical error.

HarryAllen1:
It would be better to make this 2 sentences.


orta:

PR #2239, opened 2022-01-23
Adjust line property of twoslash errors to account for multi-file

gabro:
Fix for https://github.com/shikijs/twoslash/issues/135


packages/ts-twoslasher/test/results/multiFileErrors.json
```diff
@@ -0,0 +1,85 @@
...
+      "targetString": "sum"
+    }
+  ],
+  "errors": [
+    {
+      "category": 1,
+      "code": 2345,
+      "length": 7,
+      "start": 150,
+      "line": 7,
```
gabro: this is the main change: without this PR it would (incorrectly) be `"line": 1`

packages/ts-twoslasher/src/index.ts
```diff
@@ -709,6 +709,7 @@ export function twoslasher(code: string, extension: string, options: TwoSlashOpt
   // We can't pass the ts.DiagnosticResult out directly (it can't be JSON.stringified)
   for (const err of relevantErrors) {
     const codeWhereErrorLives = env.sys.readFile(err.file!.fileName)!
+    const lineOffset = code.split("\n").indexOf(codeWhereErrorLives.split("\n")[0])
```
gabro: this is a really naive way of doing it, I'll try to push a refactor later today

orta: 👍🏻 structurally looks right though

gabro: It works in most cases, but if two filenames happen to start with the same line, it will always pick the first one.

I was thinking of either:
- matching the `// filename: ` name which corresponds to `err.file!.fileName)`
- adding the line offset info directly in `nameContent` (since that's where the code is being traverse and split) so that it can be referenced here.

Any preference or alternative approaches?

gabro:
Also @orta would you mind approving the workflow run? I'm a first-time contributor to this repo so it's currently on hold

packages/ts-twoslasher/src/index.ts
gabro: Ok, I've pushed a refactor which implements the first proposal (so that the diff is more contained).

I've also added a test case which stresses the edge case I mentioned above (two virtual files beginning with the same line). The older implementation would yield an incorrect result, while the revised one works

orta: Yeah, this looks really solid to me, nice call 👍🏻

orta:
The fails do look legit, might be that there's another test which has a multi file error which needs a snapshot update 👍🏻

orta:
( As you're going into the codebase anyway, can you do a server patch to the package json for twoslash, that'll get it deployed on merge )

gabro:
> The fails do look legit, might be that there's another test which has a multi file error which needs a snapshot update 👍🏻

yep, it was in the `handlesJSON` fixture and the updated snapshot looks legit.



> ( As you're going into the codebase anyway, can you do a server patch to the package json for twoslash, that'll get it deployed on merge )

you mean bump the version? Sure, see the last commit.

gabro:
Hi @orta can you re-run the CI? Tests should be alright now

orta (approved):
I can't, heh, my GitHub access is now read only also - but this all looks good to go

gabro:
@orta oh, I just saw your tweet from Friday 😅 Congrats for all the work you've done for the TypeScript ecosystem!

Any other maintainer who can test, merge and release this?

tonyhb:
Adding a filename to twoslash also causes this error, which the PR doesn't fix.  I'm assuming it's the same code.

EG:

```
// @filename: ./inngest/new-pr.ts
import { createFunction } from "inngest";
 
export const newPR = createFunction("New PR", "github/pull_request", ({ event }) => {
 // This function is triggered when the `github/pull_request` event is received via
 // a GH webhook.
});
```

The underlines are offset incorrectly due to filename.

jakebailey:
Is this PR still up to date and accurate? There is a conflict, though just in the version number for the package.

gabro:
Hi @jakebailey, the original fix still stands, I think. I haven't addressed @tonyhb's concern.
I'm not sure if I'll have time to put more work into this, but if you're interesting in emerging it I can at least resolve the conflict

jakebailey:
I'm sure I can review it. It's just a package.json conflict, so I will quick resolve the conflicts.

jakebailey (approved):

Issue #1875, opened 2021-06-06
Docs: v2 handbook doesn't document Excess Property Checks


jcalz:
**Page URL:** https://www.typescriptlang.org/docs/handbook/2/objects.html

**Issue:** The [documentation about excess property checks](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks) in the deprecated v1 of the handbook doesn't seem to have any analog in v2.  If you click the "go to the new page" link in the deprecation dialog, it brings you to the "Object Types" page which doesn't mention excess property checks at all.

**Recommended Fix:** Add a section on excess property checks to v2 of the handbook, probably on the "Object Types" page.  I was first thinking maybe I could just copy the section from the old handbook, but it refers to different examples, and I'm not even sure where on the new page it should go.  Seems like a bigger rewrite than I'd be comfortable just dropping a PR on, although if someone can outline how it should be done, I'd be happy to do it.

jzxchiang1:
Yes, this is actually quite important.

Excess property checks are a somewhat unexpected feature of TS, given that TS also supports structural subtyping. It's not obvious at all to a beginner.

See https://stackoverflow.com/questions/70360922/typescript-structural-typing-with-exception. There is confusion out there...

PR #2242, opened 2022-01-24
Migrate and update the excess property checking docs

### LINKED ISSUE #2240: Idea: improve documentation for strict object-literal assignment checks (including spread-syntax exception)
**Improvement Idea:**

It's possible I missed it, but in neither the Handbook nor Reference section of the site did I find a clear explanation  of the [object-literal "exception"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-6.html#stricter-object-literal-assignment-checks) to TypeScript's structural typing system (i.e., that excess properties are _not_ allowed in object literals). There's an example [here](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types) where the "Object literal may only specify known properties" message appears, but there's no real explanation in the text of the excess-property issue like there was [in the old docs](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks). I think this is an important enough feature of the language to be covered explicitly, perhaps on the [Object Types page](https://www.typescriptlang.org/docs/handbook/2/objects.html).

Relatedly, I didn't find any mention _at all_ on the website of the [spread-syntax exception to that object-literal exception](https://github.com/microsoft/TypeScript/pull/26798) (i.e., that excess properties _are_ allowed in an object literal if they're spread in). Maybe this subtlety could be included as part of a new section on the special treatment that object literals get.

**Why:**

These details are important but currently difficult to discover.

**Alternative ideas:**

None that I can think of.

orta:
Fixes #2240 and #1875 

Migrates and does an editing run over the handbook v1 version of the excess property check docs

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Interfaces.md
+++ b/packages/documentation/copy/en/handbook-v1/Interfaces.md
@@ -5,6 +5,8 @@ permalink: /docs/handbook/interfaces.html
 oneline: How to write an interface with TypeScript
 handbook: "true"
 deprecated_by: /docs/handbook/2/objects.html
+deprecation_redirects:
+  [excess-property-checks, /docs/handbook/2/objects.html#excess-property-checks]
 ---
 
 One of TypeScript's core principles is that type checking focuses on the _shape_ that values have.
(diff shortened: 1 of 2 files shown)


MichaelAllenWarner:
The lack of documentation for excess-property checks in the v2 handbook seems like a glaring omission. Would be great to see this merged!

As mentioned in #2240, I also think it would be nice to document the fact that excess properties _are_ allowed in object literals if they're spread in with the `...` operator.

jakebailey (approved):

PR #2246, opened 2022-01-26
Add the 'index.node' resolution variety

Flamenco:
Add the 'index.node' variation for completeness.

If Typscript does NOT respect this, it should also be indicated.


RyanCavanaugh:
TS does not support this and I've never heard anyone ask about it either.

Flamenco:
Then maybe you should school yourself @RyanCavanaugh , or allow a response.  It is CLEARY part of the node.js resolution.

https://www.typescriptlang.org/docs/handbook/module-resolution.html

RyanCavanaugh:
It definitely is.

While the content is correct, the Handbook in particular requires a certain amount of brevity, and needs to maintain a tight focus on the most foundational aspects of TypeScript. This content might be better suited for an alternate documentation format with fewer scope constraints.

Flamenco:
Well, given the detail that is stated in the current TS walkthough, adding the complete algorithm as implemented by node does not add much more.  

So I will leave it as that.  IMO it was an oversight, and should have been stated.  Hence my PR...

PR #2249, opened 2022-01-31
Relate importsNotUsedAsValues -> preserveValueImports

jablko:
`preserveValueImports` is already related to `importsNotUsedAsValues` in the tsconfig reference. Add the reciprocal link, like #1099.


jablko:
I think the CI failure is fixed by https://github.com/microsoft/TypeScript-Website-Localizations/pull/135?

orta:
Agree, looks good to me

jakebailey (approved):
LGTM though we should really also add `verbatimModuleSyntax`. Clearly, this PR is old so I'm not surprised it's not here.

PR #2253, opened 2022-02-05
Update modules.md

avedetvedea:
Files are referred to as TS in comments, but then immediately after imported as JS.


orta:
Thanks, but this is intentional

avedetvedea:
> Thanks, but this is intentional

Should I explicitly ask why?

PR #2260, opened 2022-02-07
Polish const enum pitfalls and workarounds

jablko:
- be plainer about the inlining/single-file transpilation dichotomy
- include `preserveValueImports` alongside `importsNotUsedAsValues`
- update TypeScript ESLint link
- better (hopefully) prose?

### DIFF
--- a/packages/documentation/copy/en/reference/Enums.md
+++ b/packages/documentation/copy/en/reference/Enums.md
@@ -367,30 +367,32 @@ let directions = [
 
 #### Const enum pitfalls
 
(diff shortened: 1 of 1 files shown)


jakebailey:
A year on, I don't know that the content here is fully accurate; `preserveValueImports` and `importsNotUsedAsValues` are now deprecated in favor of `verbatimModuleSyntax`, at least, and the link has been fixed (hence the conflict).

I think that this will probably need a rewrite, another time. To keep the PR list clean, I'm going to close this one but I'm happy to review a modified new one.

PR #2266, opened 2022-02-14
Add Nx build tool to the bootstrap page

jaysoo:


orta:
These aren't build tools but tools for bootstrapping a TS + node project, I'd have to re-think a whole new section for that and I'm not sure its appropriate as-is

jaysoo:
@orta Sure no problem. Maybe a monorepo tooling section? There are a few things like nx, turborepo, maybe even lerna (although that is more language agnostic).

Yonet (approved):

DanielRosenwasser:
TSDX used to be on this page right? It doesn't seem too unreasonable to list this here.

orta:
I dropped tsdx last year for being unmaintained - this is for nx a monorepo build tool, which this PR adds to the node CLI tool bootstrapping section (which would be incorrect). 

To add this we'd need a new section for 'monorepo tools' which would include rush and tubrorepo. There are others but they don't describe themselves as supporting typescript out of the box which is the goal of this page

Yonet:
@orta nx is also a CLI for monorepo and component creation. If there is a monorepo section, it would make sense to add it there, though it wouldn't be wrong to have it under CLIs as well.

jaysoo:
I updated the PR to point to our TS-specific page https://nx.dev/getting-started/nx-and-typescript

Can someone merge this PR, or suggest where I should list Nx? Happy to close this PR as well if someone has another PR Nx that is accepted.

jakebailey:
I'm trying to review this, but, where even is this content on the website? I cannot find it for the life of me.

jakebailey:
Okay, it must be https://www.typescriptlang.org/docs/bootstrap.

I'm not sure that this should be in the section that it's been inserted into, honestly, as Orta said. But I'm not sure what other groups it should be in either. Probably a `monorepo` section, yeah, if we're still keeping the page looking like this.

But skimming the list and seeing what's already there... I was trying to figure out when this was last edited, and it was last touched in 2020. Is this still a resource people use? It took me a long time to find it, and it hasn't been updated in so long that it probably isn't super accurate anymore. Many of the tools listed are a little on the dusty side.

orta:
It was kept around because I didn't want to break old URLs from before this version of the website, and so it got some modern content - dropping it and having it just forward to the root isn't too unreasonable.

Issue #2247, opened 2022-01-27
[Playground] Twoslash highlights don't work with TS 4.6


Andarist:
**Repro:**

[repro playground](https://www.typescriptlang.org/play?ts=4.6.0-dev.20220116#code/C4TwDgpgBAsiCiA3CA7YBnKBeKBvKokAXFAOQBiA8paVAL5QA+eB4EJpAQgIIBKtdALAAoEQHoxUAIwjC0AJIoAJhAAenEABU2AHk1Q1wVEsz45JdMABOASxQBzegD5seEVCgBtANJQ7UTU9SOVIAXVCSeFVrAEMAY2A9ABoWcyhfOicAbhE6HNFhOShFFVUIJVcStQ1tSB04JFQMJ3FJKAA9AH4RVqgAJh7CtigAVRQbAHsUTQnFIyt0CATJlD0XHHcoAAp9Q2NMGJQQKE7t1RJNAEpsF0PjkhQIZCtrveVMTa3zvxQAMwgrFBeNcsLcjicgVAHk8AYMilV1Fo2H09AZovtUmwLNY7I5Mq4xisZnMAYtllNUW8TFA7hDeEsJlYlHpPAAiOSs0IpTQuaHPFoFCRQACu9gANsc5LJhgjyn1KspqkjICiGsg0OgBUKOt1BZIAMxw4YAYQmAFswMKjDpuGijO8aUd1iwfD8oABrCAgCa-KDcCJ+nyhehQABkIpQ7pQEwA7ih8tLIMVFYjahB9ZT0Q6zFioJZbA5nK5TRarRAdAiasi1gLesAABY2TBTaBWCAxCpm4VxetQABGEGA80TChT5X1CtKVcgGbVTU1vR1gyFABYjUnK8qICvM-bqTniHmcYX8ThcJtPPA3foYph4GyOeFIrkE0MN2OlCvJ0q0zu5xqtTaLogA)

**Expected behavior:**

[...]

PR #2267, opened 2022-02-15
Fix copy URL to clipboard keyboard shortcut

andrewbranch:
I guess monaco changed their enum format 😕

### DIFF
--- a/packages/create-typescript-playground-plugin/template/package.json
+++ b/packages/create-typescript-playground-plugin/template/package.json
@@ -26,7 +26,7 @@
     "@rollup/plugin-typescript": "^3.0.0",
     "@types/react": "^16.9.23",
     "concurrently": "^5.1.0",
-    "monaco-editor": "^0.29.1",
+    "monaco-editor": "^0.32.1",
     "node-fetch": "^2.6.0",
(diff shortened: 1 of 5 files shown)


RyanCavanaugh (approved):

orta:
Looks like legit CI fails from the Monaco update FWIW

That said, this change:

> Error: ➤ YN0000: [@typescript/playground]: src/twoslashInlays.ts(5,5): error TS2322: Type '(model: ITextModel, _: Range, cancel: CancellationToken) => Promise<InlayHint[]>' is not assignable to type '(model: ITextModel, range: Range, token: CancellationToken) => ProviderResult<InlayHintList>'.

Is a tricky problem to de-tangle because of https://github.com/microsoft/TypeScript-Website/issues/2247 - if you change the API usage then doing `// ^?` will crash (the BG thread) on the current and older TypeScript builds. 

I think this change should probably happen post 4.6 and we only support the twoslash queries on 4.6 with the new inlayformat shape which that compiler error notes.

andrewbranch:
Yeah, I realized that when I went to start fixing those CI errors. I might just hack the keyboard fix without the typings update for now.

PR #2269, opened 2022-02-15
Show links to playground on focus

quiteolly:
Because "Try" links in code snippets are focusable, they should be displayed for keyboard users navigating to them. Currently that is not the case. This pull requests modifies existing `:hover` styles to also work on `:focus` (`.code-container` can be focusable on small resolutions where there is scroll on a container).

As an aside, the colour contrast level of the links does not pass WCAG 2.1 Level AA requirements (2.84, 4.5 minimum is required), but I am not sure which colour should be used instead.


orta:
Thanks for improving this! 

I think it's reasonable to use `#235a97` (seen in the the search background color)  which is available as `--link-color` in the CSS vars to hit a contrast of 7 when active

quiteolly:
Replaced `#719af4` with `var(--link-color)` where appropriate, thank you for the suggestion. The colour contrast is now better in both themes.

orta (approved):

PR #2271, opened 2022-02-15
Hide "Was this page helpful?" popup from focus when not visible

quiteolly:
Elements that get hidden with `opacity: 0` still get exposed to keyboard users and are potentially focusable, even while they are invisible and are ignored by some (not all) screen readers. While `display: none` on the elements would remove the transitions, `visibility: hidden` should not have this effect in modern browsers and should correctly remove its children from focus order. The transitions should still be tested by someone else for good measure.

Other components on the website probably require similar changes to be made accessible.

### DIFF
--- a/packages/typescriptlang-org/src/templates/documentation.tsx
+++ b/packages/typescriptlang-org/src/templates/documentation.tsx
@@ -96,7 +96,7 @@ const HandbookTemplate: React.FC<Props> = (props) => {
       <section id="doc-layout" >
         <SidebarToggleButton />
 
-        <div className="page-popup" id="page-helpful-popup" style={{ opacity: 0 }}>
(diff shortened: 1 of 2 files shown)


ghost:
[CLA assistant check](https://cla.opensource.microsoft.com/microsoft/TypeScript-Website?pullRequest=2271) <br/>All CLA requirements met.

orta (approved):
Cool, thanks!

jakebailey (approved):

jakebailey:
Hm, #2338 was accepted first. Probably this could replace it if it still works?

jakebailey:
Going to just close this; the other fix came from the accessibility team so I think the current state is "good enough".

PR #2274, opened 2022-02-16
Skip sidebar navigation with skip link on Docs pages

### LINKED ISSUE #2270: "Skip to content" link does not skip all navigation on some pages
**Improvement Idea:**
On pages with sidebar navigation, "Skip to content" link should skip to actual content, and not to the end of top menu.

Currently, the skip link on typescriptlang.org is done with a `<div id="site-content"></div>` block inside the header component that references `href="#site-content"`. This does not skip all navigation on the page in Docs and Handbook sections: users wishing to skip to main content still have to skip through all `#sidebar` links, even on mobile where sidebar navigation is not displayed by default. This should be done differently in a way that would consistently skip to actual page content instead of another navigation block.

I can provide a PR if we agree on some solution here.

**Why:**
Keyboard users rely on skip links to get to the main content on the page. Currently, "Skip to content" link does not skip to the main content on Docs and Handbook pages, and users have to skip additional 10-15 links before getting to the actual page content.

**Alternative ideas:**

quiteolly:
Fixes #2270 with the solution provided by @orta. See the issue for the longer explanation.

I was a bit weary about the possible duplication here (`#handbook-content` being present in two places), but decided to go with their suggestion given that they are a former maintainer.

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/TopNav.tsx
+++ b/packages/typescriptlang-org/src/components/layout/TopNav.tsx
@@ -8,6 +8,7 @@ import { createIntlLink } from "../IntlLink";
 
 export type Props = {
   lang: string
+  skipToAnchor?: string
 }
 
 import { navCopy } from "../../copy/en/nav"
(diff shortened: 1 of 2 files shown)


orta (approved):
Thanks, yeah, I'm OK with the duplication 👍🏻

jakebailey:
I'm not against taking this, but, as far as I can tell, on pages with `#handbook-content`, `#site-content` is empty and so they visually do the same thing. Are you seeing otherwise?

jakebailey (approved):
Ah, I understand now:

> Keyboard users rely on skip links to get to the main content on the page. Currently, "Skip to content" link does not skip to the main content on Docs and Handbook pages, and users have to skip additional 10-15 links before getting to the actual page content.

This is explicitly about keyboard use, of course. Visual isn't the problem!

PR #2275, opened 2022-02-16
code typo and grammar nits

brettz9:
The change with "browsers" has to occur or the "don't" needs to be changed to "doesn't" to match. (This requirement is despite the fact that "runtimes" is plural.)

The preceding sentence is very unclear but I didn't know how to change it:

> In a CommonJS module, this just boils down to a require() call, and in an ES module, this imports [createRequire](https://nodejs.org/api/module.html#module_module_createrequire_filename) to achieve the same thing.

I think perhaps it is supposed to be the following, but I didn't change it as I wasn't sure:

> In a CommonJS module, this will just be converted into a simple require() call, whereas in an ES module, a [createRequire](https://nodejs.org/api/module.html#module_module_createrequire_filename) will be imported to achieve the same thing.


orta (approved):
Cool, yeah agree - thanks!

RyanCavanaugh:
These look right but there are merge conflicts I can't resolve online. Can you take a look?

brettz9:
Sure, force pushed an update

PR #2277, opened 2022-02-19
Add `typings-for-npm-packages.html` into redirect paths

soartec-lab:
The next path linked at `https://github.com/microsoft/TypeScript/wiki/Typings-for-npm-packages` is 404.

https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html

this path was not included in the list of redirects.

https://github.com/microsoft/TypeScript-Website/blob/8a19ee4dc8d26ac7d5536664788fe88064bf7825/packages/typescriptlang-org/src/redirects/setupRedirects.ts#L18

It's referenced in many other projects, not just the wiki above, so I'd like to include it in the list of redirects. I added this.


orta (approved):
Makes sense to me 👍🏻

jakebailey:
I opted to fix the wiki link instead. If there's any other people still linking this, we can send a PR for it, but I don't think it's required anymore. Thanks for sending this, though, and sorry for the long delay!

soartec-lab:
Sounds good. Thank you for me contact!

soartec-lab:
@jakebailey 
Sounds good. Thank you for me contact!

PR #2279, opened 2022-02-19
Fix generateTypesForFilesInDocs.js on Windows

SamB:
Without this, `yarn bootstrap` (which the top-level README.md tells you to run) will garble AllFilenames.d.ts, and generateDocsNavigationPerLanguage.js will no longer type-check.

### DIFF
--- a/packages/documentation/scripts/generateTypesForFilesInDocs.js
+++ b/packages/documentation/scripts/generateTypesForFilesInDocs.js
@@ -6,20 +6,20 @@
 // yarn workspace documentation create-handbook-nav
 
 const fs = require("fs");
-const { join, basename } = require("path");
+const path = require("path")
 const { format } = require("prettier");
 
 // prettier-ignore
 const getFilePaths = folderPath => {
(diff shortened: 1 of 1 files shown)


SamB:
Rambling side-information:

Incidentally, I apparently had a different fix for this just sitting in my unstaged changes for probably the past 10 months, but it seems as though I had entirely rewritten the directory traversal, and when I looked at it again yesterday I hadn't the slightest idea why I had done that, so I did it over with only a little refactoring around the imports from path.

My previous attempt also used [slash](https://github.com/sindresorhus/slash), but this time around I got a bit confused about which package.json is supposed to declare that and just did it myself. I should have just copied most of this line from `slash`:

	return path.replace(/\\/g, '/');

but instead I ended up learning that String.prototype.replace doesn't work like a Python user would expect.

orta:
Makes sense, thanks - we test that it builds in windows but don't test that it type-checks!

orta (approved):

jakebailey (approved):
Verified that this works on my Windows machine. Thanks!

jakebailey:
There are other Windows specific issues I'm seeing locally, but not related to this I don't think.

PR #2282, opened 2022-02-25
Improved comments in the CLI usage examples

mattalxndr:
I was having a hard time understanding the comments in these examples.


orta:
Thanks, but these previous examples took quite a lot of team discussion to get as they are based on the TS CLI output. I think if you want to change these you'd need to start by changing the CLI output in the TypeScript repo and then bring those changes back here. 👍🏻

mattalxndr:
Makes sense to stay with the same verbiage as the output. I still humbly suggest that you have someone at least proofread for grammar.

PR #2284, opened 2022-02-27
Fix the way to refer to imported items

mizdra:
The way to refer to the imported item is wrong.

Since we are importing `helper` named from the foo module, I think it is correct to write `helper();`.


SMAKSS (approved):

mizdra:
@SMAKSS Thank you for approve! Could you merge this PR?

jakebailey (approved):

jakebailey:
This seems to no longer be a problem in the final file: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/documentation/copy/en/reference/ESM%20Support%20for%20Node.md

PR #2286, opened 2022-02-28
Adds code owners self merge to the repo

orta:
Adds the same infra as used on the lib DOM repo, and localization repos for the last 2 years. First, what _cannot_ be merged by folks in the CODEOWNERS:

 - Things which affect node modules installs
 - Infra build scripts
 - GitHub yml stuff
 - LICENSE
 - PRs which are not green

If you'd like to trust but verify, [this is the locked code for the self-merge](https://github.com/OSS-Docs-Tools/code-owner-self-merge/blob/1.6.5/index.js) ~300 lines of JavaScript.

Nearly all "_make this edit to the docs_"  PRs only affect a markdown doc, so this can be used to handle the small PRs but not to ship an update of the TypeScript release for example (because that affects the yarn lockfile).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+packages/**/* @orta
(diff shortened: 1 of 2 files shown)


jakebailey:
Thanks for sending this. I did a quick check with the team and unfortunately I don't think we're going to enable this, given the down-scoping of the repo / prod deploy permissions and so on.

PR #2293, opened 2022-03-02
if X !== E.Foo it can only be E.Bar

Zxela:
Docs currently state that if x !== E.Foo then x can only be E.Foo. 

I believe this should read X can only be E.Bar


orta (approved):
Thanks!

Zxela:
I misread the document - The text as it is currently written (before my changes) is actually correct.

PR #2304, opened 2022-03-15
Fix broken link in docs

mattalxndr:
1. Was on https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html
2. Clicked "Templates" and was brought here: https://www.typescriptlang.org/docs/handbook/declaration-files/templates.html
3. Clicked "global-modifying-module.d.ts" and got 404.

The link should go here I am assuming:

https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html


jakebailey:
Dupe of #2250.

PR #2309, opened 2022-03-24
Change the `any` type in examples into `unknown`

mprinc:


jakebailey:
In theory, this is a good change, but it only changes a few of the assertions and there are more below that are staying as `any`. But in general, I feel like these release notes are more "point in time" snapshots, matching the blog post (which is hosted elsewhere), so unless we're fixing bugs in the examples I don't think that we necessarily need to change the way these are written. If anything, we'd want to ensure all of the examples here are consistently `unknown`.

PR #2314, opened 2022-03-29
Correct relative clause part-of-speech disagreement

ecleary:


ghost:
[CLA assistant check](https://cla.opensource.microsoft.com/microsoft/TypeScript-Website?pullRequest=2314) <br/>All CLA requirements met.

orta:
You might need to leave an explanation of your changes here - this was much simpler to read at `is` vs `which will make it be` which is doing the same work but with much more words

ecleary:
Ah, good point, and sorry about that.

The reason for the change is grammar correction. The following sentence is grammatically incorrect:

``A property or index signature can now be declared with the `readonly` modifier is considered read-only.``

Either of the following changes would make this sentence grammatically correct:

Version A:
``A property or index signature can now be declared with the `readonly` modifier, which will make it be considered read-only.``

Version B:
``Now a property or index signature declared with the `readonly` modifier is considered read-only.``

Version B is shorter, so perhaps version B would be preferred? If so, then I will update the PR accordingly.

ecleary:
@orta Just in case, I have modified the change to the more concise version, which also fully corrects the grammar issue.

Please feel free to either use this change or close the PR without merging, as you see fit. Thanks.

packages/documentation/copy/en/release-notes/TypeScript 2.0.md
```diff
@@ -380,7 +380,7 @@ test(() => {
 
 ## Read-only properties and index signatures
 
-A property or index signature can now be declared with the `readonly` modifier is considered read-only.
+Now a property or index signature declared with the `readonly` modifier is considered read-only.
```
jakebailey: If anything, I'd say:

```suggestion
A property or index signature can now be declared with the `readonly` modifier.
```

Because this section is only aiming to introduce the feature; the description of what it does immediately follows in the next paragraph.

jakebailey (approved):

PR #2322, opened 2022-04-06
Fix typo in Basics.md

Trott:


orta:
Thanks for the PR, but we don't want this. The convention of starting a sidenote with `note:` is consistent in the handbook.

packages/documentation/copy/en/handbook-v2/Basics.md
```diff
@@ -343,7 +343,7 @@ let msg = "hello there!";
 Even though we didn't tell TypeScript that `msg` had the type `string` it was able to figure that out.
 That's a feature, and it's best not to add annotations when the type system would end up inferring the same type anyway.
 
-> Note: the message bubble inside the code sample above. That is what your editor would show if you had hovered over the word.
+> Note the message bubble inside the code sample above. That is what your editor would show if you had hovered over the word.
```
Trott:
```suggestion
> Note: The message bubble inside the code sample above is what your editor would show if you had hovered over the word.
```

Trott:
> Thanks for the PR, but we don't want this. The convention of starting a sidenote with `note:` is consistent in the handbook.

I've adjusted the fix to accommodate that practice.

packages/documentation/copy/en/handbook-v2/Basics.md
```diff
@@ -343,7 +343,7 @@ let msg = "hello there!";
 Even though we didn't tell TypeScript that `msg` had the type `string` it was able to figure that out.
 That's a feature, and it's best not to add annotations when the type system would end up inferring the same type anyway.
 
-> Note: the message bubble inside the code sample above. That is what your editor would show if you had hovered over the word.
+> Note: The message bubble inside the code sample above is what your editor would show if you had hovered over the word.
```
Trott: I don't know if you follow [Microsoft's Style Guide](https://docs.microsoft.com/en-us/style-guide/welcome/) for docs, but I'm guessing that it's likely since TypeScript is a Microsoft project. If so, the word _above_  in this line [should be replaced with _previous_](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/a/above) or something like that. Should I change `the code sample above` to `the previous code sample`?

```suggestion
> Note: The message bubble inside the previous code sample is what your editor would show if you had hovered over the word.
```

orta: That makes sense to me 👍🏻

Trott:
OK, all set for review (I think).

RyanCavanaugh:
Toggling this to try to get tests to run...

PR #2326, opened 2022-04-08
Remove incorrect operator identification

Trott:
The resulting compiled file uses `.concat()` and not `+`. Remove `+` from prose.


orta (approved):
Agree, thanks

https://www.typescriptlang.org/docs/handbook/2/basic-types.html#erased-types

PR #2327, opened 2022-04-08
fix: type in comment

mstruebing:
I believe this is the accurate type for this comment, am I correct?


orta (approved):
Yep, looks like it to me

PR #2331, opened 2022-04-09
Update Classes.md

adilhassanm:
I have just updated a typo as "toLowerCase" in line no 342


orta:
The typo for `toLowercse` is the point of this code sample - there's no need for this PR 👍🏻

PR #2332, opened 2022-04-11
Update Gulp.md: gulp-uglify cannot handle ES6+ syntax

Leedehai:
Use [gulp-terser](https://www.npmjs.com/package/gulp-terser) instead of [gulp-uglify](https://www.npmjs.com/package/gulp-uglify). For example, the latter (current version 3.0.2) doesn't recognize ES6 [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals):
```
GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected character '`'
```
After replacing it with gulp-terser, I managed to run the build without error.

### DIFF
--- a/packages/documentation/copy/en/tutorials/Gulp.md
+++ b/packages/documentation/copy/en/tutorials/Gulp.md
@@ -6,7 +6,7 @@ oneline: Using TypeScript with Gulp
 deprecated: true
 ---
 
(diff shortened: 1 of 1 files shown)


orta:
Yeah, looks like gulp uglify hasn't been updated in 3 years - seems reasonable to me

DanielRosenwasser:
I haven't tried this yet, but this is an older guide. Does this run? I can trust you on scout's honor 😄

Leedehai:
Yeah I ran it. Trust me on scout's honor 😉

PR #2340, opened 2022-04-17
CommonJs file module's path should be relative

renjithspace:
If it is not relative it is considered a core module instead of a file module. Here 'maths.js' is a file module, note a core module that we are requiring from `node_modules` or from the standard library.  For more information please refer the file module doc: https://nodejs.org/api/modules.html#file-modules


orta:
I think the intention in this code sample is to show you a method similar to a separate node module.

I don't think it changes much particularly to switch to a relative import under the hood. So I'm a "seems harmless" for the change.

jakebailey (approved):

PR #2344, opened 2022-04-19
Add a clarification note about `typesVersions` capabilities

Andarist:


packages/documentation/copy/en/declaration-files/Publishing.md
```diff
@@ -139,6 +139,8 @@ When you want to only change the resolution for a single file at a time, you can
 
 On TypeScript 4.0 and above, an import for `"package-name"` would resolve to `./index.d.ts` and for 3.9 and below `"./index.v3.d.ts`.
 
+**Note:** This can only be used to redirect external file requests and can't be used to redirect internal ones (internal `from './foo.d.ts'` won't be redirected based on this).
```
Andarist: I'm a little bit sad that this doesn't work - but since it doesn't, it would be great to add a note like this here.

The logic to load internal requests is handled by `nodeLoadModuleByRelativeName` - and while it loads the `typesVersions` indirectly (in the call to `getPackageJsonInfo`) that information is never utilized to redirect the internal request.

jakebailey: I'm going through all of the website PRs; is this still true? Is there a change to TS that should happen? Or should we just document it?

Andarist: perhaps @andrewbranch could verify if anything I wrote here a year ago is correct or not :p cause honestly... I don't know at this point. It seems that I was debugging this quite deeply back then but I could very well draw some incorrect conclusions or smth (but maybe not)

andrewbranch: Pretty sure this is true and intentional. It never would have occurred to me that anyone would think this could happen for relative paths.

Andarist: IIRC, I tried to somehow conditionally swap one file - had a type helper that had to be implemented differently based on the current TS version.

jakebailey: Assuming it's accurate, is it worth including?

If anything, I might just change it so state more simply that "this option only affects a package's external API"?

Andarist: Feel free to reword this as you see fit.

jakebailey: A first stab:

```suggestion
Note that redirections only affect the _external_ API of a package; import resolution within a project is not affected by `typesVersions`. For example, a `d.ts` file in the previous example containing `import * as foo from "./index"` will still map to `index.d.ts`, not `index.v3.d.ts`, whereas another package importing `import * as foo from "package-name"` _will_ get `index.v3.d.ts`.
```

I was originally worried that this would be misplaced, given this is true for any kind of redirect... until I realized that in a folder redirect, relative paths of course _do_ work fine because you're already "in" the folder.

jakebailey (approved):

PR #2346, opened 2022-04-20
New introduction for Mapped Types article

daiplusplus:
The current introduction in the Mapped Types article is unclear (and what's the relevance of an _Only Fools and Horses_ reference?): the example given didn't really demonstrate what "mapping" one type to another meant by explicit example.


daiplusplus:
@orta sorry for pinging you, but as the main active contributor to TS's docs, do you know what's holding-up review and approval of my PR?

orta:
Yeah, I looked at it but removing the only fools and horses reference kinda dinged your PR from my perspective, keeping any sense of whimsy and light heartedness in the docs is a battle that sometimes I'm just not in the mood to have. We get good PRs like this that are a net positive for the individual topic but slowly make the docs really dry when going through the whole handbook.

Which makes giving feedback hard because I think it documents the syntax well, and this page is effectively the reference for the syntax even though it lives in the handbook. So I think I land on the side of losing that is a shame, it's probably worth it in this case.

WRT it getting merged though, that's on someone with write access now I'm afraid and I think the team are still figuring out what they want there.

jakebailey (commented):
One thing I'm unsure of in this change is that the intro now effectively explains everything in one go, specifically the next "mapping modifiers" section. I'm not sure if that's good or bad, without getting others to look and see what might be good for first time readers.

jakebailey:
FYI, we can't really consider taking this without the CLA stuff above; are you still interested in finishing this PR?

daiplusplus:

@jakebailey I renamed my GitHub account within the past year, looks like the CLA-Bot isn't aware of renamed GitHub accounts - is it stuck?

packages/documentation/copy/en/handbook-v2/Type Manipulation/Mapped Types.md
```diff
@@ -5,50 +5,84 @@ permalink: /docs/handbook/2/mapped-types.html
 oneline: "Generating types by re-using an existing type."
 ---
 
-When you don't want to repeat yourself, sometimes a type needs to be based on another type.
+When you don't want to repeat yourself, sometimes a type needs to be based on another type: that is, a new type with the same JavaScript object properties as another type, but modified in some way.
 
-Mapped types build on the syntax for index signatures, which are used to declare the types of properties which have not been declared ahead of time:
+For example, supposing you have a mutable `Person` type and also want a separate type for an immutable Person, and another for a partially-built Person, and another separate type to represent per-property change-flags then you would need to repeat the property list every time, with modifications:
```
jakebailey:
```suggestion
For example, suppose you have a `Person` type and also want separate types for an immutable Person and a partially-built Person, along with a type to represent per-property change-flags. You would need to repeat the property list every time:
```

packages/documentation/copy/en/handbook-v2/Type Manipulation/Mapped Types.md
```diff
@@ -5,50 +5,84 @@ permalink: /docs/handbook/2/mapped-types.html
 oneline: "Generating types by re-using an existing type."
 ---
 
-When you don't want to repeat yourself, sometimes a type needs to be based on another type.
+When you don't want to repeat yourself, sometimes a type needs to be based on another type: that is, a new type with the same JavaScript object properties as another type, but modified in some way.
```
jakebailey:
```suggestion
When you don't want to repeat yourself, sometimes a type needs to be based on another type.
```

I don't think this extra bit adds; in fact, you can make a new object that has entirely different properties which do depend on another type and so it's not "the same object properties" in that way. I think the rest of the section is pretty self explanatory.

packages/documentation/copy/en/handbook-v2/Type Manipulation/Mapped Types.md
```diff
@@ -5,50 +5,84 @@ permalink: /docs/handbook/2/mapped-types.html
...
+/** Person property is-changed flags (e.g. for generating minimal SQL UPDATE statements)  */
+type PersonChangeFlags = {
+    age: boolean;
+    name: boolean;
+    alive: boolean
 };
 ```
 
-A mapped type is a generic type which uses a union of `PropertyKey`s (frequently created [via a `keyof`](/docs/handbook/2/indexed-access-types.html)) to iterate through keys to create a type:
+However if we define `ReadOnlyPerson`, `IncompletePerson`, and `PersonChangeFlags` as types _mapped to_ `type Person` then the amount of code we need to write is drastically reduced, as well as eliminating the maintenance burden of keeping all the types' properties lists in-sync. Doing this, the above types become just this:
```
jakebailey:
```suggestion
However, if we define `ReadOnlyPerson`, `IncompletePerson`, and `PersonChangeFlags` as types _mapped to_ `type Person` then the amount of code we need to write is drastically reduced, eliminating the maintenance burden of keeping all the types' properties lists in sync. Doing this, the above types become just this:
```

I'm also considering whether or not it's more correct to say "mapped from".

packages/documentation/copy/en/handbook-v2/Type Manipulation/Mapped Types.md
```diff
@@ -5,50 +5,84 @@ permalink: /docs/handbook/2/mapped-types.html
...
+type IncompletePerson  = { [PersonPropertyName in keyof Person]+?: Person[PersonPropertyName] };
+//   ^?
+
+// Use `: boolean` to change the type of every property in Person to boolean:
+type PersonChangeFlags = { [PersonPropertyName in keyof Person]: boolean };
+//   ^?
 ```
 
-In this example, `OptionsFlags` will take all the properties from the type `Type` and change their values to be a boolean.
+Taking this a step further: supposing in addition to `type Person` we also have `type Order`, `type Product`, `type OrderItem`, and we want immutable, partial, and flags copies of all of those types, then we don't need to manually define mapped-types for those either: we can make our mapped-types generic:
```
jakebailey:
```suggestion
Taking this a step further: suppose in addition to `type Person`, we also have `type Order`, `type Product`, `type OrderItem`, and we want immutable, partial, and flags copies of all of those types. We don't need to manually define mapped-types for those either -- we can make our mapped-types generic:
```

jakebailey (commented):
Some wording tweaks.

daiplusplus:
@jakebailey Sorry I forgot about this PR for almost a year. I've accepted your proposed changes as-is (for the sake of expediency) - there are some minor changes I'd have liked to have made or added now, but GitHub's online editor just wasn't working for me today - but anyway.

PR #2348, opened 2022-04-21
Decorators are now stage 3

Mythli:
The documentation mentions, that decorators are a stage 2 proposal. As of April 2022 they are already a stage 3 proposal and so more stable than this documentation suggests.

### DIFF
--- a/packages/documentation/copy/en/reference/Decorators.md
+++ b/packages/documentation/copy/en/reference/Decorators.md
@@ -14,7 +14,7 @@ translatable: true
 
 With the introduction of Classes in TypeScript and ES6, there now exist certain scenarios that require additional features to support annotating or modifying classes and class members.
 Decorators provide a way to add both annotations and a meta-programming syntax for class declarations and members.
-Decorators are a [stage 2 proposal](https://github.com/tc39/proposal-decorators) for JavaScript and are available as an experimental feature of TypeScript.
+Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for JavaScript and are available as an experimental feature of TypeScript.
 
 > NOTE&emsp; Decorators are an experimental feature that may change in future releases.
 


packages/documentation/copy/en/reference/Decorators.md
```diff
@@ -14,7 +14,7 @@ translatable: true
 
 With the introduction of Classes in TypeScript and ES6, there now exist certain scenarios that require additional features to support annotating or modifying classes and class members.
 Decorators provide a way to add both annotations and a meta-programming syntax for class declarations and members.
-Decorators are a [stage 2 proposal](https://github.com/tc39/proposal-decorators) for JavaScript and are available as an experimental feature of TypeScript.
+Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for JavaScript and are available as an experimental feature of TypeScript.
```
orta, line 17: This change kinda implies that TS supports the stage 3 version of decorators, which isn't strictly true because there's going to be differences between the current support and the official support
 

```suggestion
Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for JavaScript, and are currently available as an experimental feature of TypeScript using an older version of the specification.
```

domdomegg:
This seems to be something people are repeatedly correcting, only to realise this PR has been blocked for 9 months.

What can we do to get this merged? I can open a PR for the content with @orta's suggested change if this PR isn't being updated?

jakebailey:
With stage 3 decorators now a part of TypeScript proper, I think that this particular change doesn't make sense anymore; I do think that we need a proper PR for this.

That PR should not simply be a bump from 2 to 3 in this page, instead it should mention that you can pass `--experimentalDecorators` to get the legacy decorators, but otherwise without any flags we now use the officially specified ones.

I'm going to close this PR to try and keep things tidy (since tis is super old); feel free to send one that more accurately represents the current state of things.

PR #2349, opened 2022-04-22
Slightly improve readability of a sentence

jtoar:
I'm not sure what the grammatical terms are but "Having been around since 2012" describes TypeScript, so it sounds a little awkward with TypeScript at the beginning of the sentence instead of in the phrase (or clause?)


orta (approved):
Seems pretty reasonable to me

jakebailey (approved):

Issue #2353, opened 2022-04-23
Docs: Make the Handbook more readable and interesting


shoaibshebi:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary & Current scenario 😯

TS is getting hit now in the development and  a lot of companies and developers are getting benefit from it. Though, there are video resources to learn TS, but no forum can be the so accurate as official docs. I have gone through the docs and found really helpful but wasting so much time to consume this Typed Language as the docs are no so much mature like other languages docs.

### New and required behavior 🤔

As you know approximately 93% of effective communication is nonverbal & emotional, while spoken words account for only 7%. And also 1 picture = 1000 words. So to make the docs more interesting, catchy and fastly consumalbe I have updated the docs with emojies, which makes the coding world more joyful and fun, so that developers can learn and develop fast.

Note: 👉 Currently I've update first two articles "The TypeScript Handbook" & "The Basics" so that you can suggest the changes.

### Examples 🌈

[_Strapi_ ](https://docs.strapi.io/developer-docs/latest/getting-started/introduction.html#open-source-contribution)

Josh-Cena:
> 1 picture = 1000 words

That's about _meaningful_ schematics. Changing "bug" to "bug 🐛" only adds extra noise.

PR #2354, opened 2022-04-23
Docs: make "The handbook.md" more readable and interesting

shoaibshebi:
### updated "The handbook.md" page making it more readable and consumable to get a higher pace for developers.

PR under the Issue: #2353

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Basics.md
+++ b/packages/documentation/copy/en/handbook-v2/Basics.md
@@ -4,11 +4,11 @@ layout: docs
 permalink: /docs/handbook/2/basic-types.html
 oneline: "Step one in learning TypeScript: The basic types."
 preamble: >
-  <p>Welcome to the first page of the handbook. If this is your first experience with TypeScript - you may want to start at one of the '<a href='https://www.typescriptlang.org/docs/handbook/intro.html#get-started'>Getting Started</a>' guides</a>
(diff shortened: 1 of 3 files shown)


orta (changes requested):
Thanks for taking a shot at improving the docs, but I don't think emoji-filled text is really the tone we use for the TypeScript docs.

shoaibshebi:
@orta Thanks for the feedback. Firstly being a JS developer I have read all the Handbook and other docs and practiced it using Javascript and React. Then I feel it really needs the emojis due to difficulty to consume. And after spending almost 4-5 hours 🤯, on only these two pages "the handbook.md" and "the basics.md" I chose the emojies for each line after sensing the tones of lines.

So it's a humble request to you 🙏, that you can suggest the lines of doc, what changes could be made to improve it 😇. As I am on daily basics work with TS as I'm a JS developer, I found it difficult without emojis, and of course a lot of other people found, as I got the feedback from my colleagues.

shoaibshebi:
I tried to using tech emojies, and not like only Yellow ones. But where the tone was about joy and sad I used the yellowish one as well.

So please suggest the changes 🙏, so I can follow for the next pages. As I want to update the whole handbook 📖.

Thanks

orta:
Sure, my feedback is that I would recommend not spending more of your time on the pull request and you should probably close this because it's unlikely to get merged. 

Emojis have their place in written communication, but professional technical documentation shouldn't have an emoji per paragraph, it makes it read like a youtuber wrote it where they have to keep reacting to the words they just said to add emphasis. It's cool if that's your writing style, the internet is a big place and that allows for all sorts of ways to write - but it's not the TypeScript team's.

shoaibshebi:
Okay, @orta. Thanks for the feedback and I will close the PR.

PR #2359, opened 2022-04-28
Fix the wrong output type in release-notes TypeScrtipt2.8

dana322:


orta:
Confirmed:
image

jakebailey:
I don't think that we should change this; it's very likely accurate for TS 2.8. Going back to 3.3 (the oldest still usable on the playground), we can see `string & number`:

image

This behavior appears to have changed in TS 3.6.

PR #2362, opened 2022-04-30
Update Decorators.md

alexbs:
proposal is now at "stage 3".
This can be verified here: github.com/tc39/proposal-decorators


orta:
Thanks, we have an open PR with discussion: https://github.com/microsoft/TypeScript-Website/pull/2348

PR #2364, opened 2022-05-01
Use const instead of let

FalkWoldmann:
This will keep this example consistent with the other ones.


orta:
Thanks, but this is intentional (one to keep the literal types from needing to be  explained, but also because a lot of people prefer writing `let` in JS (which is at least an improvement from `var`. )

FalkWoldmann:
Alright, that makes sense :-)

PR #2365, opened 2022-05-02
explain `number` being a type in terms of practice

S11001001:
Previously it was said that

> Ultimately, a static type system has to make the call over what code should be flagged as an error in its system, even if it's "valid" JavaScript that won't immediately throw an error.

So "JavaScript does not have a special runtime value for integers" is not a reason for int and float not to exist—that is a choice made by TypeScript, not one imposed by JavaScript.

I've tried to provide an alternate reason instead, in terms of what TypeScript users are likely to want to do.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -21,7 +21,7 @@ Each has a corresponding type in TypeScript.
 As you might expect, these are the same names you'd see if you used the JavaScript `typeof` operator on a value of those types:
 
 - `string` represents string values like `"Hello, world"`
-- `number` is for numbers like `42`. JavaScript does not have a special runtime value for integers, so there's no equivalent to `int` or `float` - everything is simply `number`
+- `number` is for numbers like `42`. JavaScript users do not commonly distinguish between integers and fractional or floating-point numbers, so there's no equivalent to `int` or `float` - everything is simply `number`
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Everyday Types.md
```diff
@@ -21,7 +21,7 @@ Each has a corresponding type in TypeScript.
 As you might expect, these are the same names you'd see if you used the JavaScript `typeof` operator on a value of those types:
 
 - `string` represents string values like `"Hello, world"`
-- `number` is for numbers like `42`. JavaScript does not have a special runtime value for integers, so there's no equivalent to `int` or `float` - everything is simply `number`
+- `number` is for numbers like `42`. JavaScript users do not commonly distinguish between integers and fractional or floating-point numbers, so there's no equivalent to `int` or `float` - everything is simply `number`
```
jakebailey, line 24: I'm sorta iffy on this explanation; I would instead argue that TypeScript has a `number` type because JavaScript has `typeof x === "number"`. Under the hood, runtimes _do_ have integers, you can do `| 0` to explicitly tell the runtime that, and so on.

It is of course possible that TS could have had both `int` and `float`, but we'd never be able to narrow to either of them via `typeof`, so there would always need to be a superset of both categories. JS isn't like Python where the two types are distinct and checkable at runtime by something other than checking the value itself.

jakebailey (changes requested):

PR #2374, opened 2022-05-05
Add `import` in the non-module description

myogh:
- I think `await` doesn't play a role in deciding whether a JS file is a module or a script.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -25,7 +25,7 @@ Conversely, to consume a variable, function, class, interface, etc. exported fro
 ## Non-modules
 
 Before we start, it's important to understand what TypeScript considers a module.
-The JavaScript specification declares that any JavaScript files without an `export` or top-level `import` should be considered a script and not a module.
+The JavaScript specification declares that any JavaScript files without an `export` or top-level `await` should be considered a script and not a module.
 
 Inside a script file variables and types are declared to be in the shared global scope, and it's assumed that you'll either use the [`outFile`](/tsconfig#outFile) compiler option to join multiple input files into one output file, or use multiple `<script>` tags in your HTML to load these files (in the correct order!).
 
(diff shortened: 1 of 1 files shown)


orta:
I think this is previously accurate because only modules are allowed [top level await](https://github.com/tc39/proposal-top-level-await) but I think it's reasonable to add `import` to this list (instead of replacing await)

myogh:
Thank you @orta, I was not aware of the "top level await" case. Thanks for mentioning it. Today I learned something new.

jakebailey:
I think this one is tricky; having a top-level await doesn't cause TS to treat the file as a module. Instead, we use `import`/`export` or the `moduleDetection` setting to determine that, and then if we see a top-level await that shouldn't be there, we error.

But, I think this section is already awkward... it's not that the _JavaScript spec_ says anything about these things, right?

packages/documentation/copy/en/handbook-v2/Modules.md
```diff
@@ -25,7 +25,7 @@ Conversely, to consume a variable, function, class, interface, etc. exported fro
 ## Non-modules
 
 Before we start, it's important to understand what TypeScript considers a module.
-The JavaScript specification declares that any JavaScript files without an `export` or top-level `await` should be considered a script and not a module.
+The JavaScript specification declares that any JavaScript files without an `import` or `export` or top-level `await` should be considered a script and not a module.
```
andrewbranch:
```suggestion
The JavaScript specification declares that any JavaScript files without an `import` declaration, `export`, or top-level `await` should be considered a script and not a module.

```

andrewbranch: A dynamic `import("foo")` is allowed in a script

andrewbranch:
I’m not an expert on the ECMAScript grammar but I suspect the wording is a little weird here. The spec does distinguish between the grammars for script bodies and module bodies, and these three things are indeed relevant, but I think our wording probably implies the wrong causality. My understanding is that spec-compliant engines know _ahead_ of time whether they should be parsing a script or a module (via `<script type="module"` in the browser, or file extension / package.json `"type"` in Node, for example), and that decision determines how the rest of the file is parsed, which makes import/export/TLA a syntax error in scripts. This is different from how TypeScript parses, where it has the ability to parse either one until an import or export is seen. That said, I don’t think it’s worth diving that far into the weeds here. This page will likely be replaced via https://github.com/microsoft/TypeScript/issues/52593.

PR #2377, opened 2022-05-11
Update TypeScript 4.3.md - Contextual Narrowing for Generics

Messiahhh:


orta:
I think you might need to explain why you're requesting these changes

Messiahhh:
> If hypothetically TypeScript tried that approach, it would break the above example in a different way. At the return positions, where the function expects values with the type C, we would instead get a Set<T> and a T[] in each branch, which TypeScript would reject.

This code snippet is used to show what would happen if `TypeScript tried that approach`.

Also the error `'C' could be instantiated with a different subtype of constraint 'Set<T> | T[]'` means the type of return value of this function should be `C` rather than `Set<T> | T[]` , It's also weird that the function doesn't has any 'C' while the error log show 'C'

jakebailey (changes requested):
Throwing the new code into the playground, it doesn't show the errors that the text implies: [Playground Link](https://www.typescriptlang.org/play?ssl=9&ssc=22&pln=10&pc=44#code/GYVwdgxgLglg9mABAWwIYGsCmBVMMCOImAPACoA0iAwopgB5SZgAmAzogMqZRkB8iAH0SkA2gF1eACgBQiRBDgAbRZmjwwALmrlZ8uMgAOqAE6ZjWyXS0VEAT2sBKRAF5+YEMgBGZ6Q600Ab10AemDEAFETRVtET1QYRQBaOBAoRBhgRAB3TEQAC1QAN1zUTm4AOhCwgHUS1lYPXKg83MwVZCYodhMSxVNUZhjwAiJKuQzESQVlVVgEdLBWKFRITDhMrignILk5UygQYyRplTUEAG5dOVDduQA-B8enq8Qbs2M4c2FbA1yAck2fD+6XYYDgaVQ9RgAHMwKhPCpEFA4Eifv8qH8xrsbrdcQDuECQYhIawYXCEU0Uc1cgpFlBjPEwGl1qjfog-hjKJ5Ui8cbjdhzgQoQIpmLFcjA6StYKhGGKsjBmsTEMwMsAzJ1EA1PFA0YgWbSlgzJWl8TxSPwhKIxJjdABfXRVRDlF2OvbcQ7HJSnOZgS7XMK3J7Bu5O96fax6v7W4EwUHg4lQ2HwxHI1no20B-m3aPiWPdJPk1NUlp6OnGpn6zK6tmCrk8rPZuSCvQisXeBZLaUwWWYeWKvLK1XAdWmSvamu5A0II2M02Ai2CYR5yp2oA)

PR #2381, opened 2022-05-15
Fixed links in "What's New" section for fragments starting with a number

### LINKED ISSUE #2369: Docs: Release notes section starting with number not reachable from table of content
**Page URL:** [TypeScript 3.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html)  
Note that this issue is also technically applicable to any other release note page, tho only this one has a section that causes trouble.

**Issue:**  
When using the table of content of the TypeScript release notes (on the right side of the screen on pc), clicking the name of a section brings you to the corresponding section (further down) in the release notes.
While this works for sections with a name starting with a letter or a special character (some of those special characters are being removed from the url fragment), it doesn't work for sections starting with numbers.

**Reproduce:**  
Go to the [TypeScript 3.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html).
Try to go to the "3.7 Breaking Changes" section using the table of content on the right side.
Same with the "3.7 API Changes" section.
It works fine with all other sections.

**Source:**
https://github.com/microsoft/TypeScript-Website/blob/427b14c35f65a80a050abf58906a2fbcfc63aac9/packages/typescriptlang-org/src/templates/scripts/setupSubNavigationSidebar.ts#L7-L17

While having an URL fragment that starts with a number is completely fine,
and getting the hash part, that starts with a number, of an URL is fine too,
having an element with an id starting with one isn't.

Having an element with an illegal id is not that big of a deal on its own, but trying to retrieve it (by using the querySelector method) is.

**Stack trace:**
```trace
setupSubNavigationSidebar.ts:11 Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#37-api-changes' is not a valid selector.
    at HTMLAnchorElement.<anonymous> (https://www.typescriptlang.org/component---src-templates-documentation-tsx-8809c4b3b2cbd592e9fc.js:1:121516)
```

**Recommended Fix:**
~~Using legal id's for elements.~~
~~This can either be done by~~
- ~~not using leading numbers section titles,~~
- ~~prefixing the leading number to make the id valid~~
- ~~removing the leading number (as it is for some special characters, as with "@" or "/")~~

~~Note that chaning the ids for some sections would break the existing links leading to them (mainly talking about links in bookmarks or on random articles, outside of the scope of the TS website).~~

**Edit:**
By getting the element using a CSS attribute selector-like method (`[attribute=value]`), we can find elements with ids staring with numbers.
We just need to remove the leading hash sign in the hash from the anchror element and it should be fixed.

```ts
subnavLinks.forEach(link => {
    link.addEventListener("click", event => {
        event.preventDefault();

        const hash = (event.target! as HTMLAnchorElement).hash;
        const id = decodeURIComponent(hash).slice(1);
        const target = document.querySelector(`[id="${id}"]`);
        
        target!.scrollIntoView({ behavior: "smooth", block: "start" });
        document.location.hash = hash;
    });
});
```

Ascor8522:
Fixes https://github.com/microsoft/TypeScript-Website/issues/2369

Links in the "What's New" pages refering to sections with a fragment starting with a number weren't working.

It was due to the fact we were trying to get the anchor tag with the corresponding id, and since the id isn't meeting to the specs, trying to get it using `Document.getElementById()` was throwing an error.

There is fortunately a workaround by using `querySelector` and a CSS-like attribute selector, with the id as attribute.

### DIFF
--- a/packages/typescriptlang-org/src/templates/scripts/setupSubNavigationSidebar.ts
+++ b/packages/typescriptlang-org/src/templates/scripts/setupSubNavigationSidebar.ts
@@ -8,11 +8,12 @@ export const overrideSubNavLinksWithSmoothScroll = () => {
     link.addEventListener("click", event => {
       event.preventDefault()
 
-      let target = document.querySelector(
-        decodeURIComponent(event.target!["hash"])
-      )
+      const hash = (event.target! as HTMLAnchorElement).hash
+      const id = decodeURIComponent(hash).slice(1)
+      const target = document.querySelector(`[id="${id}"]`)
+        
       target!.scrollIntoView({ behavior: "smooth", block: "start" })
-      document.location.hash = event.target!["hash"]
+      document.location.hash = hash
     })
   })
 }


orta (approved):
Looks reasonable to me, thanks!

PR #2383, opened 2022-05-17
Update TS for JS Programmers.md

amit12cool:
Corrected spelling mistake. `conforms` -> `confirms`


orta (changes requested):
Thanks for trying to improve the docs, but conforms is correct here

PR #2389, opened 2022-05-22
docs: typo

ghost:


packages/documentation/copy/en/get-started/TS for the New Programmer.md
```diff
@@ -45,7 +45,7 @@ Every language has its own _quirks_ — oddities and surprises, and JavaScript's
   ```js
   const obj = { width: 10, height: 15 };
   // Why is this NaN? Spelling is hard!
-  const area = obj.width * obj.heigth;
+  const area = obj.width * obj.height;
```
agilgur5, line 48: per the comment above, these are _intentional_ spelling errors to demonstrate an error

agilgur5 (changes requested):

PR #2394, opened 2022-05-26
fix: add missing links to `exclude` property

agilgur5:
## Summary

Link to `exclude` option in TSConfig Reference when `exclude` is mentioned in the docs

## Details

- it was referenced in many places, but never had a link to the TSConfig Reference in those references
  - `include` and `files` always had links, so mirror that for `exclude`

- also add some oxford commas for clarity


orta (approved):
Great idea

packages/documentation/copy/en/release-notes/TypeScript 1.6.md
```diff
@@ -439,7 +439,7 @@ if (isCat(x)) {
 }
 ```
 
-## `exclude` property support in tsconfig.json
+## [`exclude`](/tsconfig#exclude) property support in tsconfig.json
```
jakebailey:
```suggestion
## `exclude` property support in tsconfig.json
```

This is a heading and cannot contain a link.

jakebailey (approved):

packages/documentation/copy/en/release-notes/TypeScript 1.6.md
agilgur5: As in, the renderer doesn't support links in headings? Markdown itself does support that

If so a lint check for that would be good to add to the automation

jakebailey: Maybe it can, but I don't think we have any examples of that. That and sometimes headings are clickable themselves and this would conflict. I don't think it's worth it for this heading.

PR #2397, opened 2022-05-27
Use `in` operator for type predicates

dirtyhenry:
I was wondering if `'swim' in pet` would make more sense here since it is more readable. What do you think?


tryonelove:
Generally speaking, if `swim` is going to be renamed in the future, it will be much harder to spot the bug. Even though we're talking about such small piece of code.

dirtyhenry:
That is a really good point @tryonelove. Closing the pull request.

PR #2404, opened 2022-05-31
Update Generics.md

MFredX:
There seems to be a typo in the section from lines 152-162. The code snippet in the section describes that a different name for the generic parameter can be used as long as the type variables match up.

However the code snippet in lines 155-159, do not use a different name as mentioned in the preceding paragraph in the example; it uses the `Type` generic parameter in the instead of the `Input` generic parameter in the `identity` function's declaration unlike the `myIdentity` function which uses `Input`.


orta (approved):
Agree, thanks!

jakebailey (approved):

PR #2408, opened 2022-06-02
More accurately describe the expected `index.d.ts` location

rauno56:
The docs say that the expected location for `index.d.ts` is in "the root of the package where the `index.js`" is. It's contradicting since the `index.js` is usually buried somewhere in a `dist` or `build` folder. However `package.json` is in the root.

**Does the resolver look for the `index.d.ts` file in the project root or next to the file specified by `main` property in the `package.json`?** Or both?

Either way the documentation could be more clear on that.

### DIFF
--- a/packages/documentation/copy/en/declaration-files/Publishing.md
+++ b/packages/documentation/copy/en/declaration-files/Publishing.md
@@ -33,7 +33,7 @@ For example:
 
 Note that the `"typings"` field is synonymous with `types`, and could be used as well.
 
-Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to the `package.json`) you do not need to mark the `types` property, though it is advisable to do so.
+Note:  TypeScript will infer the root `*.d.ts` based on the value in `main`, which means in cases like above you do not need to mark the `types` property, though it is advisable to do so.
 
 ## Dependencies
 
--- a/packages/documentation/copy/en/declaration-files/Publishing.md
+++ b/packages/documentation/copy/en/declaration-files/Publishing.md
@@ -33,7 +33,7 @@ For example:
 
 Note that the `"typings"` field is synonymous with `types`, and could be used as well.
 
-Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to `index.js`) you do not need to mark the `types` property, though it is advisable to do so.
+Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to the `package.json`) you do not need to mark the `types` property, though it is advisable to do so.
 
 ## Dependencies
 
--- a/packages/documentation/copy/en/declaration-files/Publishing.md
+++ b/packages/documentation/copy/en/declaration-files/Publishing.md
@@ -33,7 +33,6 @@ For example:
 
 Note that the `"typings"` field is synonymous with `types`, and could be used as well.
 
-Note:  TypeScript will infer the root `*.d.ts` based on the value in `main`, which means in cases like above you do not need to mark the `types` property, though it is advisable to do so.
 
 ## Dependencies
 


packages/documentation/copy/en/declaration-files/Publishing.md
```diff
@@ -33,7 +33,7 @@ For example:
 
 Note that the `"typings"` field is synonymous with `types`, and could be used as well.
 
-Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to `index.js`) you do not need to mark the `types` property, though it is advisable to do so.
+Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to the `package.json`) you do not need to mark the `types` property, though it is advisable to do so.
```
orta: I think this change is more confusing, the `index.d.ts` would be found relative to the field defined in `main` which is being described as `index.js` here - not relative to the `package.json`.  Perhaps:

```suggestion
Note:  TypeScript will infer the root `*.d.ts` based on the value in `main`, which means in cases like above you do not need to mark the `types` property, though it is advisable to do so.
```

rauno56: Not only was my change confusing, it was wrong. ;) it was also how I assumed it's working from reading the docs.
Thanks for the suggestion and making it clear for me, and hopefully others.

Why `*.d.ts`? Is there a list of files that are searched for, @orta?

orta: The filename is dynamic because it could be anything in the `"main"` - the * could also be `[file].d.ts` and I think it would also make sense

orta: The filename is dynamic because it could be anything in the `"main"` - the * could also be `[file].d.ts` and I think it would also make sense

rauno56: OK, now understand.
"main" is transformed into a form of `*.d.ts` by stripping the extension from the filename in "main" and replacing it with ".d.ts".

rauno56:
This is ready for a review from my end. Let me know if you are looking for any changes to be made.

rauno56:
I guess this has been forgotten for a while now. Any intention still merging it?

jakebailey:
> **Does the resolver look for the `index.d.ts` file in the project root or next to the file specified by `main` property in the `package.json`?** Or both?

Any and all files will get sibling `d.ts` file lookup, so if your main were `something.js`, and you stuck `something.d.ts` next to it, TS will find it even if you don't define it in `package.json`.

I don't think this PR is wrong per se, but it's not complete either. Probably, this should just wait until the module documentation is updated with all of the fine details and recommendations.

jakebailey (changes requested):
Marking as needing changes just so we hold this back; I think we should wait for the module doc update per above.

packages/documentation/copy/en/declaration-files/Publishing.md
```diff
@@ -33,7 +33,7 @@ For example:
 
 Note that the `"typings"` field is synonymous with `types`, and could be used as well.
 
-Also note that if your main declaration file is named `index.d.ts` and lives at the root of the package (next to `index.js`) you do not need to mark the `types` property, though it is advisable to do so.
+Note:  TypeScript will infer the root `*.d.ts` based on the value in `main`, which means in cases like above you do not need to mark the `types` property, though it is advisable to do so.
```
andrewbranch: Your change is correct, but I think we should just remove this note. It’s kind of silly to go out of our way to say that something is unnecessary, but “advisable,” without even explaining why it’s advisable. We want people to do this purely because it’s the thing that makes npmjs.com show the little TS icon on typed packages. It’s fine if people don’t even know they can skip it.

```suggestion
```

PR #2414, opened 2022-06-05
Improve readability of the code samples

enheit:
- Improve readability of code samples
- Replace `await` with `import` misspelled keyword


orta:
Hi, thanks for trying to improve the TS docs - but the top level `await` you changed was correct, and I don't think we probably need the whitespace / cut changes to the code samples

PR #2416, opened 2022-06-06
support for static classes was added to C#

suryanshmak:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Classes.md
+++ b/packages/documentation/copy/en/handbook-v2/Classes.md
@@ -887,7 +887,7 @@ class S {
 
 ### Why No Static Classes?
 
-TypeScript (and JavaScript) don't have a construct called `static class` the same way as, for example, C# does.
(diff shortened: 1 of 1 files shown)


orta (approved):

packages/documentation/copy/en/handbook-v2/Classes.md
```diff
@@ -887,7 +887,7 @@ class S {
 
 ### Why No Static Classes?
 
-TypeScript (and JavaScript) don't have a construct called `static class` the same way as, for example, C# does.
```
jakebailey, line 890: I don't think that this change is correct; the wording here is saying "JavaScript doesn't have static classes like C# does", which implies that C# supports static classes.

jakebailey (changes requested):
See above; another wording may be better but the sentence to me reads correctly.

PR #2417, opened 2022-06-08
add throw to documentation of noFallthroughCasesInSwitch option

leptonite:
The documentation of the `noFallthroughCasesInSwitch` option only mentions `break` and `return`. This PR adds the missing `throw`.

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/noFallthroughCasesInSwitch.md
+++ b/packages/tsconfig-reference/copy/en/options/noFallthroughCasesInSwitch.md
@@ -4,7 +4,7 @@ oneline: "Enable error reporting for fallthrough cases in switch statements."
 ---
 
 Report errors for fallthrough cases in switch statements.
-Ensures that any non-empty case inside a switch statement includes either `break` or `return`.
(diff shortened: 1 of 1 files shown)


orta (approved):
Agree, nice thinking

packages/tsconfig-reference/copy/en/options/noFallthroughCasesInSwitch.md
```diff
@@ -4,7 +4,7 @@ oneline: "Enable error reporting for fallthrough cases in switch statements."
 ---
 
 Report errors for fallthrough cases in switch statements.
-Ensures that any non-empty case inside a switch statement includes either `break` or `return`.
+Ensures that any non-empty case inside a switch statement includes either `break`, `return` or `throw`.
```
jakebailey:
```suggestion
Ensures that any non-empty case inside a switch statement includes either `break`, `return`, or `throw`.
```

jakebailey: (IIRC we try to be consistent about using the oxford comma, but, willing to be wrong.)

jakebailey (approved):

PR #2419, opened 2022-06-11
Add version compatibility details

Evoque:
In 'Anything but primitive assignable to { [x: string]: unknown }', When version > 3.0, x = [1, 2, 3] will throw Error


jakebailey:
This page is the release notes for TS 3.0; the fact that this works in 3.0 and above is already inherent to the content of the page and the full example introducing `unknown`.

PR #2421, opened 2022-06-13
Note that object literals may only specify known properties

thisisrandy:
Re: microsoft/TypeScript#49500

I'm not super happy with the object literals link, but it was the best I could find in the official docs. To anyone reviewing this, please update if there is a better link available.

### DIFF
--- a/packages/documentation/copy/en/reference/Type Compatibility.md
+++ b/packages/documentation/copy/en/reference/Type Compatibility.md
@@ -70,7 +70,7 @@ greet(dog); // OK
 Note that `dog` has an extra `owner` property, but this does not create an error.
 Only members of the target type (`Pet` in this case) are considered when checking for compatibility.
 
-Be aware, however, that [object literals](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#other-important-typescript-types)
+Be aware, however, that [object literals](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks)
 may only specify known properties. For example, because we have explicitly specified that `dog` is
 of type `Pet`, the following code is invalid:
 
--- a/packages/documentation/copy/en/reference/Type Compatibility.md
+++ b/packages/documentation/copy/en/reference/Type Compatibility.md
@@ -68,7 +68,9 @@ greet(dog); // OK
 ```
 
 Note that `dog` has an extra `owner` property, but this does not create an error.
-Only members of the target type (`Pet` in this case) are considered when checking for compatibility.
+Only members of the target type (`Pet` in this case) are considered when
+checking for compatibility. This comparison process proceeds recursively,
+exploring the type of each member and sub-member.
 
 Be aware, however, that object literals [may only specify known properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks).
 For example, because we have explicitly specified that `dog` is
@@ -77,8 +79,6 @@ of type `Pet`, the following code is invalid:
 ```ts
 let dog: Pet = { name: "Lassie", owner: "Rudd Weatherwax" }; // Error
 ```
-
-This comparison process proceeds recursively, exploring the type of each member and sub-member.
 
 ## Comparing two functions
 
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -70,6 +70,14 @@ greet(dog); // OK
 Note that `dog` has an extra `owner` property, but this does not create an error.
 Only members of the target type (`Pet` in this case) are considered when checking for compatibility.
 
+Be aware, however, that [object literals](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#other-important-typescript-types)
```
jakebailey: Er, what we should really have is a section on excess property checks... we used to have one in the old handbook, but apparently there is no current alternative? https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks

jakebailey: Aha, #2242.

jakebailey: After #2242 is merged, we should just link to that doc.

jakebailey (changes requested):

packages/documentation/copy/en/reference/Type Compatibility.md
thisisrandy: Done. Please see the latest commit.

packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -70,6 +70,14 @@ greet(dog); // OK
...
 
+Be aware, however, that [object literals](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks)
+may only specify known properties. For example, because we have explicitly specified that `dog` is
+of type `Pet`, the following code is invalid:
+
+```ts
+let dog: Pet = { name: "Lassie", owner: "Rudd Weatherwax" }; // Error
+```
+
 This comparison process proceeds recursively, exploring the type of each member and sub-member.
```
jakebailey: I would move this sentence back up above your newer text, i.e. add your stuff to the bottom of this section.

packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -70,6 +70,14 @@ greet(dog); // OK
 Note that `dog` has an extra `owner` property, but this does not create an error.
 Only members of the target type (`Pet` in this case) are considered when checking for compatibility.
 
+Be aware, however, that [object literals](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks)
+may only specify known properties. For example, because we have explicitly specified that `dog` is
```
jakebailey:
```suggestion
Be aware, however, that object literals [may only specify known properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks).
For example, because we have explicitly specified that `dog` is
```

Just a nit, but we're linking to the bit about excess properties, so that's where I'd stick the link.

jakebailey (commented):
Two last comments.

packages/documentation/copy/en/reference/Type Compatibility.md
thisisrandy: Good point. I'd go further and say it's properly part of the preceding paragraph (Note that dog...), so I'm going to merge them.

packages/documentation/copy/en/reference/Type Compatibility.md
```diff
@@ -68,9 +68,17 @@ greet(dog); // OK
 ```
 
 Note that `dog` has an extra `owner` property, but this does not create an error.
-Only members of the target type (`Pet` in this case) are considered when checking for compatibility.
+Only members of the target type (`Pet` in this case) are considered when
+checking for compatibility. This comparison process proceeds recursively,
+exploring the type of each member and sub-member.
 
-This comparison process proceeds recursively, exploring the type of each member and sub-member.
+Be aware, however, that object literals [may only specify known properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks).
```
jakebailey:
```suggestion
Be aware, however, that object literals [may only specify known properties](/docs/handbook/2/objects.html#excess-property-checks).
```

jakebailey (approved):

PR #2431, opened 2022-06-25
Adds a delayed call to re-measure the fonts used in the playground

### LINKED ISSUE #2427: Playground should call `monaco.editor.remeasureFonts()` after font loads and changes
I can see a slight change of space width while highlighting the text. I wonder if anyone ever reported this before.
Calling `monaco.editor.remeasureFonts()` manually in the console solves the issues.
(This API was introduced by https://github.com/microsoft/monaco-editor/issues/392)
That is all.

orta:
Fixes #2427 - and includes a tiny whitespace change.

The fixes the difference in font width when selecting text in the playground. 5 seconds after booting up is a reasonable enough timeout, because the font name comes through a css variable.

### DIFF
--- a/packages/playground/src/index.ts
+++ b/packages/playground/src/index.ts
@@ -191,21 +191,21 @@ export const setupPlayground = (
       const lenses = !showFileCodeLens
         ? []
         : [
-          {
-            range: {
-              startLineNumber: 1,
-              startColumn: 1,
-              endLineNumber: 2,
-              endColumn: 1,
+            {
+              range: {
+                startLineNumber: 1,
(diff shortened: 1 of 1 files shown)


jakebailey (approved):

packages/playground/src/index.ts
```diff
@@ -191,21 +191,21 @@ export const setupPlayground = (
       const lenses = !showFileCodeLens
         ? []
         : [
-          {
```
jakebailey, line 194: Blah, this code wasn't formatted properly in the first place. I guess it's fine to do this here.

PR #2434, opened 2022-06-27
Fixes #2131: code snippets are clipped in the PDF

### LINKED ISSUE #2131: Docs: Code snippet is cut off on page 60 in the PDF
For PDF created on Monday, November 1, 2021 against commit 5983a1 with TypeScript 4.4.

The code snippet on the top of page 60 reads: 

    function map<Input, Output>(arr: Input[], func: (arg: Input) => Output): O
        return arr.map(func);
    }

Based on the documentation (https://www.typescriptlang.org/docs/handbook/2/functions.html#inference), it should be:

    function map<Input, Output>(arr: Input[], func: (arg: Input) => Output): Output[] {
      return arr.map(func);
    }

OnkarRuikar:
The transpiled code is also getting clipped.
We simply need to enable wrapping on the code lines so that it'll work with .epub(variable page widths) as well.

## Screen shots

**Before fix**:
img

***After fix**:
img

### Transpiled code

**Before fix:**
before fix

**After fix:**
img

### DIFF
--- a/packages/handbook-epub/script/createPDF.ts
+++ b/packages/handbook-epub/script/createPDF.ts
@@ -87,6 +87,10 @@ pre .error-behind {
   color: white;
 }
 
+.code-container .line {
+  white-space: pre-wrap;
+}
+
     `;
   return css + thisCSS;
 };


orta (approved):
Agree, this feels better - thanks

jakebailey:
The PDF code has been dropped; so this PR no longer applies, I'm afraid.

PR #2436, opened 2022-06-29
Fix incorrect type for `compilerOptions.types`

andrsdt:
The attribute is expected to be an array of types, even when a single type is provided. This is inconsistent with [the `tsconfig` docs](https://www.typescriptlang.org/tsconfig#types)


andrsdt:
The docs refer to `package.json`, where a `string` is expected. Closing now.

PR #2437, opened 2022-06-30
Support publishing @typescript/sandbox package to NPM

oliverdunk:
## Overview

I was recently working on a project (https://github.com/oliverdunk/web-extension-playground) and to use the `@typescript/sandbox` package, I had to load it from the CDN and then vendor the types: https://github.com/oliverdunk/web-extension-playground/blob/main/src/utils/sandbox.ts#L39. I spoke to @andrewbranch about this, and he encouraged me to open a PR making this package publishable to NPM.

## Description

The changes here are fairly simple:

- Use tsdx (already used by `@typescript/vfs`) to build both a CJS and ESM bundle
- Add `@rollup/plugin-commonjs` to allow lzstring to be correctly loaded
- Update the package.json file to add more useful metadata for NPM

A potentially surprising change here is the rename from `releases.ts` to `release_data.ts`. Since there was both `releases.ts` and `releases.json` in the same directory, Rollup was resolving `import ... from "./releases` as a JSON import and failing the build. I looked at solving this but it ended up with a lot of deep configuration overriding to prefer TS files over JSON and didn't feel worth the effort when I could easily rename one of the files.

## Publishing to NPM

I did a bit of investigation to see how packages are published to NPM and what setup would be needed. This appears to be handled by a package called pleb: https://github.com/microsoft/TypeScript-Website/blob/v2/.github/workflows/v2-merged-staging.yml#L65. As far as I can tell, simply removing the private field from the package.json (as done here) should be enough to get things going.

Note: The README mentioned wanting tests to ensure a packaged build wouldn't break. I'm not quite sure what that would look like so I haven't done it here. I'm happy to add them if it's something we want and there is a clear idea about how to do it. At the same time, I can't see anything like that for the other packages so I feel like it might be fine without.

### DIFF
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -37,7 +37,7 @@
     "@typescript/vfs": "1.3.5"
   },
   "devDependencies": {
-    "@rollup/plugin-commonjs": "^22.0.1",
+    "@rollup/plugin-commonjs": "^13.0.1",
     "@types/jest": "^25.1.3",
     "jest": "*",
     "monaco-editor": "^0.32.1",
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -15,6 +15,13 @@
   "main": "./dist/index.js",
   "module": "./dist/sandbox.esm.js",
   "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/sandbox.esm.js",
+      "default": "./dist/index.js"
+    }
+  },
   "files": [
     "dist"
   ],
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -1,16 +1,27 @@
 {
   "name": "@typescript/sandbox",
   "version": "0.1.0",
-  "main": "dist/index.js",
-  "private": true,
   "license": "MIT",
-  "types": "../../typescriptlang-org/static/js/sandbox/index.d.ts",
+  "author": "TypeScript team",
+  "homepage": "https://github.com/microsoft/TypeScript-Website",
+  "repository": {
+    "url": "https://github.com/microsoft/TypeScript-Website.git",
+    "directory": "packages/sandbox",
+    "type": "git"
+  },
+  "bugs": {
+    "url": "https://github.com/microsoft/TypeScript-Website/issues"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/sandbox.esm.js",
+  "types": "./dist/index.d.ts",
   "files": [
     "dist"
   ],
   "scripts": {
     "bootstrap": "node script/downloadReleases.js",
-    "build": "tsc",
+    "build": "tsdx build --tsconfig tsconfig.lib.json && yarn make-for-website",
+    "make-for-website": "tsc",
     "test": "jest"
   },
   "jest": {
@@ -26,11 +37,13 @@
     "@typescript/vfs": "1.3.5"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^22.0.1",
     "@types/jest": "^25.1.3",
     "jest": "*",
     "monaco-editor": "^0.32.1",
     "monaco-typescript": "^3.7.0",
     "ts-jest": "^26.4.4",
+    "tsdx": "^0.14.1",
     "typescript": "*"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -76,5 +76,5 @@ packages/documentation/output/attribution.json
 .yarn/install-state.gz
 
 packages/sandbox/src/releases.json
-packages/sandbox/src/releases.ts
+packages/sandbox/src/release_data.ts
 packages/typescriptlang-org/src/lib/documentationNavigation.ts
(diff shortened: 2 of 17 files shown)


orta:
Yeah, deploy infra doesn't need tests 👍🏻 

Structurally, this looks fine and because the sandbox is independent of the typescript version then we don't to find a way to make it update with each TS version update, so this PR should be cool to go. 

That said, I had been recently wondering about moving this repo in the opposite direction and try to remove deployment to npm from the TS website repo. Since I left there's basically been no documentation update PRs merged and the more deployment infra on this repo means the less chance of me being able to get https://github.com/microsoft/TypeScript-Website/pull/2286 merged so that we can get ~80 PRs since feb upstreamed. 

Though the sandbox isn't in too active development, so its doesn't hurt to send it off an npm deploy anyway while that is still getting figured out

packages/sandbox/package.json
```diff
@@ -1,16 +1,27 @@
...
+  "repository": {
+    "url": "https://github.com/microsoft/TypeScript-Website.git",
+    "directory": "packages/sandbox",
+    "type": "git"
+  },
+  "bugs": {
+    "url": "https://github.com/microsoft/TypeScript-Website/issues"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/sandbox.esm.js",
```
andrewbranch, line 16: Did you know that `module` is something bundlers made up out of thin air? 🙃 It’s ok to keep it, but we should use the new standard `exports` too:

```
"exports": {
  ".": {
    "types": "./dist/index.d.ts",
    "import": "./dist/sandbox.esm.js",
    "default": "./dist/index.js"
  }
}
```

This is assuming that the `index.d.ts` file works equally well for the ESM and CJS file, which is probably true if there’s no default export. If there are appreciable differences, there should be a separate `sandbox.esm.d.ts`. (By the way, why not just `sandbox.mjs` or `index.mjs` for that?)

andrewbranch (changes requested):

packages/sandbox/package.json
orta: This is likely to keep it consistent with the rest of the packages in this repo, it's safe to say that tsdx has not kept up with the times: https://github.com/jaredpalmer/tsdx/releases

andrewbranch:
I can reproduce the same build error that CI is getting:

```
(commonjs plugin) Error: Insufficient rollup version: "@rollup/plugin-commonjs" requires at least rollup@2.68.0 but found rollup@1.32.1.
```

packages/sandbox/package.json
oliverdunk: I have to admit, my experience publishing packages to NPM is fairly limited! I didn't know about that key so good to know. Left the file extensions to keep things consistent per @orta's comment.

oliverdunk:
> I can reproduce the same build error that CI is getting:
> 
> ```
> (commonjs plugin) Error: Insufficient rollup version: "@rollup/plugin-commonjs" requires at least rollup@2.68.0 but found rollup@1.32.1.
> ```

Super strange! This continued to work for me locally until I re-cloned, so I'm not sure what happened there. Sticking to the theme of consistency I've continued using tsdx but downgraded that package to the most recent compatible version.

I've also added the exports key, hopefully the tests will pass. This should be good for another look!

oliverdunk:
Pushed some fixes for conflicting types, and also updated `create-typescript-playground-plugin`. While that plugin could arguably use `@typescript/sandbox` from NPM now, I think that would be a good follow-up PR. It would be nice to wait until the package is live and there are existing issues that mean it may need additional work: https://github.com/microsoft/TypeScript-Website/issues/2441

andrewbranch:
A question for @orta or @oliverdunk; if we merge this and I want to verify that it didn’t break the site, what should I look for? I assume the Playground would be totally busted?

orta:
Yeah, the playground wouldn't build I imagine - all the references which are changed are type ones, so I'd expect CI would raise that. If the playground runs on staging, then we've definitely not broke web consumers

oliverdunk:
That sounds reasonable to me. @orta, I assume since you didn't mention it the releases.ts rename seems ok? It doesn't feel impossible that people were depending on that file location to load releases data, but at the same time that seems like a slightly odd thing to do (especially with a JSON equivalent existing) - and I don't know if the location and contents of a non-index JS file are something that should be expected to stay the same 🤔 

I can look at changing that back if we want to, I'm 50/50 on it. It just got fiddly with tsdx and rollup...

orta:
Yeah, I've always recommend folks use the CDN URLs for that data: https://typescript.azureedge.net/indexes/releases.json - I think the path for that file can generally be considered internal details

andrewbranch:
Lol

image

oliverdunk:
Is the budget getting low? 😜 You may have seen this but it looks like we can either publish once from the command line or add something to the package.json to explicitly make it public: https://stackoverflow.com/questions/41981686/getting-error-402-while-publishing-package-using-npm

I wonder if we could also just create a project manually in the account but not sure about that, and maybe you don't have easy access to the credentials...

andrewbranch:
Oh yeah, for personal packages I’ve always added the `publishConfig`. I could dig through the keyvaults for the token and run some manual stuff, but I’d rather not as that seems more error prone. Mind opening a follow-up PR?

orta:
add: 

```
  "publishConfig": {
    "access": "public"
  }
```

I tried to get the error message to note this in npm, but it didn't get much momentum: https://github.com/npm/feedback/discussions/583

oliverdunk:
PR is here: https://github.com/microsoft/TypeScript-Website/pull/2442

oliverdunk:
It's live (https://www.npmjs.com/package/@typescript/sandbox)! Thank you both so much for the help here. I really appreciate how quickly you were able to help me merge it.

andrewbranch:
Thanks @oliverdunk!

PR #2438, opened 2022-06-30
Add the JSDoc const assertion section to the website version

Retsam:
This section is referenced in the first paragraph but wasn't actually included in the website version of the release notes.


orta (approved):
Sounds good, and I checked - `@type {const}` is in the JS Doc reference also

jakebailey (approved):
Checked, and it's the same as the blog post, so LGTM.

PR #2447, opened 2022-07-05
Update Object Types.md

ardagungorsoci:


orta (changes requested):
Thanks for the PR, but this looks like an accident

PR #2449, opened 2022-07-10
48441 Adding default values to properties in reference.

s-dehaan:
Using the values specified here:
https://www.typescriptlang.org/docs/handbook/compiler-options.html

Original issue:
https://github.com/microsoft/TypeScript/issues/48441

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -6,12 +6,12 @@ import ts from "typescript";
 export interface CommandLineOption {
   name: string;
   type:
-    | "string"
-    | "number"
-    | "boolean"
-    | "object"
-    | "list"
-    | Map<string, number | string>;
+  | "string"
+  | "number"
+  | "boolean"
+  | "object"
+  | "list"
+  | Map<string, number | string>;
   defaultValueDescription?: string | number | boolean | ts.DiagnosticMessage;
   category?: ts.DiagnosticMessage;
   element: CommandLineOption;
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -200,53 +200,126 @@ export const defaultsForOptions = {
...
-            option.type === "list" ? option.element.type : option.type
-          ),
+          option.defaultValueDescription,
+          option.type === "list" ? option.element.type : option.type
+        ),
     ])
   ),
+  allowJs: "`false`",
   allowSyntheticDefaultImports: [
-    "`true` if [`module`](#module) is `system`, or [`esModuleInterop`](#esModuleInterop) and [`module`](#module) is not `es6`/`es2015` or `esnext`,",
```
jakebailey, line 209: I'm not super sure why all of these trailing comments have been removed; they do look correct on the reference page.

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -6,12 +6,12 @@ import ts from "typescript";
 export interface CommandLineOption {
   name: string;
   type:
-    | "string"
-    | "number"
-    | "boolean"
-    | "object"
-    | "list"
-    | Map<string, number | string>;
+  | "string"
```
jakebailey, line 9: Why the formatting changes everywhere?

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -200,53 +200,126 @@ export const defaultsForOptions = {
...
+  importHelpers: "`false`",
+  importsNotUsedAsValues: "`remove`",
+  include: ["`[]` if [`files`](#files) is specified", "`**` otherwise."],
   incremental: trueIf("composite"),
+  inlineSourceMap: "`false`",
+  inlineSources: "`false`",
+  isolatedModules: "`false`",
+  jsx: "`preserve`",
   jsxFactory: "React.createElement",
+  keyOfStringsOnly: "``false",
```
jakebailey, line 253: Typo here.

jakebailey (changes requested):
A lot of these don't seem accurate, or are conditional on other things. e.g. `baseUrl`'s default is not `./`; it's now optional. `esModuleInterop` depends on `module`. A lot of these other ones can be "unset" and mean different things, e.g `outDir` and `outFile` are typically unset, and setting them to `./` means a different thing semantically (or is invalid).

I do appreciate the effort, but there's quite a lot that needs to be fixed up here, and at some level I'm unsure if it's worth trying to list defaults for all of these.

PR #2450, opened 2022-07-12
Corrected punctuation. Replaced comma with period

gurry:
because it should have been two sentences instead of one.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -292,7 +292,7 @@ function move(animal: Fish | Bird) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Narrowing.md
```diff
@@ -292,7 +292,7 @@ function move(animal: Fish | Bird) {
 }
 ```
 
-To reiterate optional properties will exist in both sides for narrowing, for example a human could both swim and fly (with the right equipment) and thus should show up in both sides of the `in` check:
+To reiterate optional properties will exist in both sides for narrowing. For example a human could both swim and fly (with the right equipment) and thus should show up in both sides of the `in` check:
```
jakebailey:
```suggestion
To reiterate, optional properties will exist in both sides for narrowing. For example, a human could both swim and fly (with the right equipment) and thus should show up in both sides of the `in` check:
```

jakebailey (approved):

PR #2454, opened 2022-07-15
Don't suggest using npx tsc for one-off occasions

davetapley:
Doing so will install a deprecated package which issues warning:
"This is not the tsc command you are looking for"
https://www.npmjs.com/package/tsc


orta:
Yeah, I agree, otherwise we should recommend the full `npx` command including the reference to the module of typescript

jakebailey (commented):
Honestly, I think this entire section should be dropped. Nobody should be installing TypeScript globally. I removed the "global" install instructions from the main TypeScript readme a while ago.

jakebailey (approved):
I still think we should drop this entire section, but this PR is an obvious improvement so I'll just merge it.

jakebailey:
GH looks broken; will merge manually:

image

PR #2455, opened 2022-07-16
update the description about equality operator

awxiaoxian2020:
I don't think the original text is clear. Especially for non-native English speakers.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for the New Programmer.md
+++ b/packages/documentation/copy/en/get-started/TS for the New Programmer.md
@@ -29,7 +29,7 @@ There are many developers these days that use _only_ JavaScript to program their
 To summarize, we have a language that was designed for quick uses, and then grew to a full-fledged tool to write applications with millions of lines.
 Every language has its own _quirks_ — oddities and surprises, and JavaScript's humble beginning makes it have _many_ of these. Some examples:
 
-- JavaScript's equality operator (`==`) _coerces_ its arguments, leading to unexpected behavior:
+- JavaScript's equality operator (`==`) attempts to _coerce_ the values, leading to unexpected behavior:
 
   ```js
   if ("" == 0) {
--- a/packages/documentation/copy/en/get-started/TS for the New Programmer.md
+++ b/packages/documentation/copy/en/get-started/TS for the New Programmer.md
@@ -29,7 +29,7 @@ There are many developers these days that use _only_ JavaScript to program their
 To summarize, we have a language that was designed for quick uses, and then grew to a full-fledged tool to write applications with millions of lines.
 Every language has its own _quirks_ — oddities and surprises, and JavaScript's humble beginning makes it have _many_ of these. Some examples:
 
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/get-started/TS for the New Programmer.md
```diff
@@ -29,7 +29,7 @@ There are many developers these days that use _only_ JavaScript to program their
 To summarize, we have a language that was designed for quick uses, and then grew to a full-fledged tool to write applications with millions of lines.
 Every language has its own _quirks_ — oddities and surprises, and JavaScript's humble beginning makes it have _many_ of these. Some examples:
 
-- JavaScript's equality operator (`==`) _coerces_ its arguments, leading to unexpected behavior:
+- JavaScript's equality operator (`==`) attempts to _coerce_ the values, leading to unexpected behavior:
```
orta: Given that this is for new programmers, I think we're expecting them to know what coercing types is in both versions we could go simpler:

```suggestion
- JavaScript's equality operator (`==`) acts quite loose with types, leading to unexpected behavior:
```

awxiaoxian2020: @orta good!

awxiaoxian2020:
@andrewbranch Could you have a review? :)

jakebailey (changes requested):
I'm not sure I like either of the descriptions (original or modified); there is no "attempt", the coercion always does something. I also think it's valuable to introduce this term as it's pretty integral to the way JavaScript operates. It seems like a link to a reference that defines coercion would be better than being non-specific.

awxiaoxian2020:
@jakebailey my first suggestion is https://github.com/microsoft/TypeScript-Website/pull/2455/commits/0bc67a63913f90b33fc5bcc473781c66b7b3659e .

I'm just confused by "argument", it seems we use it only at functions.
if it's common in operators in English, I will close this Pull Request.

jakebailey:
I see. Probably it should say "coerces its operands", as `==` is an operator.

packages/documentation/copy/en/get-started/TS for the New Programmer.md
```diff
@@ -29,7 +29,7 @@ There are many developers these days that use _only_ JavaScript to program their
 To summarize, we have a language that was designed for quick uses, and then grew to a full-fledged tool to write applications with millions of lines.
 Every language has its own _quirks_ — oddities and surprises, and JavaScript's humble beginning makes it have _many_ of these. Some examples:
 
-- JavaScript's equality operator (`==`) _coerces_ its arguments, leading to unexpected behavior:
+- JavaScript's equality operator (`==`) acts quite loose with types, leading to unexpected behavior:
```
awxiaoxian2020:
```suggestion
- JavaScript's equality operator (`==`) _coerces_ its operands, leading to unexpected behavior:
```

jakebailey (approved):

andrewbranch:
cue the “isn’t it all just a lisp dialect” people?

PR #2458, opened 2022-07-21
Add lib files of es2022 to knownLibFilesForCompilerOptions

tocque:
fix knownLibFilesForCompilerOptions, which will break when use ScriptTarget.ESNext


jakebailey:
Note that #2802 is likely to supersede this, in that we're hoping to not have to mirror this list forever.

jakebailey:
Closing in favor of #2802.

PR #2469, opened 2022-08-02
Update TS for OOPers.md

ododa:


orta:
I can see a case for both the before and the after to be honest

ododa:
Isn't this guide teaching TypeScript?

orta:
Yep but TypeScript doesn't run in JavaScript engines - JavaScript does and the original wording is appropriate for that

ododa:
Alright. I'm still convinced "TypeScript" would be a better choice but I'll close the PR.

PR #2470, opened 2022-08-05
Fix: void type is not a subtype of undefined type

yo-goto:
The current docs says `void` type is a subtype of `undefined`.
> a subtype of `undefined` intended for use as a return type.

However, `undefined` type is assignable to `void` type, but not vice-versa.
```ts
const v: void = undefined; // OK
const u: undefined = v; // [Error]
//    ^: Type 'void' is not assignable to type 'undefined'.
```
And also, the archived specification says `void` type is a supertype of `null` type and `undefined` type.
>The only possible values for the Void type are null and undefined. The Void type is a subtype of the Any type and **a supertype of the Null and Undefined types**, but otherwise Void is unrelated to all other types.
>( https://github.com/microsoft/TypeScript/blob/main/doc/spec-ARCHIVED.md#325-the-void-type )

It seems a mistake.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
@@ -77,7 +77,7 @@ TypeScript has corresponding primitive types for the built-in types:
 | `unknown`      | the top type.                                               |
 | `never`        | the bottom type.                                            |
 | object literal | eg `{ property: Type }`                                     |
-| `void`         | a subtype of `undefined` intended for use as a return type. |
+| `void`         | a supertype of `undefined` intended for use as a return type. |
 | `T[]`          | mutable arrays, also written `Array<T>`                     |
 | `[T, T]`       | tuples, which are fixed-length but mutable                  |
 | `(t: T) => U`  | functions                                                   |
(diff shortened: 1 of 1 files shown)


yo-goto:
If the archived spec is correct even now, type hierarchy(subtype & supertype relationship) should be like this.
```mermaid
graph LR
  U[unknown]
  N[never]
  A[any]
  n[null]
  u[undefined]
  U --> A
  A --> void --> n & u --> N
```

But I believe the hierarchy below is true with `strictNullChecks` option in the latest version, cause `null` type is not assignable to `void` type. (Anyway, `void` type is a supertype of `undefined` type.)
```mermaid
graph LR
  U[unknown]
  N[never]
  A[any]
  u[undefined]
  n[null]
  U --> A --> n --> N
  A --> void --> u --> N
```

Handbook's assignability table also implies this fact.

assignability table*[TypeScript: Documentation - Type Compatibility](https://www.typescriptlang.org/docs/handbook/type-compatibility.html#any-unknown-object-void-undefined-null-and-never-assignability)*

>In TypeScript, there are two kinds of compatibility: subtype and assignment. **These differ only in that assignment extends subtype compatibility with rules to allow assignment to and from any, and to and from enum with corresponding numeric values**.

p.s. I hope the official docs has a type hierarchy diagram like these which clarifies assignability and sub-super relationship.

yo-goto:
#2760 fixed this issue.(deleted subtype info itself) So, closed.

PR #2471, opened 2022-08-05
chore: add missing key to button list in dev/twoslash page

enricopolanski:
List items should be indexed by a key in React.

While nothing major in this specific example, as the order of the items does not change dynamically, I still wanted to get rid of the warning to reduce noise while working on the website.

Due to the lack of a unique identifier other than the text content of the button itself I decided to use that for indexing, which is consistent with similar keys found in [#1](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/components/ShowExamples.tsx#L116), [#2](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/components/search/SearchResultsDisplay.tsx#L111) or [#3](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/templates/tsconfigReference.tsx#L160)

### DIFF
--- a/packages/typescriptlang-org/src/pages/dev/twoslash.tsx
+++ b/packages/typescriptlang-org/src/pages/dev/twoslash.tsx
@@ -298,6 +298,7 @@ const Index: React.FC<Props> = props => {
                 />
                 <div id="example-buttons">
                   {codeSamples.map(code => {
+                    
                     const setExample = e => {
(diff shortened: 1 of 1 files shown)


orta (approved):
Cool, looks reasonable to me- thanks

jakebailey (approved):

PR #2472, opened 2022-08-05
Rewrite gatsby-focus-wrapper id for playground handbook

sandersn:
It shouldn't have a top-level one since it's a nested page, even though it's generated like a top-level page.


sandersn:
I haven't found a way to test this change locally, so I'm going to try merging and checking whether it changes staging-typescript.

sandersn:
Nope! Doesn't appear to have changed, eg, https://www.staging-typescript.org/play?#handbook-4

orta:
FWIW is does work on staging for me:

image

Caching could be a pretty reasonable answer for why you still seeing the old behavior 👍🏻

sandersn:
You're right, when I check it on another machine the code is correct. Guess I need to learn how to clear whatever cache is in the way.

PR #2473, opened 2022-08-06
Get the site building on node 16

orta:
Removes all the ESM imports for JSON files because different versions of node require different syntax there ( import asserts in 16, none in 14)


andrewbranch:
14 will be EOL in half a year; do you have strong feelings against making 16 the minimum?

orta:
Yep, totally reasonable IMO

orta:
This PR should get merged

andrewbranch:
I’d rather just add the import assertions than switch to `readFile`

orta:
Meh, I'm not really motivated in moving it to another syntax which could also break instead of being fully backwards compat

orta:
[import assertions went back a TC39 stage](https://twitter.com/robpalmer2/status/1621234684066762755), dropping it is still the right call imo

andrewbranch:
Yep, _now_ I agree with you 😅

andrewbranch:
Going to retrigger CI

andrewbranch:
Any idea what the build error is?

orta:
Looks like the `bootstrap` command which grabs the release info is crashing, 

```
➤ YN0000: [@typescript/sandbox]: (node:3775) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
```

Couldn't say why, and it didn't repro locally, but maybe an old version of `axios` which doesn't run on Node 16? Either way I switched it out to node-fetch which I trust more

orta:
Just rebased: would recommend against doing these sorts of build steps:

```
    - run: |
          node test/verifyPackageVersions.js
          yarn docs-sync pull microsoft/TypeScript-Website-localizations#main 1
          yarn bootstrap
          yarn build
          yarn build-site
          cp -r packages/typescriptlang-org/public site
```

Makes it harder to see which step failed unless you echo titles etc

orta:
cool, looks like we have a repo for the canvas bail - will look later

andrewbranch:
Yeah, that was the one that was angry in #2680

orta:
Alright, this looks ready - now I know that the CI should turn out red, but that is (now) expected.

What's happening is that:

```
      - name: "Validates that TypeScript plugins work"
        run: |
          cd ..
          npm init typescript-playground-plugin playground-my-plugin
```

Runs against `v2` _not_ the current files:

```
$ rollup -c rollup.config.js

src/index.ts → dist...
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2792: Cannot find module 'lz-string'. Did you mean to set the 'moduleResolution' option to 'node', or to add aliases to the 'paths' option?
src/vendor/typescript-vfs.d.ts (57:140)

57 export declare const createDefaultMapFromCDN: (options: CompilerOptions, version: string, cache: boolean, ts: TS, lzstring?: typeof import("lz-string"), fetcher?: typeof fetch, storer?: typeof localStorage) => Promise<Map<string, string>>;
```

https://github.com/microsoft/TypeScript-Website/pull/2473/commits/eb65fda8e90724f341f2779f419a4c7d419201c3 fixes this

As it is the last item in the CI, then that validates that the rest of the system is 👍🏻

orta:
If we get this in and it's all fine, I'll take a look at bumping node to 18, and updating gatsby to 5.x - so that there's a good few more years runway

andrewbranch:
Thanks Orta!

orta:
Confirming that this all seems to look good on https://www.staging-typescript.org/

andrewbranch:
Merges to v2 go straight to prod these days 😉

orta:
nice!

PR #2475, opened 2022-08-07
Correct incorrect narrowing description

zukomgwili:
In the section on `control flow analysis`,  I think a mistake has been made in this sentence, "As a result, it was able to remove number from the type of padding (narrowing from string | number to string) for the rest of the function". It should read, "As a result, it was able to remove string from the type of padding (narrowing from string | number to number) for the rest of the function."


orta:
Hi, thanks for trying tp update the TS docs, but I think the original is correct. The line which matters:

```
  if (typeof padding === "number") {
```

Is about checking for "number" which means that the type number is removed from "string | number".

zukomgwili:
Thank you @orta for the clarification. I will proceed and and close this PR

PR #2477, opened 2022-08-09
Update yarn, removes husky

orta:
Yarn was an RC, it's now the current build

Husky was used by me to ensure I didn't break the build on pushes, but as I'm not shipping so much to this repo, then there's no need for the extra overhead


jakebailey:
Given #2523 is merged, I think the bulk of this has been merged already. The only thing this seems to have is the removal of the unused husky dep, which I think we can do separately in another cleanup.

PR #2478, opened 2022-08-09
Fix module resolution config option (+mention inline `type` on imports)

orta:
Re: https://github.com/microsoft/TypeScript/pull/44501#issuecomment-1207402007 /cc @weswigham


weswigham:
Thanks!

PR #2488, opened 2022-08-17
Convert site to 4.8

orta:


orta:
So, there's not much I think I'd change here - there's no new compiler options, most of the release notes are refinements to existing language primitives. I think it might just be fine as is.

DanielRosenwasser:
Thanks for the RC bump + other relevant changes! I'll still have to fill in with the 4.8 content and update the playground for the full release.

PR #2490, opened 2022-08-18
Wrong compiled example of Downleveling section

youssefnassim:
The downleveling section compares the original template string in the "Explicit Types" section with the compiled one at the "Erased Types" section in which the target used is es5.

And while es5 actually outputs:
Screen Shot 2022-08-18 at 03 41 34

here's what the comparison says
Screen Shot 2022-08-18 at 03 43 57


jakebailey:
This change was already merged in #2820.

PR #2494, opened 2022-08-23
Added an example to make the tip clearer

jonasraoni:
Here I just missed a final sample.


packages/documentation/copy/en/handbook-v2/More on Functions.md
```diff
@@ -426,6 +426,20 @@ Functions with fewer parameters (of the same types) can always take the place of
 
 > When writing a function type for a callback, _never_ write an optional parameter unless you intend to _call_ the function without passing that argument
 
+This way, a proper declaration and usage of a callback with optional parameters at the end could look like this:
```
jakebailey, line 429: I don't think that this new text adds to the section; the section is all about writing a function which accepts a callback and what can go wrong, not about the fact that you can omit parameters when passing in a calback function.

jakebailey (changes requested):

PR #2495, opened 2022-08-23
Added extra information about the immutability

jonasraoni:
I'm not sure yet if this is covered later in the documentation, if yes, please discard =]

I also thought about adding a reference to the https://github.com/tc39/proposal-record-tuple, but as it's not in an advanced stage, I didn't ¯\_(ツ)_/¯

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
@@ -586,6 +586,20 @@ a.push(102); // error
 a[0] = 101; // error
 ```
 
+It's important to remember that TypeScript doesn't transpile the immutability directives to the run-time:
+
+```ts
+let a: ReadonlyArray<number> = [0];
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/get-started/TS for Functional Programmers.md
```diff
@@ -586,6 +586,20 @@ a.push(102); // error
 a[0] = 101; // error
 ```
 
+It's important to remember that TypeScript doesn't transpile the immutability directives to the run-time:
```
jakebailey, line 589: I don't know that we should add this; it's technically true that one can use `freeze` to achieve this behavior, but it's exceedingly rare and does have some footguns which can make it undesirable.

Additionally, inserting this section here makes the last sentence of the section a little awkward, as it's intending to refer to the previous content, not `freeze`.

jakebailey (changes requested):

PR #2496, opened 2022-08-23
Added extra tips to the function overload section

jonasraoni:


packages/documentation/copy/en/handbook-v2/More on Functions.md
```diff
@@ -526,7 +527,21 @@ len([0]); // OK
 len(Math.random() > 0.5 ? "hello" : [0]);
 ```
 
-Because both overloads have the same argument count and same return type, we can instead write a non-overloaded version of the function:
+This way, the last statement could be turned into this:
+
+```ts twoslash
+declare function len(s: string): number;
```
jakebailey, line 533: I don't think this added text is needed; this is effectively the same example as in the previous section but written differently. The added `random` call doesn't have a semantic effect on the call sites.

jakebailey (changes requested):

PR #2498, opened 2022-08-24
docs: fix typo

jobayer977:
The adjective other appears to be modifying the singular noun syntax. Consider making a change.


orta (changes requested):
Thanks for trying to improve the docs, but I think the previous version is more accurate (as 'syntax
 is being used as a collective reference to the extra syntax added, and not a singular instance of syntax changes like 'as')

PR #2501, opened 2022-08-25
eliminate spurious semicolon differences in examples

ketang:
The presence or absence of semicolons doesn't matter, but a newbie may not realize that and get hung up on the difference.


orta (approved):
Agree, no need for them to be inconsistent - thanks

PR #2502, opened 2022-08-25
make interface/type code example clearer and valid

ketang:
replace getBear() with an actual object instantiation so someone can copy the code into the playground and have it be valid without editing


orta (changes requested):
Thanks for trying to improve the docs, but I think this detracts from the points in the table and is a bit worse off because the 'Bear' type in both examples now isn't connected at all to the `bear` value in the sample (so would not type check the value at all)

we used a function to skip the messing around of assigning the type in the examples because its implicit in the reading of 'getBear' for both cases

PR #2505, opened 2022-08-28
Complete sentence

karlosmartines:
Changed sentence so that is is complete


jakebailey:
I think the original wording is acceptable; the proposed one repeats the phrase "you need" and I feel that it reads worse to my eye.

PR #2508, opened 2022-08-29
Add info about "Generic parameter defaults"

jaworek:
While working with generic types, I noticed that information about "Generic parameter defaults" is missing in handbook. 
I found it in update notes of TypeScript 2.3.
I copied this section and added it to Generics chapter.

Here's the original section:  
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults


jaworek:
Hi @orta, do you think it could be merged soon? I think this is quite important section that is currently missing from handbook :D

jakebailey:
This was a dupe of #2263, which I merged yesterday.

PR #2509, opened 2022-08-29
Drop format on save from the settings.json

orta:
Should stop folks accidentally changing the formatting


jakebailey (approved):
I'm a big format on save fan, but I think currently it's best to disable this and we can revive proper formatting at a later point in time. I know I've been frustrated working in this codebase because it's not fully formatted.

PR #2368, opened 2022-05-03
Ordering of examples is confusing

dustinstender:
The order of the functions should respect the order of the statement on line 103.

PR #2515, opened 2022-08-31
Clearly differentiate between annotation example for "Parameters" & "Return Value"

bashyio:
Since the documentation is targeted at TypeScript beginners, I think examples shouldn't leave things to "assumption" and be as clear as possible.

The preceding statement states **"You can use interfaces to annotate parameters and return values to functions:"**
However, the following examples are listed in the **opposite order** and do not **specify whether you're annotating parameter or return value**.

The minor edit easily clarifies this.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for JS Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for JS Programmers.md
@@ -109,12 +109,12 @@ interface User {
   id: number;
 }
 // ---cut---
-function getAdminUser(): User {
-  //...
+function deleteUser(user: User) {
+  // ...annotate parameter
 }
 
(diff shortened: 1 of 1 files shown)


orta (approved):
I think this is a pretty good improvement, thanks

bashyio:
@orta This pull request has been awaiting workflow approval for 27 days (out of 30 before it gets deleted) since I'm a first time contributor to this repo. Can you please approve the workflow run from your end? Thank you

orta:
I'm afraid that I don't have write access to do that

jakebailey:
I'm opting to take #2368 instead, which just swaps the order to match the previous paragraph.

PR #2517, opened 2022-09-02
Fix overload signatures of an example in Do's and Don'ts

jb-:
* Before this change, the return types of overload signatures of `fn` were void
* But the implementation of overloads actually returns Moment object


orta (approved):
Makes sense, nice spot

jakebailey (approved):

PR #2524, opened 2022-09-06
Bump vfs and twoslash versions

andrewbranch:
To publish the change from #2522


.yarn/versions/9066d04f.yml
```diff
@@ -0,0 +1,15 @@
...
+  - glossary
+  - handbook-epub
+  - "@typescript/playground"
+  - playground-examples
+  - "@typescript/playground-handbook"
+  - "@typescript/playground-worker"
+  - "@typescript/sandbox"
+  - tsconfig-reference
+  - "@typescript/vfs"
+  - typescriptlang-org
```
andrewbranch: What is this? Am I supposed to commit it?

jakebailey: IIRC @orta said you didn't need to do any version stuff; maybe we shouldn't have added the plugin and instead let whatever Orta had run work? https://github.com/microsoft/TypeScript-Website/pull/2523#issuecomment-1238621849

jakebailey: Yeah, I think https://github.com/microsoft/TypeScript-Website/blob/v2/.github/workflows/v2-merged-staging.yml actually does this automatically. Maybe the plugin should be removed.

jakebailey: Hm, though I cannot figure out if it handles version bumps, of if that is even a thing that needs to be done.

jakebailey: But to answer the question what is this and should it be committed; this is what yarn 2+ uses to track changes to packages to know which packages need to bump and when. Normally, you would commit it, but I don't think that we actually use yarn to manage versions.

jakebailey: Oh, I need to read better. orta said that you just change the version and it does the rest. In which case, I guess you could do `yarn version patch --immediate` which skips this? (But, I didn't know that this was even the default behavior, I thought deferred was opt-in)

orta: My guess is that this is a similar thing to changesets - https://github.com/changesets/changesets - basically in the monorepo you have a declare updates command, it could be some packages have minor some major, then later you do some sort of deploy command which uses this metadata. 

IMO you drop it as you're not using it during deploy.

jakebailey: Yeah. That's basically it. But, what do we do in this repo? Just bump the version by hand? Use yarn? Something else? (I have no idea, but I don't think we use yarn deferred changes)

andrewbranch: I don’t understand the “just update the versions manually” flow. You’re even updating the lockfile by hand? All I was looking for was something to bump every mention of 1.3.4 to 1.3.5.

andrewbranch:
Deleted the yarn versions artifact; going to try merging.

orta:
Ah yeah, I have a script `test/verifyPackageVersions` which ensures all the versions are correctly synced (which runs on CI I think)

PR #2526, opened 2022-09-07
Allow specifying TS lib directory for VFS

andrewbranch:
This should let us undo the kludge here https://github.com/microsoft/TypeScript-Twoslash-Repro-Action/blob/master/dist/package.json#L2


orta (approved):
Love that hack

PR #2536, opened 2022-09-18
[Fix]: Word swapping in Truthiness narrowing

SujayPrabhu96:
- Fix grammatical errors in Truthiness narrowing


orta:
I think both the before and after have the same meaning grammatically

jakebailey (approved):
Yeah, it's basically the same, though I personally have a preference for the suggested way.

PR #2543, opened 2022-09-26
Link to accessibility notice

navya9singh:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1568069/

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/SiteFooter.scss
+++ b/packages/typescriptlang-org/src/components/layout/SiteFooter.scss
@@ -12,6 +12,7 @@
 
   a {
     color: white;
+    display: inline-block;
   }
 
   section {
@@ -60,6 +61,7 @@
     @media (max-width: $screen-sm) {
       li {
         margin-left: 10px;
+
         p {
           width: 100%;
         }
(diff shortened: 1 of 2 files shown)


RyanCavanaugh:
We need to not deploy this until the landing page https://www.microsoft.com/it-it/accessibility/declarations is ready

RyanCavanaugh (approved):

packages/typescriptlang-org/src/components/layout/SiteFooter.scss
```diff
@@ -124,14 +131,15 @@
       }
     }
 
-    article > ul {
+    article>ul {
```
DanielRosenwasser, line 134: What formatted these differently?

DanielRosenwasser:
Is there a way to preview this? If not, can you provide a screenshot of what this looks like in both mobile and desktop?

packages/typescriptlang-org/src/components/layout/SiteFooter.scss
navya9singh: Just adding something new and then saving it creates these changes. I tried to remove them but saving the file again after removing something creates it.

navya9singh:
@DanielRosenwasser  This is the desktop view 
[](url)
website1
This is the mobile view 
image

DanielRosenwasser (approved):

PR #2549, opened 2022-09-29
Update wording.

daniel-schroeder-dev:
The wording of the subject line in "Installing TypeScript" seemed awkward. This seemed more natural.


hexiaokang (approved):

jakebailey (approved):
This change seems fine to me, though I do think this entire page needs a rewrite; it mentions old VS releases which are no longer supported, and it also suggests installing typescript globally, which is a very bad idea.

PR #2551, opened 2022-09-29
Fix exports field in ESM Support for Node.md

hoseungme:
Dual ESM/CJS TypeScript packages should provide `.d.mts` or `.d.cts` file depending on `type` field in `package.json`.
So I fixed `exports` field example.


hoseungme:

jakebailey:
This was fixed in #2732.

PR #2553, opened 2022-09-30
Update *Downleveling* string concatenation example

daniel-schroeder-dev:
In [Erased Types](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#erased-types), the example that renders on the screen uses the `String.prototype.concat()` method instead of the concatenation operator as the [Downleveling](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#downleveling) example shows.

Here's the rendered example from *Erased Types*:

[Erased Types concatenation example](https://imgur.com/a/UjiLPmK)


jakebailey:
This same change was merged in #2820.

PR #2307, opened 2022-03-18
Fix documentation for the website

DamianArado:
Fixes: 
- https://github.com/microsoft/TypeScript-Website/issues/2075
- https://github.com/microsoft/TypeScript-Website/issues/2186
- https://github.com/microsoft/TypeScript-Website/issues/2229

PR #2555, opened 2022-09-30
Update Generics.md | Bees have six legs, not four.

bdugg9119:
I updated the last example given to use Lions and Elephants instead of Bees and Lions. I did this because the previous example sets Bees to have four legs. Bees are insects, and thus have six legs. A pedantic point, but I think it's helpful when examples are accurate to real-world scenarios.


jakebailey:
Theoretically, the "correct" fix is to just override the legs value in the `Bee` class; that seems to be the reason why it's declared as `number` in `Animal`. Would accept that as a separate, simpler PR. See also: https://github.com/microsoft/TypeScript-Website/pull/2307#discussion_r1198104981

PR #2560, opened 2022-10-07
Update `node12` reference to `node16`

markhaslam:


orta:
This file is auto-generated from the typescript compiler's info - so this change probably has to happen upstream (or will be pretty instantly overwritten)

jakebailey:
This appears to already be fixed.

PR #2563, opened 2022-10-09
improve readonly index signature   

Ahmed-Hakeem:
properties already defined in the same type will override read-only indices behavior

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Object Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Object Types.md
@@ -321,7 +321,7 @@ interface NumberOrStringDictionary {
 }
 ```
 
-Finally, you can make index signatures `readonly` in order to prevent assignment to their indices :
+Finally, you can make index signatures `readonly` in order to prevent assignment to their indices:
 
 ```ts twoslash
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Object Types.md
```diff
@@ -337,6 +337,22 @@ myArray[2] = "Mallory";
 
 You can't set `myArray[2]` because the index signature is `readonly`.
 
+> Note that this is valid for all properties except properties already defined in the same type
```
jakebailey, line 340: I don't think that this new text adds to the docs; the `readonly` modifier is on the index signature so I believe it's already pretty clear that it only applies to that signature, just like how the `readonly` modifier only applies to a single property.

jakebailey (changes requested):

packages/documentation/copy/en/handbook-v2/Object Types.md
Ahmed-Hakeem: I can see that as well :) ..... maybe it would be useful to be added as a side note not a special case just to make this case clear especially for beginners. if you agree adding it like that I can change it .

PR #2569, opened 2022-10-16
Improve TSConfig Reference -> Downlevel Iteration -> Effects on Array Spreads

ghost:
In https://www.typescriptlang.org/tsconfig#example-effects-on-array-spreads:

- Corrected 'spreaded' and 'concated' to real words.
- Used `console.log` instead of `... in ...` to make it more obvious what's happening.
- Moved the console output under the corresponding statements.
- Improved the explanation to make it clear that the missing elements are set to `undefined` when using the spread syntax.


ghost:

jakebailey (approved):

PR #2572, opened 2022-10-19
Rectified typo

aanand07:
Rectified a typo in https://www.typescriptlang.org/docs/handbook/enums.html#union-enums-and-enum-member-types


orta:
This isn't a typo, the original description is correct

PR #2580, opened 2022-10-24
Fixed typo in Modules.md

tdkehoe:
The example tried to import `hello.js` from `hello.ts`.

This page is missing sections on `package.json` and `tsconfig.json`. If you have these misconfigured then your modules won't work. Please add sections explaining what's needed. I'm not an expert on this, all I know is that `package.json` needs to have `"type": "module",` and `tsconfig.json` needs to have `"module": "ESNext",` and `"target": "ESNext"`. I'll start two sections at the bottom of the page, please complete these sections.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -72,7 +72,7 @@ export default function helloWorld() {
 }
 // @filename: index.ts
 // ---cut---
-import helloWorld from "./hello.js";
+import helloWorld from "./hello.ts";
 helloWorld();
 ```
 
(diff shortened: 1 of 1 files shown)


tdkehoe:

tdkehoe:

packages/documentation/copy/en/handbook-v2/Modules.md
```diff
@@ -72,7 +72,7 @@ export default function helloWorld() {
 }
 // @filename: index.ts
 // ---cut---
-import helloWorld from "./hello.js";
+import helloWorld from "./hello.ts";
```
jakebailey, line 75: We definitely do not want to be sticking `.ts` imports into our examples.

jakebailey (changes requested):
Also, I think it's likely that we are completely overhauling this doc soon.

packages/documentation/copy/en/handbook-v2/Modules.md
```diff
@@ -404,3 +404,57 @@ You can see all of the available options and what their emitted JavaScript code
 ## TypeScript namespaces
 
 TypeScript has its own module format called `namespaces` which pre-dates the ES Modules standard. This syntax has a lot of useful features for creating complex definition files, and still sees active use [in DefinitelyTyped](/dt). While not deprecated, the majority of the features in namespaces exist in ES Modules and we recommend you use that to align with JavaScript's direction. You can learn more about namespaces in [the namespaces reference page](/docs/handbook/namespaces.html).
+
+## package.json
+
+To use modules your `package.json` must include `"type": "module",`. For example,
```
jakebailey, line 410: This looks to be covered by https://www.typescriptlang.org/docs/handbook/esm-node.html, right? Maybe we just need links if we don't have them already?

andrewbranch:
Thank you for the effort, but this topic really can’t be wedged into this page as-is. It’s going to require a total rewrite, which is in progress at https://github.com/microsoft/TypeScript/issues/52593.

PR #2598, opened 2022-11-07
Fix a wrong word in handbook modules page

mnajj:
Fix a wrong word in the Non-modules section.
It should be "top-level `import`" instead of "top-level `await`"


mo662008:
Fix a typo in the Non-modules section.
It should be "top-level import" instead of "top-level await"

PR #2603, opened 2022-11-09
Update Indexed Access Types.md

Asaf-S:
The documentation wrongfully states that a constant cannot be used to index, but that's not true, as this is caused by a misfitting type that could be solved by using 'as const'.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
@@ -57,17 +57,17 @@ type Age2 = Person["age"];
 //   ^?
 ```
 
-You can only use types when indexing, so pay attention variable's type of the index.
+When indexing, you cannot use types that are wider than the keys of the object being indexed, so pay attention of the variable's type used to index.
 For example, when using `const` to make a variable reference, pay attention that its type is `string` and not `"age"` (this can be changed by using [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions):
 
 ```ts twoslash
 // @errors: 2538 2749
 type Person = { age: number; name: string; alive: boolean };
 // ---cut---
-const key = "age"; // Type is 'string'
+const key = "age"; // The type is 'string', which is too wide (e.g. "foo" is also a string, but not one of Person's keys), so the next line won't compile
 type Age1 = Person[key];
 
-const keyWithConstantType = "age" as const; // Type is "age"
+const keyWithConstantType = "age" as const; // Type is "age", which is a key in Person, so that will work
 type Age2 = Person[keyWithConstantType];
 ```
 
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
@@ -57,17 +57,21 @@ type Age2 = Person["age"];
 //   ^?
 ```
 
-You can only use types when indexing, meaning you can't use a `const` to make a variable reference:
+You can only use types when indexing, so pay attention variable's type of the index.
+For example, when using `const` to make a variable reference, pay attention that its type is `string` and not `"age"` (this can be changed by using [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions):
 
 ```ts twoslash
 // @errors: 2538 2749
 type Person = { age: number; name: string; alive: boolean };
 // ---cut---
-const key = "age";
-type Age = Person[key];
+const key = "age"; // Type is 'string'
+type Age1 = Person[key];
+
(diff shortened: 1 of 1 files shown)


orta (commented):
Nice improvement!

Asaf-S:
Is something wrong with the tests?

packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
```diff
@@ -57,17 +57,21 @@ type Age2 = Person["age"];
...
 // @errors: 2538 2749
 type Person = { age: number; name: string; alive: boolean };
 // ---cut---
-const key = "age";
-type Age = Person[key];
+const key = "age"; // Type is 'string'
+type Age1 = Person[key];
+
+const keyWithConstantType = "age" as const; // Type is "age"
+type Age2 = Person[keyWithConstantType];
```
jakebailey: I'm a little confused by this; aren't both invalid? Is that what you're trying to show here? Isn't the problem not to do with `const`, but that you can't index a type with something in value space, and therefore need to `typeof` something or use a type?

I would think that if anything, the old stuff could stay, but instead say something like:

> You can only use types when indexing. For example, you cannot index a type with a variable:

And maybe include a `typeof`?

jakebailey (changes requested):

packages/documentation/copy/en/handbook-v2/Type Manipulation/Indexed Access Types.md
Asaf-S:
```typescript
// Example 1
type Age1 = Person["age"]; // Success

// Example 2
const key2 = "age";
type Age2 = Person[key]; // Failure

// Example 3
const key3 = "age" as const;
type Age3 = Person[key]; // Success
```

In example 1 above, `"age"` can be used to index `Person` as it's one of `Person`'s keys.

In example 2 above, the reason why `key2` fails - is because the TypeScript type of `key` is `string`, and it's too wide for a key of the `Person` type.
For example: `bla` is a valid string but it's not a key of `Person`, and therefore `key2` is not fit to index `Person`.
For that reason, the existing documentation says users should use a constant as an index.

In example 3 above, the reason why `key3` succeeds - is that the TypeScript type of `key` is `"age"` (and not `string`), which is a key of the `Person` type.

Asaf-S: I've edited my PR to be more self-explaining

jakebailey: Sorry, I apparently missed the push.

What I'm saying is that it isn't behaving the way you're describing; your third example fails to compile: [Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAChBOBnA9gOygXigbygQwHMIAuKVAVwFsAjBAbjL0pKkWHgEtUCG8AbDgDcW1ZMj4Q86AL50AsAChFAemVQAogA8mYCVACMi0JCgBBIvsywEKVAG0ARIQgOAug1VQAyuQDGviERERRU1LR09ACZFXzQ2KABrCBBIqyciB3kFY2hzCFSsOCQ0OySQdyhPADE8Dj5yeAgQhU9wyl1oAGYYuOBE5M605wd8RChY1DYsnLMiQcKbErKKzx9-QMQgA)

rbuckton: The issue isn't `as const`, the issue is you need to use `typeof key2` and `typeof key3`: [Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAChBOBnA9gOygXigbygQwHMIAuKVAVwFsAjBAbjL0pKkWHgEtUCG8AbDgDcW1ZMj4Q86AL50AsAChFAemVQAogA8mYCVACMi0JCgBBIvsywEKVAG0ARIQgOAug1VQAyuQDGviERERRU1LR09ACZFXzQ2KABrCBBIqyciB3kFY2hzCFSsOCQ0OxzkADNE5Mj3KE8ff0DgpQVPcMpdaABmGLjgKpAutOcHfEQoWNQ2LJyzIiHCmxKyyqTB2vq-AKCgA)

While Computed Property Names in Object Type Literals and Interface Types use _values_, Indexed Access Types can only use _types_:

```ts
interface List<T> {
  [Symbol.iterator](): Iterator<T>;
// ^^^^^^^^^^^^^^^-- 'Symbol.iterator' here used as a value
}

type ListIterator = List<number>[typeof Symbol.iterator];
//                               ^^^^^^-- must use 'typeof' here to read its type
```

rbuckton:
I think the original statement is mostly accurate ("you can't use a `const` to make a variable reference"), what's actually missing is a qualifying statement and additional example related to `typeof`, i.e.

> However, you can use a type alias for a similar style of refactor:
>
> ...
>
> Or, you can perform a type query using `typeof` to get the type of the `const` variable:
>
> ```ts
> const key = "age";
> type Age = Person[typeof key];
> ```

PR #2604, opened 2022-11-09
Update Compiler Options.md

redreceipt:


orta:
I believe this file is code-generated, and so you'll need to make this change upstream in the microsoft/typescript

PR #2605, opened 2022-11-11
Fix ES Modules spec date

vapier:
The documentation quotes 2019 for when ES Modules were standardized, but that seems to refer to ES2020's dynamic import() function.  ES Modules were part of ES2015 (ES6) released in 2015.


orta (approved):
Yeah, I agree, I think I was thinking about Node's support of ES Modules with that date

jakebailey (approved):

PR #2606, opened 2022-11-11
Correct a wrong word in the handbook's modules page

mnajj:
In the non-modules section.
It should be "top-level (import)" not "top-level (await)"


hexiaokang (approved):

orta:
This PR is a dupe of https://github.com/microsoft/TypeScript-Website/pull/2374 (and more incorrect, 'await' is correct here)

PR #2613, opened 2022-11-27
Delete Templates.md

nrz77:


orta:
We'd break links if this happens, can you explain your reasoning?

PR #2617, opened 2022-11-29
add codeql

sandersn:


sandersn:
CodeQL is running correctly; there are 8 alerts that resulted.

PR #2619, opened 2022-11-29
Removing playwright

navya9singh:


andrewbranch:
Context: We’re removing the PDF and ePub handbook generation because the playwright dependency is constantly causing build errors both in CI and in local development and slowing down every build. There are ways we can improve the process and generate these assets without playwright (perhaps without native dependencies), but we were skeptical of the value they provide. HTML is a perfectly good format for reading, printing, and converting. Moreover, anyone paying attention to this repo will realize that we are in serious need of taking steps to simplify website maintenance. We currently have zero team members who are able to build the website on their local machine. Remedying that is an important first step to being able to do anything else.

orta:
This is reasonable , I'd recommend adding a redirect of some sort to the original URLS to the docs root -note - quite a lot of people used the PDFs and just giving a 404 is going to be a surprise

PR #2620, opened 2022-11-29
Fix release notes order

andrewbranch:
`TypeScript 4.8.md` was misspelled `Typescript 4.8.md` which changes its sort order


package.json
```diff
@@ -62,5 +62,8 @@
     ],
     "onlyPublishWithReleaseLabel": true
   },
-  "packageManager": "yarn@3.2.3"
+  "packageManager": "yarn@3.2.3",
+  "volta": {
+    "node": "14.21.1"
```
andrewbranch, line 67: Drive by fix: the site doesn’t build on Node 16 currently, which was my default. The TypeScript team is using volta to manage versions in the TypeScript repo.

orta: FWIW, https://github.com/microsoft/TypeScript-Website/pull/2473 moves the site to run on Node 16

PR #2622, opened 2022-11-29
Delete release notes overview page and run docs bootstrap

andrewbranch:
The overview page is incredibly outdated and begins with an embarrassing typo. Bye 👋


packages/playground-handbook/scripts/generateJSON.mjs
```diff
@@ -3,22 +3,21 @@
...
+const root = join(__dirname, "..");
+const files = readdirSync(root);
+const mds = files.filter(f => f.includes(".md"));
+const types = `export type PHandbookPage = ${mds.map(t => `"${t}"`).join(" | ")}`;
+writeFileSync(join(__dirname, "types.d.ts"), types);
 
 /** @type {import("./types").PHandbookPage[]} */
 // prettier-ignore
 const contents = [
-  "Overview.md",            // 0
```
andrewbranch, line 21: Accidental auto format in this file I guess. This is the only substantive change.

orta:
I'd recommend adding some kind of redirect for this, so that URLs don't break - maybe just to the docs root? https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/redirects/setupRedirects.ts

PR #2627, opened 2022-11-30
Update Decorators.md

flintforge:
fix the "override constructor" example in 'Class decorators' section


orta (changes requested):
I don't think we want this PR, as it's showing how using decorators in this way does not affect the type system, and your changes make it look like it does

flintforge:
Trying the given example, shows that `bug.reportingURL` holds the value given in the decorator (`"http://..."`)
So what's the point in the last line. Is it an implicit `return` ?
This doesn't seem to be informative about how the type system isn't affected.

orta:
It shows an error in the example on the site:

image

That reportingURL exists at runtime, but not in the type system is the point of that extra set of that code sample

jakebailey:
See above; this change doesn't look correct. I think the example is correct as-is.

PR #2411, opened 2022-06-05
Improve readability of type P

enheit:

PR #2632, opened 2022-12-06
Enclose type names with backticks in Keyof page

netotz:
# Page

- https://www.typescriptlang.org/docs/handbook/2/keyof-types.html

# Issue

The 2 types in this paragraph, `P` and `"x" | "y"`, don't look like code:

image

# Changes

- Enclose those 2 types in backticks so they look like code.


jakebailey:
Fixed in #2411.

PR #2636, opened 2022-12-08
Type-Checking Static Members of a Class

ajvincent:
This is something that was non-obvious to me at first glance.  I figured writing down a simple example for others would not hurt.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Classes.md
+++ b/packages/documentation/copy/en/handbook-v2/Classes.md
@@ -909,6 +909,38 @@ const MyHelperObject = {
 };
 ```
 
+### Type-Checking Static Members of a Class
+
+Under the hood, each class is an object (really a function) with a prototype property.  You can define a type for the class's static members, which will be properties of the class (not instances of it).
+
+```ts twoslash
+interface VehicleTypeAndMaker {
+  manufacturer: string;
+}
+
+interface VehicleTypeAndMakerStatic {
+  readonly wheelCount: number;
+}
+
+class Vehicle implements VehicleTypeAndMaker {
+  readonly manufacturer: string;
+
+  constructor(maker: string) {
+    this.manufacturer = maker;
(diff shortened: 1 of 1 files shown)


ajvincent:

ajvincent:
I'm thinking about moving this down to the next section, "Static blocks in classes", as an illustrative example which fits better there.

```typescript
interface VehicleType {
  manufacturer: string
}

interface VehicleTypeStatic {
  readonly wheelCount: number;
}

class Vehicle implements VehicleType 
{
  static {
    const temp: VehicleTypeStatic = Vehicle;
    void(temp);
  }

  readonly manufacturer: string;

  constructor(maker: string) {
    this.manufacturer = maker;
  }

  static readonly wheelCount = 4;
}
```

I'm wondering what people think of this alternative.

packages/documentation/copy/en/handbook-v2/Classes.md
```diff
@@ -909,6 +909,38 @@ const MyHelperObject = {
 };
 ```
 
+### Type-Checking Static Members of a Class
+
+Under the hood, each class is an object (really a function) with a prototype property.  You can define a type for the class's static members, which will be properties of the class (not instances of it).
```
jakebailey, line 914: I'm not totally sure that this section is worth including, at least not as-is.

Are you hoping to have an example that shows how to work around the lack of `class Vehicle extends static VehicleTypeStatic`? There's probably a simpler way to do that, if so; I'd think we'd want to be pretty targeted about showing the difference between the the "value" of the class and the "type" of a class instance. Not sure if this doc already includes some of that language, though.

jakebailey (changes requested):

ajvincent:
I may not be able to respond in a timely manner.  I have a new job which takes up a lot more of my time than I've been used to.  If anyone wants to close this pull request, I have no objection.

PR #2637, opened 2022-12-09
Fix module typo in ESM Support for Node.md

jersonmorrow:
While using:

```
{
    "compilerOptions": {
        "module": "nodenext",
    }
}
```
The compiler throws this warning:
```
Value is not accepted. Valid values: "CommonJS", "AMD", "System", "UMD", "ES6", "ES2015", "ES2020", "ESNext", "None", "ES2022", "Node16", "NodeNext".
```


orta:
TypeScript supports these values as lowercase, I assume its some other tooling giving you that message - but its probably worth a merge as it's not much different to readers.

jakebailey (approved):

PR #2649, opened 2022-12-18
Fixed module name extension in Modules.md

singhAmandeep007:
I believe it should be `"./animal.ts"` instead of `"./animal.js"` for clear understanding.


Gerrit0:
This is intentional -- using `.ts` module specifiers is not supported. TypeScript will resolve the `.js` extension to TS files.

PR #2651, opened 2022-12-21
Fix(1654907): Ensures role attribute has an appropriate value

navya9singh:
1. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1654907 
   Verified that `aria-allowed-role`: Ensures role attribute has an appropriate value for the element, does not throw any error.
2. This change also fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1654908 and `presentation-role-conflict` does not throw any error.

### DIFF
--- a/packages/typescriptlang-org/src/components/index/EditorExamples.tsx
+++ b/packages/typescriptlang-org/src/components/index/EditorExamples.tsx
@@ -36,34 +36,34 @@ export const EditorExamples = () => {
     return false;
   };
 
-   return (
+  return (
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):

andrewbranch:
This probably needs additional work to be really correct, but removing `role="presentation"` will make a big difference. If you’re interested in going further, you might want to read through https://www.w3.org/WAI/ARIA/apg/patterns/carousel/

PR #2656, opened 2022-12-25
Update More on Functions.md

gmach:
Fixed spelling typo removing unnecessary 'this' and 'meh' in text.


orta:
I think it is fine as-is IMO - folks will always try make text super dry on PRs, but I think having at least some minor bits of levity make these long documents more readable on the whole.

PR #2657, opened 2022-12-27
Update Template Literal Types.md

gmach:
typo can instead of can't


orta:
I believe this `can't` is correct here, because it comes from the compiler - not any .d.ts files shipped with TS

PR #2659, opened 2022-12-28
Fix typo in page "Object Types"

hotpxl:


Gerrit0:
I'm pretty sure this was intentional - if you remove the types, that code block is valid JacaScript too.

orta:
Confirmed, thanks for trying to improve the docs @hotpxl but this change isn't correct

hotpxl:
oh my bad. I guess I was more looking at the type annotation, not the default values

PR #2661, opened 2023-01-03
A note about the "in" operator

SiarheiBobryk:


orta (approved):
Looks reasonable to me

SiarheiBobryk:
@orta, the second approval is required (I'm a "first-time contributor"). Could you please ping someone and ask for a review?

SiarheiBobryk:
One more approval is required. Please review 🙂

jakebailey:
It's a bit unfortunate that this is then the first mention of "prototype chain" on the page, where the existing code which mentions it makes sure to say "don't worry about what that is".

Is it worth mentioning at all?

jakebailey (changes requested):
^

SiarheiBobryk:
> It's a bit unfortunate that this is then the first mention of "prototype chain" on the page, where the existing code which mentions it makes sure to say "don't worry about what that is".

> Is it worth mentioning at all?

I'd say so, but I don't insist.

jakebailey (approved):

PR #2662, opened 2023-01-05
Update the documentation on compilerOptions.module to include ES2022

Retsam:
Noticed the `module` documentation hasn't been updated since module ES2022 was added.

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/module.md
+++ b/packages/tsconfig-reference/copy/en/options/module.md
@@ -85,7 +85,7 @@ import { valueOfPi } from "./constants";
 export const twoPi = valueOfPi * 2;
 ```
 
-#### `ES2015`/`ES6`
+#### `ES2015`/`ES6`/`ES2020`/`ES2022`
 
 ```ts twoslash
 // @showEmit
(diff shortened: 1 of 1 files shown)


Retsam:
I also notice the next section down says:

> Available from 4.7+, the `node16` and `nodenext` modes integrate with Node's [native ECMAScript Module support](https://nodejs.org/api/esm.html). The emitted JavaScript **uses either `CommonJS` or `ES2020` output** depending on the file extension and the value of the `type` setting in the nearest `package.json`. 

I'm not sure if the ES2020 there should also be updated to ES2022 or not.

packages/tsconfig-reference/copy/en/options/module.md
```diff
@@ -96,7 +96,7 @@ import { valueOfPi } from "./constants";
 export const twoPi = valueOfPi * 2;
 ```
 
-If you are wondering about the difference between `ES2015` (aka `ES6`) and `ES2020`, `ES2020` adds support for dynamic `import`s, and `import.meta`.
+In addition to the base functionality of `ES2015`/`ES6`,`ES2020` adds support for [dynamic `import`s](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import), and [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) while `ES2022` further adds support for [top level `await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await).
```
jakebailey:
```suggestion
In addition to the base functionality of `ES2015`/`ES6`, `ES2020` adds support for [dynamic `import`s](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import), and [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) while `ES2022` further adds support for [top level `await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await).
```

jakebailey (approved):

missinglink:
> BigInt literals are not available when targeting lower than ES2020.ts (2737)

PR #2667, opened 2023-01-06
Fix titles in Discord community callout

hluup:
Current titles reference Stack Overflow, changing these to Discord.


orta (approved):
Nice spot

hluup:
Shouldn’t these be also translatable? Maybe this pull request should deal with that also?

jakebailey (approved):

PR #2678, opened 2023-01-15
Lowercase for the "t" letter

SiarheiBobryk:
Although I'm not a native English, I believe it's supposed to be lowercased. Please correct me if I'm wrong.


SiarheiBobryk:
Please review
cc @orta

jakebailey (changes requested):
I don't think there's hard and fast rule, however, I'll note that this form is used in many, many more files than just this one, e.g. in the "TS for OOPers" page, "TS for JS Programmers", and other pages. I don't think it's worth fixing up.

PR #2685, opened 2023-01-18
Support preview environments for PRs from forks via manual dispatch and magic label

andrewbranch:
I think this works around the limitations mentioned at https://github.com/Azure/static-web-apps/issues/1, but requires a maintainer to manually trigger a deployment or add a label to a PR. I think this is fine to start with, but we could make it more automatic in the future if we feel it’s safe and necessary.

I think I have to merge this in order to test it, since there are changes to CI.yml that have to run before the new workflows.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,14 +36,22 @@ jobs:
         env:
           CI: true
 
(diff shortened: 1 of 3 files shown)


orta:
Clever!

PR #2691, opened 2023-01-23
Fix newline removal in twoslash inlays

jakebailey:


orta:
I think if you want to do this, you'd need to at least make sure there is a corresponding preceding whitespace for the extra lines, or you'd probably get this:

```
console.log(a)
//                   ^? {
  a: string
}
```

jakebailey:
I'm a little confused how the above would be possible; doesn't this remove the newlines entirely so you have to end up with `^? {  a: string  }`?

orta:
Ah, I misread! Yeah, I could imagine this working? I'm no pro on regex

andrewbranch (approved):

andrewbranch:
Haha I always wondered why those showed up

jakebailey:
Now looking at this on the playground, and it doesn't look good at all. [Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gWzAGwsCAeAKgPigXigG8AoASAG0BpKASwDsoBrCEOAMygCUIBHAV1oAnCABNsOALoAuKFmqSA3CQC+ykiQboh7AIYBjaAGVEaABYMA5sXLs4cAPyz6-BACMIQ5WTe6hs0jIfXQAvJygAZ2AhK28VVQ1QSFhEFDQIEwRzKwIUpFR0DEzs+kscZQB6CqgagD0HEiA)

These probably should have been replaced with spaces. I didn't realize that it was being replaced with the empty string.

PR #2694, opened 2023-01-25
Return a promise from ATA

melbourne2991:
This does a couple of things:

- More ergonomic
- Know when a specific type request has completed. (Current finish event is insufficient, doesn't tell you which request is completed and is never called if the provided source code has no imports).


orta:
Seems reasonable to me, I'd recommend also bumping the package json version so when the PR is merged it will also release 👍🏻

melbourne2991:

melbourne2991:
> Seems reasonable to me, I'd recommend also bumping the package json version so when the PR is merged it will also release 👍🏻

Done 👍

jakebailey (approved):

jakebailey:
With bump versions with #2904 to minimize the waiting within the current infrastructure.

jakebailey (approved):

PR #2700, opened 2023-01-30
Docs: Add correct code examples

### LINKED ISSUE #30976: Docs are wrong: export = and import = require(), says latter is required incorrectly
The documentation regarding `export = and import = require()` at https://www.typescriptlang.org/docs/handbook/modules.html is wrong. It says:

```
When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.
```

Whereas I've been able to `import from` and `import as` for modules that export using `export =`.

hiteshvijayhv:
fixes https://github.com/microsoft/TypeScript/issues/30976

### DIFF
--- a/packages/documentation/copy/en/reference/Modules.md
+++ b/packages/documentation/copy/en/reference/Modules.md
@@ -261,7 +261,7 @@ TypeScript supports `export =` to model the traditional CommonJS and AMD workflo
 The `export =` syntax specifies a single object that is exported from the module.
 This can be a class, interface, namespace, function, or enum.
 
-When exporting a module using `export =`, TypeScript-specific `import module = require("module")` must be used to import the module.
+The modules exported by using `export =` can be imported by either of the following ways such as `import module = require(“module”)` or `import module from “./module”` and `import * as module from “./module”`.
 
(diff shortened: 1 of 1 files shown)


jakebailey (changes requested):
Probably not going to try and fix this; there are big module doc updates pending which I believe will cover this.

packages/documentation/copy/en/reference/Modules.md
```diff
@@ -261,7 +261,7 @@ TypeScript supports `export =` to model the traditional CommonJS and AMD workflo
 The `export =` syntax specifies a single object that is exported from the module.
 This can be a class, interface, namespace, function, or enum.
 
-When exporting a module using `export =`, TypeScript-specific `import module = require("module")` must be used to import the module.
+The modules exported by using `export =` can be imported by either of the following ways such as `import module = require(“module”)` or `import module from “./module”` and `import * as module from “./module”`.
```
andrewbranch: Neither the previous content or the suggested content is correct/complete. This will be covered in more detail via https://github.com/microsoft/TypeScript/issues/52593. Let’s just remove this line for now.

```suggestion
```

PR #2707, opened 2023-02-10
Update baseUrl.md to reflect that typescript does not rewrite paths

AWare:
A lot of the writing around `baseUrl` online totally ignores that typescript does not rewrite the paths.

While it is not uncommon to having tooling set up to do this, or otherwise not use tsc. It does mean that running the output of tsc in node will fail when this setting is used. 

Furthermore, this is explicitly in line with the projects goals as stated in https://github.com/microsoft/TypeScript/issues/26722

As this is counterintuitive to a developer without a good understanding of the philosophy of Typescript, and a fairly serious pitfall; this deserves documenting.


jakebailey (commented):
The prose above this line regarding shortening of imports has been removed; as such I don't know that we actually need to mention that we don't rewrite paths here. A better place would be the docs for `paths`, if it doesn't already mention it (which I think it does?)

PR #2712, opened 2023-02-15
Adds missing dependencies

arcanis:
- `lz-string` is directly [referenced in types](https://github.com/microsoft/TypeScript-Website/blob/4a670b334df7be35c480f640009ce698a7bab02b/packages/ts-twoslasher/src/index.ts#L8). While the package includes [some types](https://cdn.jsdelivr.net/npm/lz-string@1.4.4/typings/lz-string.d.ts), they aren't picked up by TS since they are in a `typings` folder (and they don't export anything anyway). Without adding `@types/lz-string` to the dependencies (or `skipLibCheck`), TS will report an error.

- similarly, `typescript` is referenced [here](https://github.com/microsoft/TypeScript-Website/blob/4a670b334df7be35c480f640009ce698a7bab02b/packages/ts-twoslasher/src/index.ts#L9) but isn't listed as a dependency. Since the intent is to use whatever version the consumer uses, it should be a peer dependency.

Fixes the two following errors:

```
../.yarn/berry/cache/@typescript-twoslash-npm-3.1.0-878bc67728-9.zip/node_modules/@typescript/twoslash/dist/index.d.ts:1:33 - error TS7016: Could not find a declaration file for module 'lz-string'. '/Users/mael.nison/.yarn/berry/cache/lz-string-npm-1.4.4-59a2091d3f-9.zip/node_modules/lz-string/libs/lz-string.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/lz-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'lz-string';`

1 declare type LZ = typeof import("lz-string");
                                  ~~~~~~~~~~~

../.yarn/berry/cache/@typescript-twoslash-npm-3.1.0-878bc67728-9.zip/node_modules/@typescript/twoslash/dist/index.d.ts:8:79 - error TS2694: Namespace 'ts' has no exported member 'Map'.

8         type: "list" | "boolean" | "number" | "string" | import("typescript").Map<any>;
```

### DIFF
--- a/packages/ts-twoslasher/package.json
+++ b/packages/ts-twoslasher/package.json
@@ -51,7 +51,11 @@
   },
   "dependencies": {
     "@typescript/vfs": "1.4.0",
+    "@types/lz-string": "^1.3.0",
     "debug": "^4.1.1",
     "lz-string": "^1.4.4"
+  },
+  "peerDependencies": {
+    "typescript": "*"
   }
 }
--- a/packages/ts-twoslasher/package.json
+++ b/packages/ts-twoslasher/package.json
@@ -50,8 +50,8 @@
     "typescript": false
   },
   "dependencies": {
-    "@typescript/vfs": "1.4.0",
     "@types/lz-string": "^1.3.0",
+    "@typescript/vfs": "1.4.0",
     "debug": "^4.1.1",
     "lz-string": "^1.4.4"
   },
--- a/packages/ts-twoslasher/package.json
+++ b/packages/ts-twoslasher/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@typescript/twoslash",
-  "version": "3.2.1",
+  "version": "3.2.2",
   "license": "MIT",
   "author": "TypeScript team",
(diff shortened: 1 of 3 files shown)


orta (approved):
Yeah, this makes sense - I don't have write access, so can't merge, but I'd recommend also bumping the patch number so that on merge it will make a deploy

orta:
( also: I am always a tad suspicious of peer dependencies in a post npm 7 world, but I figure you know better than me in this domain )

arcanis:
Honestly I don't use npm much, so I can't be absolutely certain there ... 🥲

I *think* it should work though. Even with their auto-installs, it's only supposed to kick in if the consumers didn't provide a version themselves, and it would only error if the version they provide is lower than 3.0. At least I think...

jakebailey (approved):

Not sure what the deal is with CI; things work locally so I'll rerun this here again...

jakebailey:
This seems good and doesn't require bumping any packages except this `ts-twoslasher`; I'm going to merge it to see if the package publishing stuff is working alright.

PR #2714, opened 2023-02-15
Explain use of `never[]` as most generic function argument

tvogel:
For newbies like me that read the manual start-to-end to get an introduction into the concepts, the appearance of `never[]` here is surprising, inexplicable and confusing. So, I was very happy about the thorough explanation linked in this commit.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
@@ -204,6 +204,8 @@ type Str = GetReturnType<(x: string) => string>;
 type Bools = GetReturnType<(a: boolean, b: boolean) => boolean[]>;
 //   ^?
 ```
+If you are curious about the use of `never[]` as the most generic type for function arguments, there is an excellent explanation in this [reddit](https://www.reddit.com/r/typescript/comments/muyl55/comment/gv9ndij/?utm_source=reddit&utm_medium=web2x&context=3).
+
 
 When inferring from a type with multiple call signatures (such as the type of an overloaded function), inferences are made from the _last_ signature (which, presumably, is the most permissive catch-all case). It is not possible to perform overload resolution based on a list of argument types.
 


tvogel:
Ping - anybody out there?

tvogel:
@andrewbranch Is this of interest?

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
```diff
@@ -204,6 +204,8 @@ type Str = GetReturnType<(x: string) => string>;
 type Bools = GetReturnType<(a: boolean, b: boolean) => boolean[]>;
 //   ^?
 ```
+If you are curious about the use of `never[]` as the most generic type for function arguments, there is an excellent explanation in this [reddit](https://www.reddit.com/r/typescript/comments/muyl55/comment/gv9ndij/?utm_source=reddit&utm_medium=web2x&context=3).
```
jakebailey, line 207: I think that we probably need an explainer about this, but I don't think that this wording or a link to reddit is appropriate. Note that TS 5.0 greatly refined the type relationship between the different function "top types" such that this answer may not longer be accurate. Though the end result of "you probably want `never[]`" may still apply.

Related PR: https://github.com/microsoft/TypeScript/pull/52282

jakebailey (changes requested):

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
andrewbranch: Honestly, I think the example should replace `never[]` with `any[]`. We try to avoid having to include these “by the way did you know about this very confusing thing???” asides in the handbook specifically. An explainer might be better suited in the reference pages of the docs, but I don’t know if there’s any existing content where it would make sense to slot in.

andrewbranch (changes requested):

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
jakebailey: I did some sleuthing to figure out the origin of this; it seems like this has been there since the beginning of the new handbook in https://github.com/microsoft/TypeScript-New-Handbook/commit/1a21a7fd1215c79425df3a26e9200dd4009f025b; @DanielRosenwasser do you have an opinion on this one?

rbuckton: I think that `any[]` should be sufficient for the purposes of this example, especially since we use `any[]` in the definition of `Flatten<T>` further up on the same page.

PR #2715, opened 2023-02-17
Use .ts file extensions instead of .js

westonruter:
There appears to be some typos where imports are referring to .js files whereas the `@filename` annotations indicate they are .ts files. This corrects that mismatch.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -72,7 +72,7 @@ export default function helloWorld() {
 }
 // @filename: index.ts
 // ---cut---
-import helloWorld from "./hello.js";
+import helloWorld from "./hello.ts";
 helloWorld();
 ```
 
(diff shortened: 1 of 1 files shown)


orta (changes requested):
Deno is a runtime which supports this syntax, but realistically, we assume nodejs which _does not_ support this syntax. We should not make this change, as it's probably going to be very confusing to the majority of our readers

westonruter:
@orta I'm clearly a novice here, but could you explain further? Given these lines:

https://github.com/microsoft/TypeScript-Website/blob/4a670b334df7be35c480f640009ce698a7bab02b/packages/documentation/copy/en/handbook-v2/Modules.md?plain=1#L68-L77

Do the filenames get resolved _after_ `tsc` has processed the `.ts` files to generate the `.js` ones? Is that why even though `index.ts` references `hello.js`, this works because `tsc` would have compiled it to that new target to import? I'm confused as to how `tsc` would import from `hello.js` during parsing since it would be working with `hello.ts`.

orta:
You're welcome to read up on the different module resolution strategies TypeScript has built in via the reference docs on the you edited  [and here](https://www.typescriptlang.org/docs/handbook/module-resolution.html) - but the code samples on the side default to node-style resolution where you give the .js because javascript that is the only lanaguge natively supported in nodejs. A type would crash the engine..

To give people code samples which wouldn't throw in that environment isn't something we'd want to do.

PR #2717, opened 2023-02-18
Update to Node 18 & gatsby 5

orta:
Needs a 2nd pass because it looks like the build command has issues which the local dev server does not, but this updates everything!

image


orta:
OK, that's building fine locally now

orta:
( this PR is worth some attention )

jakebailey:
Hoping we can make some progress on this; there seem to be a bunch of unrelated content changes. Do you want me to push to this branch to get it de-conflicted? Is there any other work left to be done here?

orta:
It's been a while, but I think most of those changes are either running bootstrap or updating libs to versions which support node 18 - sure give it a shot, I think it was done

jakebailey:
Well, I tried to update this PR... it passes locally but fails in CI.

jakebailey:
Figured it out, `gatsby-plugin-typegen` doesn't generate code when `CI=true`: https://github.com/cometkim/gatsby-plugin-typegen/blob/b452c4274e35d3e0eebcad930696eb7131d8be8b/plugin/src/utils.ts#L14

The "fix" is to commit the file. If we ever modify it, it'll be modified on disk and should be committed anyway. gitignored but required generated files are such a pain 🙁 

I also modified CI to make it complain, replacing #2901 with the same simple diff we do in TS's CI (and I do in my own repos for generated code).

.github/workflows/CI.yml
```diff
@@ -31,25 +31,43 @@ jobs:
...
-          path: site
-
       # danger for PR builds
       - if: github.event_name == 'pull_request' && github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
         run: "yarn danger ci"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - run: |
+          if ! git diff --staged --exit-code --quiet; then
```
jakebailey, line 54:
```suggestion
          git add .
          if ! git diff --staged --exit-code --quiet; then
```

orta:
eyyy, nice one

navya9singh (approved):

PR #2719, opened 2023-02-20
remove unnecessary underscores

Kalmaegi:


orta (changes requested):
Personally, I think they're pretty useful as a way emphasis device

PR #2720, opened 2023-02-21
Update TS for JS Programmers.md

Shoyur:
typeof string to show the obj, else to show [obj] as an array.


orta:
Look at what the function name is `wrapInArray` - the original code is correct

PR #2723, opened 2023-02-24
Update Variable Declarations.md

Amarjeetdev012:


orta:
This is a HTML comment and would be invisible to readers?

Amarjeetdev012:
Yes, I have updated the removed comments to be visible to the reader

orta:
I don't really think this PR adds enough to be accepted, as it is a summary of the next paragraph which goes into more detail about what (and why) this behavior happens

Amarjeetdev012:
Thank you for your feedback. I understand your concerns about the PR not adding enough value to be accepted. It's helpful to have a more detailed explanation of the behavior in the next paragraph, but I believe the summary still provides a useful overview for readers who may not have time to read the entire paragraph. However, I'm open to suggestions on how we can improve the PR further. Let me know if you have any specific ideas or recommendations.

orta:
I don't really think the docs need sentence summaries before the _nuanced_ explanation for something which would be easily considered confusing, if you don't have the time to read 4 sentences then you're not probably going to be learning how this weird behavior works

Amarjeetdev012:
I understand your point of view, but I believe that providing a brief sentence summary before the detailed explanation can be helpful for some readers who want to quickly get a sense of what the topic is about before diving into the details. It can also help readers who may be unfamiliar with the terminology or concepts to understand the context better. However, I agree that a nuanced explanation is crucial for a complete understanding of the topic, and I will make sure to provide clear and concise explanations that are easy to follow.

PR #2729, opened 2023-02-28
Clarification in the order of type annotation examples

import-that:
This change reorders the examples of parameter and return type annotation such that the example code provided matches the description (first listing parameter type annotation, followed by return type annotation).

I believe this change, while minor, serves to clarify the purpose of the documentation by presenting code examples of type annotation in the same order that they are presented in the English description of the features, making it more obvious which lines of code  demonstrate which feature.


jakebailey:
Duplicate of #2368.

PR #2731, opened 2023-02-28
fix a typo in `TS for JS Programmers.md `

elijah0kello:


orta:
I think the original is fine FWIW, they both are saying the same thing with a different target

adeacetis:
> I think the original is fine FWIW, they both are saying the same thing with a different target

FWIW the correction changes the tone from informal to casual (direct)

jakebailey:
The new code does not read as grammatically correct to my eye.

PR #2732, opened 2023-03-01
Fix export map example and explain

andrewbranch:


andrewbranch:
@DanielRosenwasser can you take a look? We should maybe update the blog post too.

jakebailey (approved):

RyanCavanaugh (approved):

andrewbranch:
Huh, I was going to copy these changes back to the release notes, but the export map example there is different and already correct

weswigham (commented):
:+1:

PR #2736, opened 2023-03-01
Try catch gitInit

jakebailey:
The builder's git identity appears to break tests for this; I don't think this step is critical and so try/catching the problem away seems fair. If making a git repo with a single commit breaks on user machines, they can figure it out.

### DIFF
--- a/packages/create-typescript-playground-plugin/index.js
+++ b/packages/create-typescript-playground-plugin/index.js
@@ -35,10 +35,14 @@ const install = () => {
 }
 
 const gitInit = () => {
-  exec("git --version", { stdio: "inherit" })
-  exec("git init", { stdio: "inherit" })
-  exec("git add .", { stdio: "inherit" })
-  exec('git commit -am "Init"', { stdio: "inherit" })
+  try {
+    exec("git --version", { stdio: "inherit" })
+    exec("git init", { stdio: "inherit" })
+    exec("git add .", { stdio: "inherit" })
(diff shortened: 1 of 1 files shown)


packages/create-typescript-playground-plugin/index.js
```diff
@@ -35,10 +35,14 @@ const install = () => {
...
-  exec("git init", { stdio: "inherit" })
-  exec("git add .", { stdio: "inherit" })
-  exec('git commit -am "Init"', { stdio: "inherit" })
+  try {
+    exec("git --version", { stdio: "inherit" })
+    exec("git init", { stdio: "inherit" })
+    exec("git add .", { stdio: "inherit" })
+    exec('git commit -am "Init"', { stdio: "inherit" })
+  } catch (e) {
+    console.error(e);
```
jakebailey:
```suggestion
    console.error(e)
```

sigh

andrewbranch (approved):

jakebailey:
Wait, what? why did this still fail?

jakebailey:
... was it running out of `node_modules`?

andrewbranch:
I assume something like that

jakebailey:
??? then what are we testing?

orta:
IMO, I think it's pretty reasonable to remove this from CI - it's code which rarely changes (it's testing that the playground plugin template npc command works) but its also not critical and someone can just make an issue if it bails

PR #2740, opened 2023-03-06
Update Classes.md

krychaxp:
typo fix


justingrant:
The typo is intentional here to highlight that the error is not expected to be flagged by TS.

PR #2741, opened 2023-03-08
Update Type Compatibility.md

pisv:


jakebailey (approved):
This took a lot of thinking and asking others, but I believe that the new text is correct. I think it'd be even better to have an example of code that works and code that doesn't to more clearly explain this.

PR #2744, opened 2023-03-11
Make playground more accessible in dark mode

beerose:
I was working with some of the playground's extensions, like `Symbols`, `AST viewer` and it was extremely hard to read the information in the dark mode. The accessibility scores are quite low, too. 

I played a bit with different colors, and found some that make the playground more accessible. **The changes are only for the dark mode. The light mode styles stay the same.**

Some screenshots with comparisons:

**BEFORE**

CleanShot 2023-03-11 at 21 54 16@2x

**AFTER**

CleanShot 2023-03-11 at 21 34 31@2x

---

**Node name BEFORE**

CleanShot 2023-03-11 at 21 29 24@2x

**Node name AFTER**

CleanShot 2023-03-11 at 21 28 13@2x

---

**Node string BEFORE**

CleanShot 2023-03-11 at 21 33 53@2x

**Node string AFTER**

CleanShot 2023-03-11 at 21 33 40@2x

---

**Node number BEFORE**

CleanShot 2023-03-11 at 21 04 28@2x

**Node number AFTER**

CleanShot 2023-03-11 at 21 28 58@2x

### DIFF
--- a/packages/typescriptlang-org/src/templates/play.scss
+++ b/packages/typescriptlang-org/src/templates/play.scss
@@ -796,18 +796,30 @@ main #editor-toolbar {
 
       span.ast-node-number {
         color: blue;
+        .dark-theme & {
+          color: lightskyblue;
+        }
       }
 
       span.ast-node-string {
         color: green;
+        .dark-theme & {
+          color: greenyellow;
+        }
       }
 
       span.ast-node-boolean {
         color: red;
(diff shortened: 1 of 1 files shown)


beerose:

orta (approved):
Yeah, looks good to me

jakebailey (approved):

PR #2745, opened 2023-03-12
Update DOM Manipulation.md

ptu14:
Since the `lib` folder has been deleted in microsoft/TypeScript#52226 , this link does not work. Now the correct link to the source is this one.

I did not add the information that this file is generated by [TypeScript-DOM-lib-generator](https://github.com/microsoft/TypeScript-DOM-lib-generator) and this file becomes `lib.dom.d.ts`, because I felt that the most important thing for the user is to have a quick look at the types, and not necessarily to analyze in depth the library .


orta (approved):
Thanks

jakebailey (approved):

PR #2746, opened 2023-03-13
Make tsconfig-reference display values Title Case

mark-wiemer:
Some values were `camelCase`, this makes all titles `Title Case`.

Found using `grep -r "display: \"[a-z]" packages/tsconfig-reference/copy/en/options`

Other minor changes auto-generated and appear to fix up whitespace

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/clean.md
+++ b/packages/tsconfig-reference/copy/en/options/clean.md
@@ -1,5 +1,5 @@
 ---
-display: "clean"
+display: "Clean"
 oneline: "Delete the outputs of all projects."
 ---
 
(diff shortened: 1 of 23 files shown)


mark-wiemer:

orta:
I don't think we need this, the format is made like this for both SEO, and so you can find on the page for the exact field you are looking for

mark-wiemer:
What is the criteria for a title to be `camelCase` vs `Title Case`? It feels random

orta (approved):
Oh yeah, just went and looked - we have the real name value next to the title case, I bet this was probably just new flags getting added or non-complete automation - this should get merged 👍🏻

jakebailey (approved):
Checked the other files `tsconfig-reference` and this looks to correct an inconsistency.

PR #2748, opened 2023-03-14
Update link to library components under ts-config reference > `lib`

aaronhipple:
On https://www.typescriptlang.org/tsconfig#lib the link to "TypeScript source code" for available typings files was broken.

This PR updates the link to point to `src/lib` instead, which appears to be the new path for those typings.


orta (approved):
Thanks

jakebailey:
I merged #2779 which turned out to be the same change.

PR #2752, opened 2023-03-17
Disable legacy decorators in 5.0+

jakebailey:


packages/sandbox/src/compilerOptions.ts
```diff
@@ -50,11 +50,18 @@ export function getDefaultSandboxCompilerOptions(config: SandboxConfig, monaco:
     emitDecoratorMetadata: true,
     moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs,
 
-    target: monaco.languages.typescript.ScriptTarget.ES2017,
+    target: monaco.languages.typescript.ScriptTarget.ESNext,
```
jakebailey: I can split this into another PR, I just know we get surprised by this all the time.

DanielRosenwasser: It's probably time.

packages/sandbox/src/compilerOptions.ts
```diff
@@ -7,7 +7,12 @@ type Monaco = typeof import("monaco-editor")
  * These are the defaults, but they also act as the list of all compiler options
  * which are parsed in the query params.
  */
-export function getDefaultSandboxCompilerOptions(config: SandboxConfig, monaco: Monaco) {
+export function getDefaultSandboxCompilerOptions(
+  config: SandboxConfig,
+  monaco: Monaco,
+  ts: typeof import("typescript")
```
DanielRosenwasser: Instead of casting, just type this as `{ versionMajorMinor: string }` for now.

packages/sandbox/src/compilerOptions.ts
jakebailey: Er. Wait. I meant lib=esnext... This is the target. Maybe this is not a good idea?

jakebailey: Target implies lib, right?

DanielRosenwasser: It does - if you want, you can switch to `lib: esnext`, I'm not sure if that excludes the DOM.

jakebailey: Alrighty.

I also realized that this probably will make decorators not work in the browser, because the browsers don't have native support yet. So, I think I'll just revert this for now?

Touching `lib` is difficult here; I tried to do a similar thing for our tests and I think CompilerOptions is a little funky, and I'd need to specify different names than in tsconfig.

packages/sandbox/src/compilerOptions.ts
```diff
@@ -50,11 +55,16 @@ export function getDefaultSandboxCompilerOptions(config: SandboxConfig, monaco:
     emitDecoratorMetadata: true,
     moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs,
 
-    target: monaco.languages.typescript.ScriptTarget.ES2017,
+    target: monaco.languages.typescript.ScriptTarget.ESNext,
```
jakebailey:
```suggestion
    target: monaco.languages.typescript.ScriptTarget.ES2017,
```

andrewbranch (approved):

hecham1964:
US driver's license

PR #2756, opened 2023-03-22
Update Template Literal Types.md

a90120411:
Fix word error


orta:
`Uppercase` et al cannot be found in the `.d.ts` files, the original docs are correct here

a90120411:
Thank you for your explanation.

PR #2757, opened 2023-03-23
Fix typo in union enums (overlap example)

bootoffav:
Hey!
enum E {
  Foo,
  Bar,
}
 
function f(x: E) {
  if (x !== E.Foo || x !== E.Bar) {
This condition will always return 'true' since the types 'E.Foo' and 'E.Bar' have no overlap.
    //
  }
}

After first comparison check fails x is left to be of type E.Bar (not E.Foo) as text suggests


orta:
Same answer as https://github.com/microsoft/TypeScript-Website/pull/2572

bootoffav:
got it, there is !==. thank you.

PR #2759, opened 2023-03-24
Add 5.0 compiler options

andrewbranch:


andrewbranch:
It would be great if all the errors were reported to me at one time instead of waiting 20 minutes to find out the next one

orta:
Yeah, I would always update locally and so going through errors was trivial - I imagine that change might need to happen inside the shiki-twoslash project as I think we throw in there?

andrewbranch:
These errors don’t appear for me locally when I run any `build` command I can find

andrewbranch:
The one that’s failing now is one I’ve already updated. Stale cache or something?

orta:
Yeah, it not showing up locally sounds like a [stale cache](https://github.com/shikijs/twoslash/blob/main/packages/remark-shiki-twoslash/src/caching.ts#L30) - I wonder why I never hit this each update

If you run `rm -rf node_modules/.cache/twoslash` it should empty the cache

WRT the server, it might be hitting the same thing but I doubt it, I think the node_modules cache gets ignored when the last build failed

milesrichardson:
Regarding `verbatimModuleSyntax`, I think the docs could be slightly more clear, in terms of how it behaves differently when it's `true` vs. `false`. Right now it just says "by default, TypeScript does import ellision..." - how does this translate to the default value for `verbatimModuleSyntax`? It's not clear from the docs, AFAICT, but I might be missing something. Additionally, the type hint of "do not transform..." is confusing because - I'm assuming - it means "if set to `true`, do not transform..."

Maybe this won't be so confusing once all the docs are pushed out and there is more information about this on the web, but I just thought I'd mention I found it confusing while upgrading tonight. 

Thanks for all the hard work on 5.0!

andrewbranch:
Error codes are scattered throughout https://github.com/microsoft/TypeScript-Website-Localizations and have to be changed there first. 🤬

PR #2761, opened 2023-03-27
Update Classes.md

a90120411:
Fix the type definition of the derived class.


orta (approved):
Looked at this, from a user's perspective, it would fix the hover on the `d` to now be `Derived`

jakebailey (approved):

PR #2766, opened 2023-03-29
Update By Example.md

a90120411:
Maybe 'Greeter' should be used instead of 'MyGreeter'.

### DIFF
--- a/packages/documentation/copy/en/declaration-files/By Example.md
+++ b/packages/documentation/copy/en/declaration-files/By Example.md
@@ -119,7 +119,7 @@ _Declaration_
 You can use a type alias to make a shorthand for a type:
 
 ```ts
-type GreetingLike = string | (() => string) | MyGreeter;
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/declaration-files/By Example.md
```diff
@@ -119,7 +119,7 @@ _Declaration_
 You can use a type alias to make a shorthand for a type:
 
 ```ts
-type GreetingLike = string | (() => string) | MyGreeter;
+type GreetingLike = string | (() => string) | Greeter;
```
jakebailey, line 122: I'm not sure this much matters for the example; theoretically both are present in this sample and are functionally equivalent.

jakebailey:
I think a better fix for this is going to be to actually declare a `Greeter` class somewhere, which the example doesn't currently do. But, the example already plays a little fast and loose with the types because it's attempting to show basic syntax, as far as I can tell.

PR #2777, opened 2023-04-06
Clarify documentation of ConstructorParameters

benjaminjkraft:
I get confused every time I use `ConstructorParameters`, because I forget it's `ConstructorParameters<typeof C>`, not `ConstructorParameters<C>` (if `C` is some class). And the documentation doesn't help, because all of its examples use builtin constructors that have special types. Of course if I spend 5 minutes thinking about it I remember that that's what makes sense, but it would be nice for the documentation to remind me more clearly so I don't have to think from first principles.

So in this commit I add an example with a user-defined class, which makes clear that you want `typeof C`.


orta (approved):
Nice idea 👍🏻

jakebailey (approved):

PR #2780, opened 2023-04-09
Grammar fix in TS For The New Programmer

seriouscatsoserious:
Original sentence structure was improper use of commas. The fix presented is sufficient, another option would be to replace the original comma with a dash.


andrewbranch:
Thank you!

PR #2784, opened 2023-04-12
Update DOM type def link in DOM Manipulation.md

thomasclaudiushuber:
The link to DOM type definitions has changed, as those are now generated and in a new folder and in a new file.


jakebailey:
This same change was accepted in #2745, but thanks for sending this!

PR #2790, opened 2023-04-16
Improve URL Structure playground handbook docs

KonstantinSimeonov:
I was trying to implement a link in code blocks on my blog site which would open the code in the typescript playground. After reading the docs, it took me a while (and actually looking at the playground source code) to figure out the correct incantation to construct the correct url. In the process, I also noticed that the docs for `#src` are off - `getInitialCode.ts` processes `#src=code`, not `#src/code`.

This PR adds more details on constructing `#code/compressed_code` urls and fixes the `#src` section in the docs :v:


orta (approved):
Nice work, sorry about that!

spss20 (approved):

jakebailey (approved):

PR #2796, opened 2023-04-20
Fixes localStorage reference causing error

### LINKED ISSUE #2762: typescript-vfs still accesses local storage even when storer is provided
Typescript-vfs module caches data is the local storage or the provided storer. This should -in theory- make typescript-vfs compatible with environments without local storage access, like the web workers. However, even when a storer is provided, the code accesses `localStorage` instead of `storelike` here, which is clearly a bug:
https://github.com/microsoft/TypeScript-Website/commit/920a54021725d2412f7963f93ec39fe4daeb37f1#diff-221727bfd6227563593353939e5f0944f6d1465c0f207a55a4a248990e821171L293

ashubham:
Fixes https://github.com/microsoft/TypeScript-Website/issues/2762


ashubham:

ashubham:
Please review @orta

orta (approved):
Yeah, looks good to me - thanks, you'll need to wait until someone with write access merges it though ( I'd recommend using patch-package till then)

jakebailey:
This LGTM but I'm not sure what's going on with CI. I'll try a close/reopen.

jakebailey:
Hm, maybe it's that version bump that is failing.

jakebailey:
Fixed the lock, but, I'm unsure if we need to also start bumping reverse deps to stop multiple versions from getting pulled in due to the use of exact versions. Not sure if there's something about version bumping I'm missing in this repo.

jakebailey (approved):

orta:
There's a script https://github.com/microsoft/TypeScript-Website/blob/36c79ce12a870332410a00304b477e3ca9877218/.github/workflows/CI.yml#L25 which will sync verify the internal references are consistent ( and update them to be afterwards)

jakebailey:
With bump versions with #2904 to minimize the waiting within the current infrastructure.

jakebailey (approved):

PR #2800, opened 2023-04-25
Explain an important limitation in decorators

wwahammy:


wwahammy:
@microsoft-github-policy-service terminate

packages/documentation/copy/en/release-notes/TypeScript 5.0.md
```diff
@@ -341,7 +341,7 @@ function loggedMethod<This, Args extends any[], Return>(
 We had to separately model out the type of `this`, the parameters, and the return type of the original method, using the type parameters `This`, `Args`, and `Return`.
 
 Exactly how complex your decorators functions are defined depends on what you want to guarantee.
-Just keep in mind, your decorators will be used more than they're written, so a well-typed version will usually be preferable - but there's clearly a trade-off with readability, so try to keep things simple.
+Just keep in mind, your decorators will be used more than they're written, so a well-typed version will usually be preferable - but there's clearly a trade-off with readability, so try to keep things simple. Also, remember that it's not currently possible for decorators to mutate the declaration it is decorating. For example, while it's possible to add properties to a class via a decorator, TypeScript itself won't recognize that new property on the class.
```
jakebailey, line 344: I'm not sure I'm convinced that this is worth mentioning in the 5.0 release notes, especially given it hasn't been possible from the beginning of decorators in TS many many years ago.

It's likely that this feature is going to be added to TS at some point in the somewhat near future, so I feel like we are safe to leave this off until then.

wwahammy: Glad to hear it's a feature that will be implemented; super excited for that!

The reason I submitted this PR is that I didn't see the limitation listed in the documentation and spent about an hour trying to get it working. It wasn't super clear to me either way so I figured if I had the confusion, others probably did it as well.

jakebailey:
Thanks for the PR; I'm going to close this as I think the blog post is not the best place to document this. The page is also a copy/paste from the blog CMS which is hosted elsewhere, so they'd end up drifting.

I think we really need to be updating the decorators page. That page is now updated to at least point out that it's documenting the "legacy" decorators, but we're going to have to have a much larger change to get all of this looking good.

PR #2809, opened 2023-04-27
Remove gist sidebar and docs about playground plugins

andrewbranch:


RyanCavanaugh (approved):

orta:
Probably also worth editing the Type or Treat blog posts e.g. https://devblogs.microsoft.com/typescript/type-treat-2021-day-1/ if you can , as these now won't work

andrewbranch:
I’m not sure how to do that, and Daniel is out, but the Azure service is already down so I’m going to go ahead and merge on green

RyanCavanaugh:
Blog links (all the ones I could find anyway) are updated to the underlying gist

PR #2811, opened 2023-04-28
Possible typo.

kumarUjjawal:
Shouldn't it be  `sanitizeInput()` as return rather than `sanitize()`.


orta:
No, this would be an infinite loop

PR #2812, opened 2023-04-28
Made the utility of InstanceType clearer

patelatharva:


packages/documentation/copy/en/reference/Utility Types.md
```diff
@@ -403,7 +403,7 @@ Released:
 
 </blockquote>
 
-Constructs a type consisting of the instance type of a constructor function in `Type`.
+Constructs a type consisting of the instance type of a constructor function in `Type`. Declaring a variable to have `InstanceType` ensures that only an object that is an instance of the given type can be assigned to it and not any other object that may have similar properties.
```
jakebailey, line 406: I'm not sure that this is directly applicable to `InstanceType` more than any other type; the whole selling point of types is that "only an object of a certain type can have things of that type assigned to it".

Given this example doesn't contain any assignments, I don't think that this change adds much compared to people having read up on assignability.

jakebailey (changes requested):

PR #2815, opened 2023-04-30
Add documentation about re-exporting modules as a namespace

LoganDark:
We hit this when trying to compile code for a webpage that had packages loaded into global variables. It took us days to find out the solution and it caused tons and tons of stress. The docs really need to be clearer about this.

-Emily

### DIFF
--- a/packages/documentation/copy/en/reference/Modules.md
+++ b/packages/documentation/copy/en/reference/Modules.md
@@ -633,6 +633,23 @@ It can also be used as a global variable, but only inside of a script.
 mathLib.isPrime(2);
 ```
 
+#### Re-exporting modules into a global namespace
+
+Some environments have existing ES6 modules available under global variables instead.
+For example, on a web page, libraries might be loaded directly by `<script>` tags and create global variables for other scripts to use.
+Their type definitions might not account for this if the modules were not designed to be loaded this way.
+However, `export as namespace` can also be used from ambient modules to re-export other libraries.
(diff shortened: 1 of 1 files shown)


LoganDark:

LoganDark:
wtf? after making this PR I just randomly got an email in Chinese from some random website claiming to be [MurphySec](https://github.com/murphysecurity/murphysec) (but they were not murphysec.com). I guess I got picked up by some spambots

andrewbranch:
Thank you for your recent pull request for documentation! We appreciate your time and effort in submitting your work for review.

This seems to be a workaround for a bug in the typings of an external dependency, and has its own caveats that would need to be spelled out, but may be too niche of a topic for these reference pages. We would always encourage users who encounter buggy dependency typings to file an issue or PR with the library / DefinitelyTyped typings rather than implement a workaround like this.

LoganDark:
> This seems to be a workaround for a bug in the typings of an external dependency

It's... not? There's nothing wrong with the library's typings. It's just the the typings are ES6 modules and this is being used in the browser through a global variable. This isn't a workaround for anything, this is literally just declaring the existence of a global variable that contains a library. It could be any library and any variable name, and it's not the library's job to declare itself a global variable if it's not normally used in a browser.

> We would always encourage users who encounter buggy dependency typings to file an issue or PR with the library / DefinitelyTyped typings rather than implement a workaround like this.

The dependency typings are perfectly fine and this is not a workaround for anything. Again this is just declaring a global variable that contains a library. This is a useful feature for compiling TypeScript for environments that already have global modules, and the fact that it's not easily discoverable by developers is a big problem.

PR #2816, opened 2023-05-01
Update Narrowing.md

schoblaska:
The error here is not because we're "adding a `number | string` to a `number `", it's because we're passing a `number | string`  to `String.repeat`, which only accepts a `number`.

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABABwIYBMAyBTYUAUa66MYA5gFyJggC2ARtgE6IA+iAzlE6WQDSJSyEFCpce5AJSIA3gChEiJtiggmSAESINAOmXJsqAkRJTEAakFhhUANxyAvkA


andrewbranch (approved):
Thanks!

PR #2819, opened 2023-05-05
Update The Handbook.md

acmu:


andrewbranch:
I don’t think this is a meaningful change. Would be open to a PR spelling out “object oriented” instead of using the acronym though.

acmu:
> I don’t think this is a meaningful change. Would be open to a PR spelling out “object oriented” instead of using the acronym though.

I change that because of this:
image

image

page title and link title is not same,I think they should be same.

andrewbranch:
Oh, I see. Which title is more recent according to `git blame`?

acmu:
> Oh, I see. Which title is more recent according to `git blame`?

I can't distinguish which is recent, becaus they are same pull request and same time.

But according to article's content, I think it should be "TypeScript for Java/C# Programmers".

As for another, you can determine it.(I'm not good at English😅)

andrewbranch (approved):
Thanks for the explanation, this title seems to be more consistent.

PR #2827, opened 2023-05-14
docs: Proposed link updates for TS learning paths

mehicalen:
While reading the TypeScript documentation, I went from [The TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/intro.html) intro page to [TypeScript for JavaScript Programmers](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html) page which then leads me back intro page in the "Next Steps" section. This PR proposes a change to direct users directly to first section of the handbook instead of the intro page.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
@@ -593,5 +593,5 @@ consistently used in TypeScript code.
 
 This doc is a high level overview of the syntax and types you would use in everyday code. From here you should:
 
-- Read the full Handbook [from start to finish](/docs/handbook/intro.html)
+- Read the full Handbook [from start to finish](/docs/handbook/2/basic-types.html)
 - Explore the [Playground examples](/play#show-examples)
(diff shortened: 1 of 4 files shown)


mehicalen:

packages/documentation/copy/en/get-started/TS for Functional Programmers.md
```diff
@@ -593,5 +593,5 @@ consistently used in TypeScript code.
 
 This doc is a high level overview of the syntax and types you would use in everyday code. From here you should:
 
-- Read the full Handbook [from start to finish](/docs/handbook/intro.html)
+- Read the full Handbook [from start to finish](/docs/handbook/2/basic-types.html)
```
jakebailey, line 596: I think that if anything, we should explicitly state that this is going to skip ahead in the handbook, like the intro page itself. Something like:

> Get started with The Basics

Which is the beginning of the final section of the intro.

jakebailey (changes requested):

jakebailey:
On a second look, I'm not sure that it's a good idea to link to the specific page anyhow; it may be the case that the first page changes, or is moved. Or, the user may go back to the main page and read a different guide.

PR #2831, opened 2023-05-19
Replace newlines with spaces when one-lining inlay hints

jakebailey:
Theoretically fixes how bad this looks after #2691

image

### DIFF
--- a/packages/playground/src/twoslashInlays.ts
+++ b/packages/playground/src/twoslashInlays.ts
@@ -32,7 +32,7 @@ export const createTwoslashInlayProvider = (sandbox: Sandbox) => {
         if (!hint || !hint.displayParts) continue
 
         // Make a one-liner
-        let text = hint.displayParts.map(d => d.text).join("").replace(/\r?\n/g, " ").replace(/  /g, "")
+        let text = hint.displayParts.map(d => d.text).join("").replace(/\s+/g, " ")
         if (text.length > 120) text = text.slice(0, 119) + "..."
 
         const inlay: import("monaco-editor").languages.InlayHint = {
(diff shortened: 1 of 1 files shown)


jakebailey:
Hoping I get a staging link but I suspect not because I used my fork to send the PR

packages/playground/src/twoslashInlays.ts
```diff
@@ -32,7 +32,7 @@ export const createTwoslashInlayProvider = (sandbox: Sandbox) => {
         if (!hint || !hint.displayParts) continue
 
         // Make a one-liner
-        let text = hint.displayParts.map(d => d.text).join("").replace(/\r?\n/g, "").replace(/  /g, "")
+        let text = hint.displayParts.map(d => d.text).join("").replace(/\s+/g, " ")
```
jakebailey: I'm not a fan of this, but the reality is that this code already does "the wrong thing" by nature of touching the contents of the tooltip. It's totally possible to have a string literal type which consists of multiple spaces and yet the old code would futz with it.

The old code (`/  /g, ""`) would also do very odd things like `" " -> " "`, but `"  " -> ""` and then `"   " -> " "`.

andrewbranch: Maybe replacing `/\r?\n\s*` would be better? Assuming you’re just trying to get rid of leading indentation in multi-line object types

jakebailey: That's a great idea!

jakebailey:
Seems like it needs some other change: 
image

Will look later.

jakebailey:
Nevermind, that was the old deployment without my last fix. Looks good now!

image

PR #2832, opened 2023-05-21
Update Everyday Types for clarity

nickangtc:
### The problem

I was reading [Everyday Types > Literal Inference section](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-inference) for the first time to refresh my knowledge on TypeScript, and when I came across the example, I was very confused:

CleanShot 2023-05-21 at 11 11 07@2x

The confusion stems from ambiguity that I think needs to be edited to be clearer. I have some experience using TypeScript, and even so, I wasn't sure what the example's warning was about. Fairly certain that newcomers reading the doc for the first time will be more confused than me, so let's fix that for them.

### The proposed fix

I think the most suitable fix, which happens to be a one-liner, is to allow the function definition of `handleRequest` to render once on the page. The first time it is shown on the page, show the function definition, so readers will immediately understand - _oh, `handleRequest` has the following signature, and `req.method` has type `string` that is too broad_...  

```js
declare function handleRequest(url: string, method: "GET" | "POST"): void;
```

I'm not very familiar with `twoslash` but I think removing `// ---cut---` in one of the code blocks should work? 

### Counter-arguing the fix

I can see why you might not want to fix this, since technically one could view the type definition of `handleRequest` by hovering over the code like so:

CleanShot 2023-05-21 at 11 32 49@2x

I'm just a passerby who has already benefitted from the amazing docs at typescriptlang.org, and I'm sharing that it was confusing for me, which probably means it was confusing to some others who don't know that the code blocks are souped-up code blocks where hovering shows type definitions like in an IDE. The dotted underlines are also quite hard to perceive in dark mode.

No strong opinion here – please merge if you want to, or close the PR if you don't wish to.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -597,7 +597,7 @@ The same applies to strings:
 ```ts twoslash
 // @errors: 2345
 declare function handleRequest(url: string, method: "GET" | "POST"): void;
-// ---cut---
+
 const req = { url: "https://example.com", method: "GET" };
 handleRequest(req.url, req.method);
 ```


jakebailey:
Looking at https://delightful-glacier-09aa73710-2832.centralus.azurestaticapps.net/docs/handbook/2/everyday-types.html#literal-inference

This seems fine but the removed line could instead be an empty line for clearer reading.

packages/documentation/copy/en/handbook-v2/Everyday Types.md
```diff
@@ -597,7 +597,6 @@ The same applies to strings:
 ```ts twoslash
 // @errors: 2345
 declare function handleRequest(url: string, method: "GET" | "POST"): void;
-// ---cut---
 const req = { url: "https://example.com", method: "GET" };
```
jakebailey, line 601:
```suggestion

const req = { url: "https://example.com", method: "GET" };
```

nickangtc: Oh, agreed. Updated [here](https://github.com/microsoft/TypeScript-Website/compare/8b852733cf9aca4dc5ed9c94a8b0ab3f4e71b725..17517b8f5240a1bbb14993368ba009733a614d2e)

jakebailey (approved):

PR #2834, opened 2023-05-23
fix: Grammar and formatting in `exactOptionalPropertyTypes`

rschristian:
Simply correcting a few minor issues I noticed when checking up on this option

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/exactOptionalPropertyTypes.md
+++ b/packages/tsconfig-reference/copy/en/options/exactOptionalPropertyTypes.md
@@ -16,7 +16,7 @@ interface UserDefaults {
 
 Without this flag enabled, there are three values which you can set `colorThemeOverride` to be: "dark", "light" and `undefined`.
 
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/copy/en/options/exactOptionalPropertyTypes.md
```diff
@@ -16,7 +16,7 @@ interface UserDefaults {
 
 Without this flag enabled, there are three values which you can set `colorThemeOverride` to be: "dark", "light" and `undefined`.
 
-Setting the value to `undefined` will allow most JavaScript runtime checks for the existence to fail, which is effectively falsy. However, this isn't quite accurate `colorThemeOverride: undefined` is not the same as `colorThemeOverride` not being defined. For example `"colorThemeOverride" in settings` would have different behavior with `undefined` as the key compared to not being defined.
```
jakebailey, line 19: The original text is the actual expression, like `"key" in obj`; I don't think this change is correct.

jakebailey (changes requested):

packages/tsconfig-reference/copy/en/options/exactOptionalPropertyTypes.md
rschristian: Whoops, that didn't even dawn on me.

Mixing expressions in the middle of a sentence is a little difficult to understand though, as seen here, I assumed typos/formatting issues before the real intent. Moving that into the code block is an idea, but I'll revert.

jakebailey (approved):

Issue #2829, opened 2023-05-17
Accessibility improvements


DanielRosenwasser:
We have an internal accessibility team. This issue mostly provides a way to keep track of that work on our public issue tracker and come back to it from our iteration plan.

PR #2836, opened 2023-05-24
Fixing accessibility bugs

navya9singh:
These improvements fix the following bugs:

1. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1798038
2. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1798021
3. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1799731
4. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1799747
5. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1802376
6. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1802380
7. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1803076
8. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1797977
9. https://devdiv.visualstudio.com/DevDiv/_queries/edit/1779572/?triage=true
10. https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1802371

 I have tested these changes using NVDA and Accessibility Insights for Web.

https://github.com/microsoft/TypeScript-Website/issues/2829


RyanCavanaugh (approved):

packages/typescriptlang-org/src/components/layout/Sidebar.scss
```diff
@@ -143,7 +143,7 @@ nav#sidebar {
       }
 
       &.highlight a {
-        color: $ts-main-blue-color;
+        color: #2866a9;
```
jakebailey: Where does this color come from?

navya9singh: To ensure that the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds, this is the suggested color by Accessibility Insights keeping the background color, font size and font weight all the same.

jakebailey: This doesn't feel right... Before:

image

After:

image

This seems like we're getting less contrast, not more...

orta: Might be dark mode vs light mode too - it's been a few years but I think those are changed via those constants

navya9singh: Yeah that's true, it has to do with the dark vs light mode. Let me see what to do about that.

jakebailey: Yeah, I think that if we're tweaking colors, we probably need to be introducing new variables and using them rather than hardcoding color values in specific places. e.g. we could include a "ts-sidebar-link" color or something for this, right?

jakebailey: I'm confused, we added a variable, but did it actually get applied only to the light theme?

PR #2837, opened 2023-05-25
Update Decorators.md

xupea:
Already updated to stage 3 for one year:
https://github.com/tc39/proposal-decorators/commit/232899356421269b056e1c1e3d6c68d9c332be88#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R3


jakebailey (changes requested):
If we're going to change this, we can't just change the number because the whole section doesn't make sense now that TS supports _two_ kinds of decorators. See https://github.com/microsoft/TypeScript-Website/pull/2348#issuecomment-1552220140 (and the other 14 linked PRs which also tried to submit this change).

jakebailey:
To further clarify, I don't think it's a good idea to just change the number from 2 to 3, because that will make the section imply that to get stage 3 decorators, you need to pass `--experimentalDecorators`, which isn't true.

PR #2839, opened 2023-05-27
Describe defaults of jsconfig.json

MatthewPinheiro:
Adds a section to describe the changes that a `jsconfig.json` makes over a `tsconfig.json`, which is very helpful when figuring out why behavior such as https://github.com/microsoft/TypeScript/issues/50862 seems to occur. 

Defaults referenced from TypeScript src in [`getDefaultCompilerOptions`](https://github.com/microsoft/TypeScript/blob/main/src/compiler/commandLineParser.ts#L3395) and [`getDefaultTypeAcquisition`](https://github.com/microsoft/TypeScript/blob/main/src/compiler/commandLineParser.ts#L3413).

### DIFF
--- a/packages/documentation/copy/en/project-config/tsconfig.json.md
+++ b/packages/documentation/copy/en/project-config/tsconfig.json.md
@@ -109,7 +109,7 @@ The `"compilerOptions"` property can be omitted, in which case the compiler's de
 
 ## JSConfig Defaults
 
-A `jsconfig.json` has the following defaults over a `tsconfig.json`:
+A `jsconfig.json` enforces different defaults, being equivalent to the following `tsconfig.json`:
 
 ``` json tsconfig
 {
--- a/packages/documentation/copy/en/project-config/tsconfig.json.md
+++ b/packages/documentation/copy/en/project-config/tsconfig.json.md
@@ -11,7 +11,7 @@ translatable: true
 The presence of a `tsconfig.json` file in a directory indicates that the directory is the root of a TypeScript project.
 The `tsconfig.json` file specifies the root files and the compiler options required to compile the project.
 
-JavaScript projects can use a `jsconfig.json` file instead, which acts almost the same but has some JavaScript-related compiler flags enabled by default.
+JavaScript projects can use a `jsconfig.json` file instead, which acts almost the same but has some [JavaScript-related flags enabled by default](#jsconfig-defaults).
 
 A project is compiled in one of the following ways:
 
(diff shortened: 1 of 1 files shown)


MatthewPinheiro:

packages/documentation/copy/en/project-config/tsconfig.json.md
```diff
@@ -107,6 +107,25 @@ This lets your `tsconfig.json` focus on the unique choices for your project, and
 
 The `"compilerOptions"` property can be omitted, in which case the compiler's defaults are used. See our full list of supported [Compiler Options](/tsconfig).
 
+## JSConfig Defaults
```
jakebailey, line 110: This doesn't feel like the right place to be defining these defaults, given the existing contents of the page (very short, minimal examples, links to other pages). This isn't even a page I think people visit, compared to https://www.typescriptlang.org/tsconfig.

Maybe it'd be clearer to put these into the actual tsconfig page per-option as part of the defaults?

MatthewPinheiro: I'm not opposed to also listing the defaults with each option on the full reference page, but I chose this page to centralize it because it's the _only_ place on the entire site that even has a full sentence about the `jsconfig.json`. Other pages may make a reference to it, but if they even do that, they'll often link back to this page as well. It's also the first site result for "jsconfig" in both the built-in site search and google (the full reference page never comes up in either).

Actually, the first google result that comes up at all is the [jsconfig.json reference from vscode](https://code.visualstudio.com/docs/languages/jsconfig). It's a little weird that a reference page for the jsconfig.json exists on VSC docs but not on the TS docs, but unless you'd be in favor of a dedicated jsconfig.json page I think keeping this here would be a happy medium? I think only adding the defaults per-option in the reference would not be great for SEO and would continue to make it feel hidden, and at least anecdotally and based on quite a few GitHub issues, I think others feel that's currently the case.

rbuckton: I think I agree with @MatthewPinheiro's reasoning. https://www.typescriptlang.org/tsconfig is more of a reference page with very specific per-option documentation, while https://www.typescriptlang.org/docs/handbook/tsconfig-json.html is structured more like an article, explaining the "why"s and "how"s of the feature in broader strokes.

I think this section makes sense here, but also agree with @jakebailey that we should update https://www.typescriptlang.org/tsconfig to indicate where per-option defaults differ between `tsconfig.json` and `jsconfig.json`.

MatthewPinheiro: Just checking in to ask what next steps from here would be? Since @rbuckton agrees with adding to both pages, should I add the per-option JSConfig defaults to the reference to this PR as well?

jakebailey: Sure; each of the options has its own `md` file, and if any of them differ, you should be able to note that in the text there at the bottom or something.

Feasibly, the TS code that generates the default column could be changed, but that may get a little hairy for info that few people need.

thernstig:
Aside: I feel it would make more sense and be easier to explain if `jsconfig.json` was deprecated, and instead only `tsconfig.json` was allowed. But that there was a `javascript` base added to https://github.com/tsconfig/bases. Where users then would have to install it and use `extends` and refer to it.

The rationale being is that it is much more clear for readers of `tsconfig.json` what **is actually** set. Users reading `tsconfig.json` could easily see what is overridden and set. There could also be explanations for each option overridden and the **rationale** why they are overridden on the TS bases website (or even in the `json` file since I believe it is read as `jsonc` by `tsc`.) Now there is no rationale at all for some options. Like why is `"skipLibCheck": true,` set for `jsconfig.json` and not `tsconfig.json`?

That way, no defaults would have to be explained specifically in https://www.typescriptlang.org/tsconfig, which is already large and quite daunting/complex for users.

If this makes sense I could break it out into its own issue, as it is half-and-half unrelated to this issue here. If anything should be done in this issue, I feel that you should add the rationale in the example (with comments) for each option and why it differs.

PR #2840, opened 2023-05-28
fix: remove-display-table for mobile device

candymask0712:
While I reading typescript handbook with tablet PC, I found that part of document is hidden in tablet PC screen.
스크린샷 2023-05-28 오후 6 01 08

To fix this I remove 'display: table' property from 'table.full-width-table'
스크린샷 2023-05-28 오후 6 03 17

table.full-width-table selector is used in 2 pages. (one of them is deprecated)
and after removing property, it works well.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -387,6 +387,7 @@ Being concerned only with the structure and capabilities of types is why we call
 Type aliases and interfaces are very similar, and in many cases you can choose between them freely.
 Almost all features of an `interface` are available in `type`, the key distinction is that a type cannot be re-opened to add new properties vs an interface which is always extendable.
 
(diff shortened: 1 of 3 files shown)


jakebailey (changes requested):
Comparing the live site with this PR, this breaks the table when there _is_ enough space:

- https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#interfaces
- https://delightful-glacier-09aa73710-2840.centralus.azurestaticapps.net/docs/handbook/2/everyday-types.html#interfaces

Now:

image

This PR:

image

candymask0712:

candymask0712:
@jakebailey 
I'm sorry. I did not notice that my code changes original layout of document. Since I checked the page with my laptop. Therefore It seemed like layout was same as original one. I fixed my code. Please check it.

jakebailey (approved):

PR #2841, opened 2023-05-30
fix logical order of documentation Object Types

FalkoKa:
Going over the documentation, I saw this minor logical error. The author said that 'index signatures' will be discussed later, however, they just got discussed in the previous section.


FalkoKa:

packages/documentation/copy/en/handbook-v2/Object Types.md
```diff
@@ -419,7 +419,7 @@ interface SquareConfig {
 }
 ```
 
-We'll discuss index signatures in a bit, but here we're saying a `SquareConfig` can have any number of properties, and as long as they aren't `color` or `width`, their types don't matter.
+As discussed in the section index signatures, we can saying that `SquareConfig` can have any number of properties, and as long as they aren't `color` or `width`, their types don't matter.
```
jakebailey:
```suggestion
Here we're saying that `SquareConfig` can have any number of properties, and as long as they aren't `color` or `width`, their types don't matter.
```

This was just the section previously; this little comment was previously intended to be "don't worry about this syntax, we'll describe it later", but since it's already introduced in the current page, we can just not mention it because knowledge of it is assumed.

jakebailey (changes requested):
This PR checks in a bunch of random lockfile changes.

jakebailey:
This PR hasn't changed since I last reviewed it but you pinged me; have you forgotten to push something?

jakebailey (approved):

PR #2844, opened 2023-05-31
Fix accessibility issue for css variable highlight

jakebailey:
For #2840

We wasted a bunch of time testing this before I stumbled on the problem that there's some cache somewhere that duplicates this theme, such that changing it didn't actually work. Cleaning the repo and reinstalling/rebuilding/etc actually shows that changing this _does_ work.

This color upstream is https://github.com/shikijs/shiki/blob/main/packages/shiki/themes/light-plus.json#L294 (the same as what the accessibility team/tool say to use).


navya9singh (approved):

jakebailey:
Works:

image

PR #2849, opened 2023-06-03
Update Narrowing.md

dnamkoong:
The first example under `Exhaustiveness checking` should not raise an error

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -710,7 +710,7 @@ In those cases, TypeScript will use a `never` type to represent a state which sh
 
 The `never` type is assignable to every type; however, no type is assignable to `never` (except `never` itself). This means you can use narrowing and rely on `never` turning up to do exhaustive checking in a `switch` statement.
 
-For example, adding a `default` to our `getArea` function which tries to assign the shape to `never` will not raise an error when every possible case has not been handled.
+For example, adding a `default` to our `getArea` function which tries to assign the shape to `never` will not raise an error when every possible case has been handled.
 
 ```ts twoslash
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Narrowing.md
```diff
@@ -710,7 +710,7 @@ In those cases, TypeScript will use a `never` type to represent a state which sh
 
 The `never` type is assignable to every type; however, no type is assignable to `never` (except `never` itself). This means you can use narrowing and rely on `never` turning up to do exhaustive checking in a `switch` statement.
 
-For example, adding a `default` to our `getArea` function which tries to assign the shape to `never` will raise an error when every possible case has not been handled.
+For example, adding a `default` to our `getArea` function which tries to assign the shape to `never` will not raise an error when every possible case has not been handled.
```
jakebailey: This isn't right; the word `not` already appears later in the sentence ("when every possible case has not been handled").

But, I think it may be worth moving `not` forward to where you have it now and drop the second not. That way the fact that there's no error is right up front.

jakebailey (changes requested):

packages/documentation/copy/en/handbook-v2/Narrowing.md
dnamkoong: Sounds good

jakebailey: You closed this and pinged me for a review, but haven't updated the PR; is there a change you have forgotten to push to address the above?

dnamkoong: Thought I updated the PR, change has been pushed

jakebailey (approved):

PR #2850, opened 2023-06-03
note on stage3 decorators

amitport:

PR #2857, opened 2023-06-12
Update Decorators.md

lor3k:
29.03.2022 it was changed to Stage 3

https://github.com/tc39/proposal-decorators/commit/232899356421269b056e1c1e3d6c68d9c332be88


jakebailey:
Per https://github.com/microsoft/TypeScript-Website/pull/2348#issuecomment-1552220140, it's not enough to just change this version; the stage 3 decorator spec is completely different than what is described in the page.

#2850 will cover this.

PR #2860, opened 2023-06-16
Fix diagnostics.md

sinde530:
compile => compiled


jakebailey:
"a compile" is a valid phrase; compile is a noun.

PR #2864, opened 2023-06-19
docs: update Module Resolution.md

### LINKED ISSUE #2863: Docs: "types" and "typings" are synonymous in Module Resolution.md
**Page URL:** https://www.typescriptlang.org/docs/handbook/module-resolution.html#how-typescript-resolves-modules

**Issue:** `package.json` can contain any of the two fields `types` and `typings`. When TypeScript resolves a module it may look for either of the two fields. It would be beneficial from a clarity and documentation standpoint if the `Module Resolution.md` mentioned that the `typings` field can be used interchangeably with the `types` field.

**Recommended fix:** Rewrite for clarity in Module Resolution.md under the following section:
1. Add a note under "How TypeScript resolves modules" mentioning `typings` is also permissible in `package.json`
2. Under relative import pt. 4, mention `typings` within parenthesis.
3. Under non-relative import pt. 4, 12, and 20, mention `typings` within parenthesis.
4. Leave output under "Tracing module resolution" in the current state as it is not incorrect.
5. Add a note under "Things to look out for" pt 3, indicating that the output may contain a `typings` field.

HamdaanAliQuatil:
Closes #2863 

**- What I did**

Update `Module Resolution.md` to include clarification that `types` and `typings` are synonymous during module resolution in TypeScript.

**- How I did it**

- [x] Added a note under "How TypeScript resolves modules" mentioning `typings` is also permissible in `package.json`
- [x] Under relative import pt. 4, mentioned `typings` within parenthesis.
- [x] Under non-relative import pt. 4, 12, and 20, mentioned `typings` within parenthesis.
- [x] Added a note under "Things to look out for" pt 3, indicating that the output may contain a `typings` field.

### DIFF
--- a/packages/documentation/copy/en/reference/Module Resolution.md
+++ b/packages/documentation/copy/en/reference/Module Resolution.md
@@ -140,12 +140,14 @@ TypeScript will mimic the Node.js run-time resolution strategy in order to locat
 To accomplish this, TypeScript overlays the TypeScript source file extensions (`.ts`, `.tsx`, and `.d.ts`) over Node's resolution logic.
 TypeScript will also use a field in `package.json` named `types` to mirror the purpose of `"main"` - the compiler will use it to find the "main" definition file to consult.
 
+Please note that the field `typings` is equivalent to `types` and can be used interchangeably. Thus, it is permissible for `package.json` to contain any of the two fields.
+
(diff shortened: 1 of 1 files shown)


orta:
Typings is the older and largely deprecated name for this field, I feel like the team discussed this and opted to not include it in a few places during the tsconfig refernce also - so this PR probably isn't neededIMO

jakebailey:
Yeah; I don't think we want anyone using typings at this point.

PR #2865, opened 2023-06-19
Update hybrid esm package.json example

ajwootto:
Change fall-back "types" field to point to the `.d.cts` file so Typescript correctly resolves the CommonJS types

addresses https://github.com/microsoft/TypeScript/issues/54620


andrewbranch:
Answered why this example is intentional in https://github.com/microsoft/TypeScript/issues/54620

PR #2875, opened 2023-07-05
copy/en/get-started: make `type/value` distinction

chee-zaram:
This change tries to make the distinction between a type and a value more explicit. Indeed, in the TS world, something such as `"a string"` could be well regarded as a type, but it is more naturally known as a value. Given that this piece of information is in a learners' section, I believe that the change is worth making as it could be confusing.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for JS Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for JS Programmers.md
@@ -128,7 +128,7 @@ With TypeScript, you can create complex types by combining simple ones. There ar
 
 ### Unions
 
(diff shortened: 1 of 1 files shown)


jakebailey:
I'm not sure that I like this change; `"a string"` is certainly a type (a literal type). I'm not sure that we want to ambiguate type space by including values.

chee-zaram:
Okay. That's fine.

chee-zaram:
@jakebailey would you like me to close this PR or will you?

jakebailey:
I'll click it; thanks for sending the PR, though!

PR #2881, opened 2023-07-10
Prevent minified files from cluttering `git grep` results

not-my-profile:


jakebailey (approved):
Not sure that this will help retroactively, but it doesn't seem like a bad idea.

Issue #2776, opened 2023-04-06
Dead external link.


MicahZoltu:
https://github.com/microsoft/TypeScript-Website/blob/62f87a80f571762ff252eb38b53b890f9f864acb/packages/documentation/copy/en/release-notes/TypeScript%202.6.md?plain=1#L15

This link is dead, and I'm not sure what to link to instead.

shuuji3:
While I couldn't find another URL for the page, it looks like the blog was down due to Heroku's decision to shut down its free plan.

I found the archived page by Internet Archive: http://web.archive.org/web/20220823104433/https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance Maybe we could temporarily replace the dead link with this one.

agilgur5:
Looks like this was fixed by OP in #2808

shuuji3:
Thanks for the heads-up.🙂  It seems that the new link was broken again so I'd raised a new PR to ensure the stable URL.

PR #2889, opened 2023-07-18
docs: fix broken link in `TypeScript 2.6.md`

shuuji3:
Hello, this is the follow-up PR for #2808.

The previous PR changed to using archive.md on April but now it's already inaccessible. This is because this archive site historically changed its domain many times due to the crawls ignoring robots.txt. (ref. https://en.wikipedia.org/wiki/Archive.today)

This changes to use the reliable Internet Archive link so that we won't need to change URLs in the future. (Additional information: https://github.com/microsoft/TypeScript-Website/issues/2776#issuecomment-1500269003)


MicahZoltu (approved):

MicahZoltu (approved):

shuuji3:
@MicahZoltu Thanks for the review :+1:

jakebailey (approved):

PR #2892, opened 2023-07-18
fix: remove redundant `ES2020` section in `module`

agilgur5:
## Summary

`ES2020` is referenced twice under `module`

## Details

- `ES2020` is already covered in the section with `ES2015` - `ES2022`
  - same exact output and already mentioned there
  - easier to read and more consistent

- did not touch `ESNext` as that isn't mentioned in the below section
  - and theoretically could change in the future
  
### Screenshot

Screenshot 2023-07-18 at 1 21 42 PM

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/module.md
+++ b/packages/tsconfig-reference/copy/en/options/module.md
@@ -74,17 +74,6 @@ import { valueOfPi } from "./constants";
 export const twoPi = valueOfPi * 2;
 ```
 
-#### `ES2020`
-
-```ts twoslash
-// @showEmit
-// @module: es2020
-// @noErrors
-import { valueOfPi } from "./constants";
-
-export const twoPi = valueOfPi * 2;
-```
-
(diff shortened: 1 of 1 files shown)


jakebailey:
I'm not sure this is right (before or after); my recollection is that there are in fact differences between 2015, 2020, and 2022, which should be document but may not be in full. Something to do with which ones can do top-level-await, `import.meta`, etc. @andrewbranch ?

agilgur5:
Yes, as you can see in the screenshot, those differences are enumerated at the bottom of the `ES2015` - `ES2022` section. However, those differences do not change the compiled output of the example code (as the example code does not make use of those features). And they are not mentioned in the duplicate `ES2020` section that I remove here.

As in to say, this change makes no semantic changes nor removes any existing descriptions, just a duplicate that is the odd one out.

jakebailey:
Oops, completely missed that; I skipped right to the diff.

jakebailey (approved):

PR #2894, opened 2023-07-18
feat: format and link `package.json` Defaults

agilgur5:
## Summary

Add consistent formatting and linking to `resolvePackageJsonExports` / `Imports`' "Defaults" section

## Details

- previously the defaults did not consistently use backticks
  - unlike the rest of the docs

- also add a link to `moduleResolution` when it is referenced in the defaults
  - consistent with existing docs
  
- no actual content changes, just formatting/linking

### Screenshot

The right side "Default" did not have consistent formatting previously:

Screenshot 2023-07-18 at 1 30 55 PM

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -240,6 +240,14 @@ export const defaultsForOptions = {
   noImplicitThis: trueIf("strict"),
   preserveConstEnums: trueIf("isolatedModules"),
   reactNamespace: "React",
+  resolvePackageJsonExports: [
+    "`true` when [`moduleResolution`](#moduleResolution) is `node16`, `nodenext`, or `bundler`;",
+    "otherwise `false`",
+  ],
+  resolvePackageJsonImports: [
+    "`true` when [`moduleResolution`](#moduleResolution) is `node16`, `nodenext`, or `bundler`;",
+    "otherwise `false`",
+  ],
   rootDir: "Computed from the list of input files.",
   rootDirs: "Computed from the list of input files.",
   strictBindCallApply: trueIf("strict"),


not-my-profile:
While I see that this repository apparently defines a bunch of markdownified default strings, I think we should actually avoid manually overriding these defaults here at all.

This message appears to be defined in [`diagnosticMessages.json` in the typescript repo](https://github.com/microsoft/TypeScript/blob/97ef321fa04b95e00ec0a3ff18d40457dd0e8c67/src/compiler/diagnosticMessages.json#L5625), which apparently often uses `'` instead of `` ` `` ... so I think it would make much more sense to consistently convert the quotation marks via a regex. And the same with the linkification ... we should also be able to do this automatically via regex.

Because if we hardcode the default strings in the documentation, they will just get out of date when they change in the typescript repository.

agilgur5:
> Because if we hardcode the default strings in the documentation, they will just get out of date when they change in the typescript repository.

I don't disagree with this sentiment, but this could be said about the entirety of the docs, not just this one configuration.

If there were a more holistic approach to generate large portions of these docs, not just this one configuration, then this would be more consistent in that broader context.

IMO that would make more sense as a separate PR that generates more than just these two lines. The context of this PR itself is pretty minimal.

> so I think it would make much more sense to consistently convert the quotation marks via a regex.

A regex could be similarly brittle as well; it may not apply when the string is changed. Regexes are also significantly less readable than a markdown string. That makes them less maintainable, especially for what is only 2 LoC.


> This message appears to be defined in [`diagnosticMessages.json` in the typescript repo](https://github.com/microsoft/TypeScript/blob/97ef321fa04b95e00ec0a3ff18d40457dd0e8c67/src/compiler/diagnosticMessages.json#L5625)

Nice find! In that specific file, the key _is_ the message, so accessing it requires knowing the exact string as well 😕
It's been a while since I've touched the diagnostic messages of TS though, so I don't quite remember all the nuances here.

agilgur5:
[Preview](https://delightful-glacier-09aa73710-2894.centralus.azurestaticapps.net/tsconfig#resolvePackageJsonExports) looks correct to me:

Screenshot_2023-07-20-19-28-33-83_40deb401b9ffe8e1df2f1cc5ba480b12

jakebailey:
@andrewbranch may have imported these more recently than I have; it does seem unfortunate to have to copy and paste the entire thing to fix this up, but I'm also unsure where they come from.

andrewbranch:
We cannot afford to overthink automation of content on this site 😅

PR #2895, opened 2023-07-18
docs: consistently use semicolons for Defaults

agilgur5:
## Summary

Attempt to clarify long lists of Defaults with lots of commas by using some semicolons instead

## Details

- when multiple options are listed out, there A TON of commas, so separating out pieces can be tough
  - use a semicolon instead of a comma between different options to clarify the separations
    - could also use a period, but a semicolon is consistent with the `resolvePackageJsonExports` / `Imports` docs (see also #2894)
    
- also add a missing Oxford comma in `moduleResolution`

- this should hopefully make it _a bit_ easier on the eyes 😅

### Screenshot

`moduleResolution` is a particularly good example of how hard this is to read currently:

Screenshot 2023-07-18 at 1 56 08 PM

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -189,7 +189,7 @@ export const relatedTo: [AnOption, AnOption[]][] = [
 
 function trueIf(name: string) {
   return [
-    `\`true\` if [\`${name}\`](#${name}),`,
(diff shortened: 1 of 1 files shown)


agilgur5:
seems slightly more readable in [the preview](https://delightful-glacier-09aa73710-2895.centralus.azurestaticapps.net/tsconfig#moduleResolution) 😅 :

Screenshot 2023-07-18 at 7 57 34 PM

PR #2897, opened 2023-07-18
feat: add `package.json` options to Related

agilgur5:
## Summary

New features like `customConditions`, `resolvePackageJsonExports`, and `resolvePackageJsonImports` are Related to `moduleResolution`

## Details

- these influence how `moduleResolution` works, and so should be Related to it
  - and mention `moduleResolution` in their description as well (see also #2891)
- and also Related to each other

### DIFF
--- a/packages/tsconfig-reference/scripts/tsconfigRules.ts
+++ b/packages/tsconfig-reference/scripts/tsconfigRules.ts
@@ -167,8 +167,8 @@ export const relatedTo: [AnOption, AnOption[]][] = [
   ["declarationDir", ["declaration"]],
   ["emitDeclarationOnly", ["declaration"]],
 
-  ["module", ["moduleResolution"]],
-  ["moduleResolution", ["module", "customConditions", "resolvePackageJsonExports", "resolvePackageJsonImports"]],
+  ["module", ["moduleResolution", "esModuleInterop", "allowImportingTsExtensions", "allowArbitraryExtensions", "resolveJsonModule"]],
+  ["moduleResolution", ["module", "paths", "baseUrl", "rootDirs", "moduleSuffixes", "customConditions", "resolvePackageJsonExports", "resolvePackageJsonImports"]],
   ["customConditions", ["moduleResolution", "resolvePackageJsonExports", "resolvePackageJsonImports"]],
   ["resolvePackageJsonExports", ["moduleResolution", "customConditions", "resolvePackageJsonImports"]],
   ["resolvePackageJsonImports", ["moduleResolution", "customConditions", "resolvePackageJsonExports"]],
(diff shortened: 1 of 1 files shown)


agilgur5:
Alright think this is my last PR for the time being 😅

jakebailey:
Hm, by this argument, I'm pretty sure there are loads of other options that affect module resolution, like suffixes, paths, json modules, etc; where do we draw the line?

agilgur5:
It's a good question, I thought about that too. I'm not opposed to adding more `relatedTo` actually.
These 3 just are _particularly_ relevant in the `node16`+ world.

I actually didn't know these options existed -- and thought that changing `moduleResolution` was the only way to get these to work (one of the reasons why we prioritized https://github.com/ezolenko/rollup-plugin-typescript2/pull/453) -- until I read through the reference more recently.

In a similar line of thought, could argue that these 3 specifically are impacted by `moduleResolution`'s setting, whereas the others are not necessarily impacted (or not so directly).
Again though, I'd be fine adding more `relatedTo` because they are all relevant for different use cases; `node16`+ is just _more_ relevant to a wide audience (the vast majority of TS users, I would guess) and is timely right now

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -167,8 +167,11 @@ export const relatedTo: [AnOption, AnOption[]][] = [
   ["declarationDir", ["declaration"]],
   ["emitDeclarationOnly", ["declaration"]],
 
-  ["moduleResolution", ["module"]],
   ["module", ["moduleResolution"]],
+  ["moduleResolution", ["module", "customConditions", "resolvePackageJsonExports", "resolvePackageJsonImports"]],
```
andrewbranch:
```suggestion
  ["moduleResolution", ["module", "paths", "baseUrl", "rootDirs", "moduleSuffixes", "customConditions", "resolvePackageJsonExports", "resolvePackageJsonImports"]],
```

packages/tsconfig-reference/scripts/tsconfigRules.ts
```diff
@@ -167,8 +167,11 @@ export const relatedTo: [AnOption, AnOption[]][] = [
   ["declarationDir", ["declaration"]],
   ["emitDeclarationOnly", ["declaration"]],
 
-  ["moduleResolution", ["module"]],
   ["module", ["moduleResolution"]],
```
andrewbranch:
```suggestion
  ["module", ["moduleResolution", "esModuleInterop", "allowImportingTsExtensions", "allowArbitraryExtensions", "resolveJsonModule"]],
```

andrewbranch (approved):
I think the completeness of these lists is bound to be a little arbitrary, but I like the idea of increasing the discoverability of these lesser-known related options. I’ve added my own suggestions to the list, as well as some to `module`.

PR #2899, opened 2023-07-21
chore: update tslib

Zamiell:
this PR updates the tslib dep to latest in the package.json file and yarn.lock, which fixes `yarn bootstrap` failing on node 18.

the other changes are simply a result of performing the `yarn bootstrap` command, so I can take them out of the PR if you want.


jakebailey:
Pulled out the other ones into #2900; this is an ongoing problem, unfortunately.

Zamiell:
do you wanna rebase this on main for me? not sure what to do specifically

packages/documentation/scripts/types/AllFilenames.d.ts
```diff
@@ -57,8 +57,8 @@ export type AllDocsPages =
   | "reference/Advanced Types.md"
   | "reference/Declaration Merging.md"
   | "reference/Decorators.md"
-  | "reference/ESM Support for Node.md"
   | "reference/Enums.md"
+  | "reference/ESM Support for Node.md"
```
jakebailey: I honestly have no idea why these files are still changed after I merged #2900. Probably best to drop them from the PR and limit this to the package.json and lockfile updates, although my impression is that even if we fix this one thing, you get an even more cryptic issue later in non-node-16 anyway so...

jakebailey:
> do you wanna rebase this on main for me? not sure what to do specifically

There's no need for that; Github figures it out! other than the... other changes? No idea.

Zamiell:
ok i did some manual edits and it should be good now

PR #2900, opened 2023-07-21
Update files modified by bootstrap

jakebailey:
These files are modified in bootstrap because some of their source files changed. We have no mechanism to detect this at the moment so here's a fixup.


orta:
A lot of these were intentionally made to be marked as changed in git (so that it was easy to know all of the changes on an TypeScript version update) but I'm pretty sure they can be safely gitignored if that's more annoying than useful

jakebailey:
I don't mind them being committed per se, but the problem is that the CI doesn't actually error when bootstrap changes files and they aren't committed. That and I've frequently had issues with bootstrap in general not making changes unless I fully `git clean`'d and re-bootstrapped, so if that happens, it's a little frustrating for PR senders to fix up. I don't have a great solution here.

PR #2901, opened 2023-07-22
Adds some infra to warn on files which changed in the PR but aren't accounted for

orta:
Re: https://github.com/microsoft/TypeScript-Website/pull/2900

I'm not entirely sure about the fully cleaned bootstrap issues (it was only really meant to be ran once to download sets of files, and not really be needed to run again except for new ts versions) - so I've moved the one thing which definitely was only triggering in bootstrap only from the most recent PR into the build flow, its a tiny codegen and seems reasonable. I think most of the changes from the recent PR came from OSS folks not running the build but just making direct changes.

This PR switches Danger to run on all PRs and removes my old infra for doing static web app deploys ( https://github.com/Azure/static-web-apps/issues/1#issuecomment-675757849 ) using the same code as we use in DT. It won't fail the build, but it will list what files have changed locally at the end of the run and that could be a good indicator of what's going on.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,29 +40,14 @@ jobs:
         env:
           CI: true
 
-      # Deploy preview environment for non-forks
-      - uses: Azure/static-web-apps-deploy@v1
-        if: github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
-        with:
-          azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN_PROD }}
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          action: upload
-          app_location: site
-          skip_app_build: true
-
-      # Upload site artifact for forks so it can be deployed by a maintainer on-demand
-      - uses: actions/upload-artifact@v3
-        if: github.event.pull_request.base.repo.id != github.event.pull_request.head.repo.id
-        with:
-          name: site-${{ github.event.number }}
-          path: site/
-
       # danger for PR builds
-      - if: github.event_name == 'pull_request' && github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
-        run: "yarn danger ci"
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_DEPLOY_URL_ROOT: ${{ steps.deploy.outputs.static_web_app_url }}
+      - name: "Run Danger"
+        run: |
+          # Exposing this token is safe because the user of it has no other public repositories
+          # and has no permission to modify this repository. See DefinitelyTyped #62638 for the discussion.
+          TOKEN='ghp_i5wtj1l2AbpFv3OU96w6R'
+          TOKEN+='On3bHOkcV2AmVY6'
+          DANGER_GITHUB_API_TOKEN=$TOKEN yarn danger ci
 
   windows:
     if: github.event.action != 'closed'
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -40,29 +40,14 @@ jobs:
...
-        with:
-          azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN_PROD }}
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          action: upload
-          app_location: site
-          skip_app_build: true
-
-      # Upload site artifact for forks so it can be deployed by a maintainer on-demand
-      - uses: actions/upload-artifact@v3
-        if: github.event.pull_request.base.repo.id != github.event.pull_request.head.repo.id
```
jakebailey: Er, is this not the code which powers the `deploy-preview` label? https://github.com/microsoft/TypeScript-Website/blob/b03935a55750acfadc496e5fdf4258730c96cb5f/.github/workflows/deploy-preview.yml

Probably, just the deploy is what needs to be deleted?

orta: Oh yeah, I thought it was doing a more native version of it, will drop next time I get back to the computer

orta:
This got cleaned up 👍🏻

.github/workflows/CI.yml
```diff
@@ -58,11 +58,13 @@ jobs:
...
-      - if: github.event_name == 'pull_request' && github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
-        run: "yarn danger ci"
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_DEPLOY_URL_ROOT: ${{ steps.deploy.outputs.static_web_app_url }}
+      - name: "Run Danger"
+        run: |
+          # Exposing this token is safe because the user of it has no other public repositories
+          # and has no permission to modify this repository. See DefinitelyTyped #62638 for the discussion.
+          TOKEN='ghp_i5wtj1l2AbpFv3OU96w6R'
```
rbuckton, line 65: Even after reading through the comment thread at https://github.com/Definitelytyped/DefinitelyTyped/issues/62638, I'm still not a fan of using an unprotected token. Making this token public means anyone could use it to attempt to DDoS GitHub and have it falsely attributed to the DangerBotOSS account, or DoS Danger by artificially using up its rate limit. If making this a secret is not viable for usability reasons, are there any other mechanisms that could be employed to avoid exposing the token?

dangerfile.ts
```diff
@@ -2,9 +2,16 @@
 // yarn danger pr https://github.com/microsoft/TypeScript-Website/pull/115
 
 import spellcheck from "danger-plugin-spellcheck"
-
-// Blocked on PR deploys, see CI.yml
-// import lighthouse from "danger-plugin-lighthouse"
+import { warn, danger }from "danger"
```
rbuckton: nit: whitespace
```suggestion
import { warn, danger } from "danger"
```

rbuckton (changes requested):

.github/workflows/CI.yml
jakebailey: I'm not really sure why this needs a token at all, either; isn't this just a CI check that can fail and print out the files that were forgotten? a la https://github.com/microsoft/TypeScript/blob/main/.github/workflows/ci.yml#L239

I guess because this tries to print a fancy comment?

orta: It's because you need a token to make a comment on an issue, yes!

image

Yep, it's a comment because no-one would read a non-failing CI build for things like warnings

I did explore having a central github app https://github.com/danger/danger-js/pull/1126 but it requires giving too much github access to the bot IMO, and I didn't want to centralize that many people's tokens on my spare time

jakebailey:
Done via another PR.

PR #2907, opened 2023-07-27
Use CTRL+A to select output code

meowtec:


packages/playground/src/ds/createDesignSystem.ts
```diff
@@ -159,6 +159,16 @@ export const createDesignSystem = (sandbox: Sandbox) => {
       createCodePre.appendChild(codeElement)
       container.appendChild(createCodePre)
 
+      // When <pre> focused, Ctrl+A should select only code pre instead of the entire document
+      createCodePre.addEventListener('keydown', e => {
```
jakebailey:
```suggestion
      createCodePre.addEventListener("keydown", e => {
```

packages/playground/src/ds/createDesignSystem.ts
```diff
@@ -159,6 +159,16 @@ export const createDesignSystem = (sandbox: Sandbox) => {
...
+        if (e.key.toUpperCase() === 'A' && (e.ctrlKey || e.metaKey)) {
```
jakebailey:
```suggestion
        if (e.key.toUpperCase() === "A" && (e.ctrlKey || e.metaKey)) {
```

jakebailey (approved):
Thanks; the preview seems to work, including the d.ts output as well.

PR #2911, opened 2023-07-30
Update Enums.md: Fix wrong description

green961:
in this section
[Union enums and enum member types](https://www.typescriptlang.org/docs/handbook/enums.html#union-enums-and-enum-member-types)

```ts
enum E {
  Foo,
  Bar,
}

function f(x: E) {
  if (x !== E.Foo || x !== E.Bar) {
```

In that example, we first checked whether x was not E.Foo. If that check succeeds, then our || will short-circuit, and the body of the ‘if’ will run. However, if the check didn’t succeed, `then x can only be E.Bar`, so it doesn’t make sense to see whether it’s equal to E.Bar.

when the first check didn’t succeed, then x is `E.Foo`, so it doesn’t make sense to see `x !== E.Bar`

### DIFF
--- a/packages/documentation/copy/en/reference/Enums.md
+++ b/packages/documentation/copy/en/reference/Enums.md
@@ -220,7 +220,7 @@ function f(x: E) {
 
 In that example, we first checked whether `x` was _not_ `E.Foo`.
 If that check succeeds, then our `||` will short-circuit, and the body of the 'if' will run.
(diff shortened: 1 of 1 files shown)


0xScratch (approved):
Now that's correct!!
But do provide some description and a valid 'heading' or 'commit line' for pull requests, easy for others to review

jakebailey:
Presumably, #2868 was in fact wrong?

jakebailey (approved):
Yeah, I think I misread the other PR, as did the other author.

PR #2912, opened 2023-07-31
Update Generics.md: Some Grammar checks

0xScratch:
Made changes that are kind of pretty clear and understood by the users, and they suit grammatically well!


jakebailey (changes requested):
Thanks for the PR, but I don't think this is any clearer than the existing wording; it's effectively the same.

PR #2921, opened 2023-08-19
Update Everyday Types.md

usmajeed:
I think the solution of making "as const" apply to .req only instead of the object makes more sense given we are focusing on .req in the context. We also avoid enforcing literal type on "https://example.com" which is unintended (potentially).


andrewbranch:
It kind of feels like the `"GET" as "GET"` example should just be deleted in favor of `"GET" as const`. @jakebailey what do you think?

jakebailey:
Yeah, though see also #2924.

andrewbranch:
I don’t love the copy editing in either PR, to be honest. I agree the code change is better code, but I don’t think that makes a material difference on the lesson being explained. At the moment, I would rather keep the existing content.

jakebailey:
I agree.

PR #2922, opened 2023-08-20
remove deprecated link

ymoukhli:
when going to v1 page a get a message telling that v1 is deprecated and i should return back to v2 handbook page


zamazzal (commented):
May it can be used by someone who have an old version.

jakebailey (approved):

PR #2924, opened 2023-08-24
Update Everyday Types.md for literal type work around

ferhadquluzade:
Work around for specific parameters in case of literal types


jakebailey:
https://github.com/microsoft/TypeScript-Website/pull/2921#issuecomment-1692540379

PR #2928, opened 2023-08-28
Fix DOM element check

HubertY:
Fix the fact that checking for the presence of DOM id/element in playground settings config always succeeds due to hitting the default value; it should check user-provided config only.


packages/sandbox/src/index.ts
```diff
@@ -144,7 +144,10 @@ export const createTypeScriptSandbox = (
 
   const language = languageType(config)
   const filePath = createFileUri(config, compilerOptions, monaco)
-  const element = (config as any).elementToAppend ? (config as any).elementToAppend : document.getElementById(config.domID)
+  const element = "elementToAppend" in config ? config.elementToAppend : document.getElementById(config.domID)
```
jakebailey, line 147: Why the switch to `in` here?

HubertY: I think its nicer to leverage the automatic type narrowing; I had to stop and figure out why the cast to any was necessary and whether it implied anything was wrong. A third cast to any is avoided here only because Typescript has inferred that `config.domID` is always at least an empty string due to the default value.

If the other way is preferred it can be changed back.

Relatedly, if the ternary was flipped to `(config as any).domID ? ...`  the logic would be wrong. The same applies to my version of course, but I think the potential for error is more obvious there. Specifically, we are explicitly checking for the _presence_ of `elementToAppend`. We could not mistake that we could rely on the truthiness or falsiness of `domID`'s default value.

jakebailey (approved):
Seems good to me, then, though I'll probably wait to merge this until I have some time to do all of the shenanigans to bump versions so the release happens.

HubertY:

PR #2943, opened 2023-09-13
DOM Manipulation Clarification

robhowe-A:
Adding parentheses clarifies statements.


jakebailey:
I'm not sure that this is an improvement; it reads roughly the same to me.

robhowe-A:
Going through the doc, I read it as ambiguous/imprecise. Id be happy to help the change, of it helps.
DOM manipulations are fairly easy, there should be nothing ambiguous at that level of interpretation. I learned much from the article.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

jakebailey:
I'm not sure what we'd change it too; both before and after it refers to a code snippet above and I don't see how it can be ambigious.

robhowe-A:
The proper grammatical implementation is to use parentheses, I'm sure I looked it up to confirm it. That's why I'm offering an improvement.

Proper scripts documents deserve proper use of grammer to avoid amiguous situations from any sense. A clearer version may be helpful to someone, so I submitted the version change.


All the best,

Robert H.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

robhowe-A:
Pushing requires testing. That appears to be without my attention?

PR #2944, opened 2023-09-18
Add link to definition for 'solution' in Project References

mattpocock:


jakebailey:
As much as this link was good in the moment to explain what "solution" meant, we definitely need to have a full explainer of this whole thing in the docs, so I'm a little bit apprehensive about merging this lest people try and use the VS docs to explain anything 😅

PR #2945, opened 2023-09-21
Changing ```ts twoslash``` to ```ts```

navya9singh:
Changing ```ts twoslash``` to ```ts``` in places without any errors or cuts


orta:
( to make sure you're aware of the consequences for these changes, this would remove all of the hover states from these code samples too, which is a bit of a shame and you might need to look at the text surrounding the code samples to make sure there are no assumptions you can do this )

(( if you are doing this to try save on speed for CI - I'd look into whether the caching for the results of the run is not working for example ))

((( and I guess you might want to check for twoslash commands like `// ^?` or `@compilerFlag` too )))

jakebailey:
The main problem is the build speed, yes; removing these comments at least halves the build time. If there's supposed to be caching, it definitely is not working. IIRC we build from scratch at each commit. And I honestly don't get a working website build locally without cleaning anyway :(

orta:
Weird! This is the code which handles caching and returning the twoslash call: https://github.com/shikijs/twoslash/blob/main/packages/remark-shiki-twoslash/src/caching.ts#L4-L68

Maybe there are issues with it running on windows? This is running on CI for windows here, but maybe the cache lookup always fails?

It caches the results of the files into `node_modules/.cache/twoslash` - so check to see if they are in there at all, or maybe something is getting corrupted?

jakebailey:
Well, in CI, we're not going to preserve that directory. But otherwise locally, I still have to clean build quite a bit to avoid broken caching in other places, so I'm sure I'm removing that one.

orta:
You don't preserve node modules in CI? That's is(/was?) default in GitHub actions for setup-node so I figured that was a safe assumption as that pattern is pretty well used in other npm modules in the ecosystem also - maybe that folder was the removed for this repos CI  cache at some point?

For you locally, hard to say offhand - what commands are you running to clean the build?

jakebailey:
> You don't preserve node modules in CI? That's is(/was?) default in GitHub actions for setup-node so I figured that was a safe assumption as that pattern is pretty well used in other npm modules in the ecosystem also - maybe that folder was the removed for this repos CI cache at some point?

All CI caching mechanisms I know save the npm/yarn/pnpm cache only; saving `node_modules` itself is very error prone as it can contain symlinks and post-install artifacts that are machine or path specific. `setup-node`'s caching (when enabled, it's not enabled by default) only saves a single package manager's cache.

> For you locally, hard to say offhand - what commands are you running to clean the build?

I most definitely `git clean -fdx` and do a full reinstall and rebuild. It doesn't get cleaner than that, but it's the only way I can be sure that a local website build will contain the updated releases files, styles (we hit this bug recently), etc. We've wasted days debugging accessibility fixes only to find out that caching was busted and a full clean, reinstall, and rebuild fixed everything...

I of course expect no caching at that point, but I certainly don't generally expect a website build to take 4 minutes on my local machine or 20 in CI...

orta:
That's all pretty surprising yep! It wasn't really like that when I was working on it, happy to help debug some of those and/or move the twoslash caching to work with how this repo is caching between CI builds too

I probably won't have much time over over September, but I'll see if I can get a repro for some of those locally

PR #2946, opened 2023-09-22
New modules docs

### LINKED ISSUE #52593: Module documentation tracking issue
I’m working on new documentation for modules in TypeScript. I’ll use this issue to track progress. Related:

- #51876
- #52412
- #51267

andrewbranch:
Closes https://github.com/microsoft/TypeScript/issues/52593


packages/documentation/copy/en/modules-reference/Theory.md
```diff
@@ -0,0 +1,415 @@
...
+  subgraph Input files
+    src/main.mts
+    src/math.mts
+  end
+  src/main.mts -. Map to output .-> dist/main.mjs
+  dist/main.mjs -- "#quot;./math.mjs#quot;" --> dist/math.mjs
+  dist/math.mjs -. Map to input .-> src/math.mts
+```
+
+Understanding this mental model may not immediately eliminate the strangeness of seeing output file extensions in input files, and it’s natural to think in terms of shortcuts: “`"./math.mjs"` refers to the input file `math.mts`. I have to write the output extension, but the compiler knows to look for `.mts` when I write `.mjs`.” This shortcut is even how the compiler works internally, but the more robust mental model explains _why_ module resolution in TypeScript works this way: given the constraint that the module specifier in the output file will be [the same](#module-specifiers-are-not-transformed) as the module specifier in the input file, this is the only process that accomplishes our two goals of validating output files and assigning types.
```
orta: This sentence felt a bit iffy because of the intermixed double quotes, it could be solved by italicising?

```suggestion
Understanding this mental model may not immediately eliminate the strangeness of seeing output file extensions in input files, and it’s natural to think in terms of shortcuts: _“`"./math.mjs"` refers to the input file `math.mts`. I have to write the output extension, but the compiler knows to look for `.mts` when I write `.mjs`.”_ This shortcut is even how the compiler works internally, but the more robust mental model explains _why_ module resolution in TypeScript works this way: given the constraint that the module specifier in the output file will be [the same](#module-specifiers-are-not-transformed) as the module specifier in the input file, this is the only process that accomplishes our two goals of validating output files and assigning types.
```

packages/documentation/copy/en/modules-reference/Introduction.md
```diff
@@ -0,0 +1,14 @@
...
+title: Introduction
+layout: docs
+permalink: /docs/handbook/modules/introduction.html
+oneline: How TypeScript models JavaScript modules
+translatable: true
+---
+
+This document is divided into four sections:
+
+1. The first section develops the [**theory**](/docs/handbook/modules/theory.html) behind how TypeScript approaches modules. If you want to be able to write the correct module-related tsconfig options for any situation, reason about how to integrate TypeScript with other tools, or understand how TypeScript processes dependency packages, this is the place to start. While there are guides and reference pages on these topics, building an understanding of these fundamentals will make reading the guides easier, and give you a mental framework for dealing with real-world problems not specifically covered here.
```
orta: WRT the rest of the docs, I used `TSConfig` or `tsconfig.json` depending on what contextually makes sense, if folks overall prefer it all in lowercase that's also fine too but you might want to consider syncing that around

packages/documentation/copy/en/modules-reference/Theory.md
```diff
@@ -0,0 +1,415 @@
...
+    <script src="a.js"></script>
+    <script src="b.js"></script>
+  </head>
+  <body></body>
+</html>
+```
+
+This approach had some downsides, especially as web pages grew larger and more complex. In particular, all scripts loaded onto the same page share the same scope—appropriately called the “global scope”—meaning the scripts had to be very careful not to overwrite each others’ variables and functions.
+
+Any system that solves this problem by giving files their own scope while still providing a way to make bits of code available to other files can be called a “module system.” (It may sound obvious to say that each file in a module system is called a “module,” but the term is often used to contrast with _script_ files, which run outside a module system, in a global scope.)
```
orta, line 26: The opening line for this paragraph feels a little unwieldy (lots of abstract terms) when you try say it out loud could try starting with the term and then building backwards?

```suggestion
We call a system which tries to break up that single global scope a “module system“, these are usually build around providing a per-file scope which allows for ways to pass bits of code between each other. (It may sound obvious to say that each file in a module system is called a “module,” but the term is often used to contrast with _script_ files, which run outside a module system, in a global scope.)
```

packages/documentation/copy/en/modules-reference/Theory.md
```diff
@@ -0,0 +1,415 @@
...
+```ts
+export * from "./utils";
+```
+
+Assuming `./utils.ts` (or `./utils/index.ts`) exists, a bundler would be fine with this code, so `"moduleResolution": "bundler"` doesn’t complain. Compiled with `"module": "esnext"`, the output JavaScript for this export statement will look exactly the same as the input. If that JavaScript were published to npm, it would be usable by projects that use a bundler, but it would cause an error when run in Node.js:
+
+```
+Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/dependency/utils' imported from .../node_modules/dependency/index.js
+Did you mean to import ./utils.js?
+```
```
orta, line 355: You have the ability to [inline custom errors](https://shikijs.github.io/twoslash/playground?#code/MYewdgziA2CmB0sBOSRIBQCIBysCWAFgAQEgCumAlAFAD0tRAAsqkgFxG6EnlA) if you want

orta (commented):
This PR deletes a bunch of markdown files which will break URLs people have been using, you could consider adding them back and adding the `deprecated` YML tag and there's a tag you can use to forward someone to the replacement also. It should handle the HTML tags to tell search engines to lower the rank on those pages too

Only looked at the theory page, but got 1-2 comments

andrewbranch:
> This PR deletes a bunch of markdown files which will break URLs people have been using, you could consider adding them back and adding the deprecated YML tag and there's a tag you can use to forward someone to the replacement also

I added redirects for the deleted pages in `setupRedirects`.

PR #2947, opened 2023-09-23
Rephrase thing to identifier

prajwalkulkarni:
While defining type annotations to entities like variables, functions, etc., I feel it makes much more sense to use the terminology **identifier** instead of simply calling it a *thing*.


prajwalkulkarni:
Hey @jakebailey, what are your thoughts on this?

jakebailey:
I personally don't think that it's an improvement as-is; this page is mainly a page for beginners and doesn't use the word "identifier" at all except in one error message. It's also inaccurate for destructurings like `const { foo } = ...`, because a colon after the identifier renames the variable, it doesn't type it.

I think this sentence is mainly trying to indicate the "general scheme" of TS's typing where the type annotations for everything go after, i.e. return types too, and not be too specific about one particular syntax.

PR #2950, opened 2023-09-30
Update tsconfig.json.md

xn1cklas:
Since the tsconfig can also extend frameworks besides runtime environments, I wanted to highlight that option to clarify the configuration process.

### DIFF
--- a/packages/documentation/copy/en/project-config/tsconfig.json.md
+++ b/packages/documentation/copy/en/project-config/tsconfig.json.md
@@ -92,6 +92,40 @@ For example, if you were writing a project which uses Node.js version 12 and abo
 }
 ```
 
+Besides runtime environment, some frameworks also require an extension from their base configuration. Those are also listed at [github.com/tsconfig/bases](https://github.com/tsconfig/bases/). 
+For example Next.js:
+
+```json tsconfig
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Next.js",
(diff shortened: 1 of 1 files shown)


xn1cklas:
@microsoft-github-policy-service agree [company="Nicklas Scharpff"]

xn1cklas:

orta:
I don't think it's a good idea to put the whole JSON for one of these tsconfigs in these docs - it's going to get outdated pretty quickly as these frameworks move fast, and it's not the way in which [next recommends](https://nextjs.org/learn/excel/typescript/create-tsconfig) setting up a tsconfig making it slightly misleading

jakebailey:
Yeah, I think the existing link suffices.

That and the following paragraph gets its flow broken because it now has a large next-specific block before it.

xn1cklas:
I agree to both of you. Yet I would like to clarify that the config does not only extend a runtime environment, but also frameworks / other technologies.

Am 30.09.2023 um 16:18 schrieb Jake Bailey ***@***.***>:

﻿

Yeah, I think the existing link suffices.

That and the following paragraph gets its flow broken because it now has a large next-specific block before it.

jakebailey:
I'm going to close this; even if it's true that some frameworks would have a base on the bases repo, it's certainly the case that next themselves already set up the tsconfig for you, so nobody should be writing it by hand (https://nextjs.org/learn/excel/typescript/create-tsconfig)

PR #2955, opened 2023-10-18
Fix satisfies example in docs

nickbarrett645:
The code example for the `satisfies`  keyword indicates that there would be an error related to 
`const redComponent = palette.red.at(0);` when the type is added to the palette object because red could be type RGB (Array) or a string. However, there wouldn't be an error there because the function `at` is on both Array and String prototypes. There would however be an error on the next line in the example because the `toUpperCase` function does only apply to Strings.

### DIFF
--- a/packages/documentation/copy/en/release-notes/TypeScript 4.9.md
+++ b/packages/documentation/copy/en/release-notes/TypeScript 4.9.md
@@ -14,17 +14,17 @@ For example:
 ```ts
 // Each property can be a string or an RGB tuple.
 const palette = {
-    red: [255, 0, 0],
-    green: "#00ff00",
+    red: "#ff0000",
+    green: [0, 255, 0],
     bleu: [0, 0, 255]
 //  ^^^^ sacrebleu - we've made a typo!
 };
 
-// We want to be able to use array methods on 'red'...
-const redComponent = palette.red.at(0);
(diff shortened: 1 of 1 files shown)


orta:
This example could use twoslash being turned on to show the errors (and hovers) but the original looks correct to me still as its talking about the lack of type refinement due to `as const` not being applied (because it is `string | RBG` in the type system)

nickbarrett645:
Here is what I am seeing when I implement the example from the docs:
Screenshot 2023-10-18 at 9 35 16 AM

The error is not on `const redComponent = palette.red.at(0)` because the function `at` is applicable to both strings and arrays. The type of `palette` defines the value in the record to be either a string or an RGB type, which is an array, so either way the `at` function will be available on whatever value exists. We do see an error on `const greenNormalized = palette.green.toUpperCase()` because toUpperCase only applies to strings and not arrays. There for that function may not be available depending on the type of the value.

 The docs originally say
```
// But we now have an undesirable error here - 'palette.red' "could" be a string.
const redComponent = palette.red.at(0);
```
This is not correct because in this example is does not matter if the value of red in the palette object is a string or an array because `at` exists on either type.

The change is just to clarify which line the error is seen on.

jakebailey:
Thanks for sending this, but I merged #3050 instead as I felt that was clearer.

PR #2958, opened 2023-10-18
Add docs for using/await using declarations

### LINKED ISSUE #55001: `await using` hides potential unhandled promises.
# Bug Report

See explanation below the code.

### 🔎 Search Terms

await using  disposable

5.2 beta

### 💻 Code

```
(Symbol as any).dispose ??= Symbol("Symbol.dispose");
(Symbol as any).asyncDispose ??= Symbol("Symbol.asyncDispose");

import {promisify} from "util";
const delay = promisify(setTimeout);

class Resource {
  constructor(public readonly path: string) {}
  async [Symbol.asyncDispose]() {
    await delay(500);
    console.log('closed ' + this.path);
  }
}

class Boom extends Error {
  constructor() {
    super("Boom!");
  }
}

async function main() {
  await using file = await new Resource("thefile.txt");
  console.log(file.path);
  return Promise.reject(new Boom());
}

try {
  await main();
} catch (err) {
  console.log(`All good. Caught ${err.message}`);
}
```

### 🙁 Actual behavior

```
Boom [Error]: Boom!
    at main (file:///Users/luciano/git/ts-using/lib/index.mjs:70:31)
    at async file:///Users/luciano/git/ts-using/lib/index.mjs:83:5
```

The boom error is thrown due to being unhandled.

### Explanation

My understanding is that this is happening due to `return Promise.reject(new Boom());` returning a rejected promise, and the delay (with `setTimeout`) in the disposable function. The delay makes the rejected promise be seen and thrown.

It would be similar if the delay were in a `finally` section. However, the disposable function could be in an object within a library and thus not easily seen by users.

### 🙂 Expected behavior

```
thefile.txt
closed thefile.txt
All good. Caught Boom!
```

### Caveat

The boom error should have been caught.

It could be that this isn't a bug, but working as expected. If so, let's consider making it clear in the documentation (e.g. use `throw` or `return await Promise.reject()`

rbuckton:
Adds documentation regarding `using` and `await using` declarations, as well as caveat suggested in https://github.com/microsoft/TypeScript/issues/55001#issuecomment-1640060943

Fixes microsoft/TypeScript#55001


packages/documentation/copy/en/reference/Variable Declarations.md
```diff
@@ -704,3 +704,108 @@ clone.m(); // error!
...
+`await using` declaration:
+
+```ts
+async function f() {
+  await using x = new C();
+} // `await x[Symbol.asyncDispose]()` is invoked
+```
+
+An `await using` declaration invokes, and _awaits_, its value's `[Symbol.asyncDispose]()` method as control leaves the containing
+block. This allows for asynchronous cleanup, such as a database transaction performing a rollback or commit.
```
jakebailey: Should there be an example with an async right side, just to help dismambiguate that `await using x = foo()` does not await the result of `foo`?

jakebailey (approved):

packages/documentation/copy/en/reference/Variable Declarations.md
rbuckton: I can look into expanding this. I'm not sure how much detail we want to add here. I could also cover how `using` and `await using` can be used in `for` and `for..of`, but worried that might be too much in the weeds.

andrewbranch (approved):
Feel free to merge whenever. It will deploy right away.

rbuckton:
I've added a few more sections to incorporate Jake's feedback and describe how to define your own resources.

rbuckton:
I feel like I should move the `using` and `await using` sections after the "let vs. const" section to keep things logically grouped together.

rbuckton:
Any further feedback on the updated text? If not, I'll merge.

jakebailey (approved):

PR #2959, opened 2023-10-19
Remove color customizations from .vscode/settings.json

rbuckton:
These customizations only apply to the VSCode editor when you open the repo in the editor and can conflict with your chosen color theme in VSCode, reducing visibility and making things harder to use.


jakebailey (approved):

andrewbranch (approved):
Thank you, I’ve wanted to do this for a long time

PR #2965, opened 2023-10-23
Move localized fonts to newer CDN

### LINKED ISSUE #2964: MINOR: Segoeui font loading from different CDN host
The TypeScript playground has an issue loading when sharepointonline.com is blocked by VPNs (that are trying to cut off data sharing channels).

This appears to be due to the following two requests for fonts:
https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-regular.woff
https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-semibold.woff

If these fonts could be served from the same CDN as the main website that would be much appreciated.

jakebailey:
Fixes #2964

These fonts exist for localizations (which are no longer supported), but since we haven't removed that stuff, it's easier to just update these files to point to the more current CDN used in https://github.com/microsoft/fluentui/blob/4c36df0e2a7bf9d54d5137fabe74c720bdddebbc/packages/theme/src/fonts/DefaultFontStyles.ts#L60.

Perhaps a better method would be to just remove all of this anyway and let the user's system provide the right fonts for their locale, but I am a lot less certain of my ability to do that.

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/fonts.scss
+++ b/packages/typescriptlang-org/src/components/layout/fonts.scss
@@ -67,13 +67,13 @@
   font-display: swap;
 }
 
-// Fluent App Fonts, taken from the source code from the fluent gatsby template
+// https://github.com/microsoft/fluentui/blob/4c36df0e2a7bf9d54d5137fabe74c720bdddebbc/packages/theme/src/fonts/DefaultFontStyles.ts#L60
 
 @font-face {
   font-family: "Leelawadee UI Web";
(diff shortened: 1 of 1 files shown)


jakebailey:
@acutmore I am not sure if we can rehost these assets, but is this new CDN blocked by your firewall? I would assume you don't block `cdn.office.net` but if you do, then this fix won't be enough.

RyanCavanaugh (approved):

acutmore:
> but is this new CDN blocked by your firewall?

These URLs seem good, thank you!

PR #2966, opened 2023-10-25
Update Conditional Types.md to cover unions impacting conditional type evaluation.

CraigMacomber:
Add except about unions impacting conditional type evaluation.

Without this clarification this page is particularly confusing since the example relies on a behavior that differs from how the explication claims they work.

I considered trying to refactor the document to fit this in more naturally, but to me this behavior of TypeScript doesn't fit naturally into a justification of why conditional types are useful, so I failed to incorporate it more inline.

I also considered including it as a separate section, but it really seems like it should immediately follow the statement which is oversimplifying things to avoid people seeing that and incorrectly assuming that simplified behavior applies in all cases when it does not apply to unions, any, boolean (or probably enums).

I looked for a definition in the docs for what types behave in this union like way, but didn't find any good existing terms for these types: if there is one, my section should be updated to use (and maybe link) it.

This is also my first time working with the playground powered code blocks in markdown, and the preview feature on git-hub just hung, so its possible I got the formatting wrong: pay extra attention to that in review as I don't know how to test it.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
@@ -38,6 +38,46 @@ type Stuff =
 
 When the type on the left of the `extends` is assignable to the one on the right, then you'll get the type in the first branch (the "true" branch); otherwise you'll get the type in the latter branch (the "false" branch).
 
+> Be aware that this is not true for union types.
+> When the type on the left of the `extends` is a union, each of its members are evaluated for the `extends` clause independently and the results are unioned together.
+> The example below depends on this when it evaluates `number | string extends number`.
+> This is false (`number | string` is not assignable to `number`) however this type expression evaluates to the union `trueExpression | falseExpression` instead of just `falseExpression`.
+> This also applies to types that are implicitly treated like unions, for examples `boolean`: `boolean extends true` is false, but when used in a conditional type returns the union of the true and false expressions.
+> The type `any` is also handled like a union, always producing the `trueExpression` unioned with the `falseExpression`.
+> 
+> This does not generalize to types which contain unions.
+> For example `{ id: number | string } extends { id: number }` is false, and only produces the false expression.
+> For this to produce the union behaviour, the input type would need to be rewritten to `{ id: number } | { id: string }`.
+> This version of the type corosponds to the exact same set of runtime values as the previous form, but behaves differently when used in conditional types.
+> If you are writing a generic type which is conditional on its input, you may want to document which of these representations it supports or structure it to ensure that both are handled correctly.
+> 
+> In the above case, an example of handleing both in an inconsistant way:
+> ```ts twoslash
+> type UsesNumberId<T extends { id: number | string }> = T extends  { id: number } ? true : false;
+> 
+> type A = UsesNumberId<{ id: number }>;
+> //  ^?
+> type B = UsesNumberId<{ id: string }>;
+> //  ^?
+> type C = UsesNumberId<{ id: number | string }>;
+> //  ^?
+> type D = UsesNumberId<{ id: number } | { id: string }>;
+> //  ^?
+> ```
+> This version handles both formats for its argument the same, making cases C and D both `boolean` unlike the above form:
+> ```ts twoslash
+> type UsesNumberId<T extends { id: number | string }> = T extends  { id: number } ? true : T extends { id: string } ? false : boolean;
+> 
+> type A = UsesNumberId<{ id: number }>;
+> //  ^?
+> type B = UsesNumberId<{ id: string }>;
+> //  ^?
+> type C = UsesNumberId<{ id: number | string }>;
+> //  ^?
+> type D = UsesNumberId<{ id: number } | { id: string }>;
+> //  ^?
+> ```
+
 From the examples above, conditional types might not immediately seem useful - we can tell ourselves whether or not `Dog extends Animal` and pick `number` or `string`!
 But the power of conditional types comes from using them with generics.
 


packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
```diff
@@ -38,6 +38,46 @@ type Stuff =
 
 When the type on the left of the `extends` is assignable to the one on the right, then you'll get the type in the first branch (the "true" branch); otherwise you'll get the type in the latter branch (the "false" branch).
 
+> Be aware that this is not true for union types.
```
RyanCavanaugh, line 41: This is the entire theme of https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types - any discussion of union distributivity belongs there and should be integrated into that section.

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
```diff
@@ -38,6 +38,46 @@ type Stuff =
...
+> When the type on the left of the `extends` is a union, each of its members are evaluated for the `extends` clause independently and the results are unioned together.
+> The example below depends on this when it evaluates `number | string extends number`.
+> This is false (`number | string` is not assignable to `number`) however this type expression evaluates to the union `trueExpression | falseExpression` instead of just `falseExpression`.
+> This also applies to types that are implicitly treated like unions, for examples `boolean`: `boolean extends true` is false, but when used in a conditional type returns the union of the true and false expressions.
```
RyanCavanaugh, line 45: The docs explicitly say

> The type boolean itself is actually just an alias for the union true | false.

So it's not really "implicitly treated as a union"; it *is* a union

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
```diff
@@ -38,6 +38,46 @@ type Stuff =
...
+> The type `any` is also handled like a union, always producing the `trueExpression` unioned with the `falseExpression`.
+> 
+> This does not generalize to types which contain unions.
+> For example `{ id: number | string } extends { id: number }` is false, and only produces the false expression.
+> For this to produce the union behaviour, the input type would need to be rewritten to `{ id: number } | { id: string }`.
+> This version of the type corosponds to the exact same set of runtime values as the previous form, but behaves differently when used in conditional types.
+> If you are writing a generic type which is conditional on its input, you may want to document which of these representations it supports or structure it to ensure that both are handled correctly.
+> 
+> In the above case, an example of handleing both in an inconsistant way:
```
RyanCavanaugh, line 54: This section is confusing; what is the reader thinking the code is trying to do?

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
```diff
@@ -38,6 +38,46 @@ type Stuff =
...
+> The example below depends on this when it evaluates `number | string extends number`.
+> This is false (`number | string` is not assignable to `number`) however this type expression evaluates to the union `trueExpression | falseExpression` instead of just `falseExpression`.
+> This also applies to types that are implicitly treated like unions, for examples `boolean`: `boolean extends true` is false, but when used in a conditional type returns the union of the true and false expressions.
+> The type `any` is also handled like a union, always producing the `trueExpression` unioned with the `falseExpression`.
+> 
+> This does not generalize to types which contain unions.
+> For example `{ id: number | string } extends { id: number }` is false, and only produces the false expression.
+> For this to produce the union behaviour, the input type would need to be rewritten to `{ id: number } | { id: string }`.
+> This version of the type corosponds to the exact same set of runtime values as the previous form, but behaves differently when used in conditional types.
```
RyanCavanaugh, line 51: Treating the special case where distributivity applies as if it were the general case is going to lead to confusion later on. For a generic type that consumes a union at multiple points, for example, these types *aren't* equivalent.

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
CraigMacomber: I missed that, since I read this document until I found something that's incorrect and contradicts the examples it uses. Having a section people don't know they need which contradicts statements in the introduction is a good way to leave people like me confidently misinformed after skimming the top of the document.

To mitigate people in the future making the same mistake, I think a note and a link to that section should be included here (and possible in the example where it depends on that behavior), then I can integrate my examples into that section.

Does that seem like a good approach?

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
CraigMacomber: Thanks for the quote: searching for it allowed me to find where this is documented. 

I had checked the docs for unions and https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean which seems like the official docs for booleans and its not there.

The quote you have is buried many paragraphs into https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types, which isn't really very linkable. Since that design decision impact how the type functions in rather important ways (and makes it work quite differently from strings or numbers, and the design decision to make them different was subjective so it's not something someone can infer from context: they need to know/learn it) it might be better to emphasize that somewhere more obvious like in https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean and link that from here.

Is that a change you think would make sense to include in this PR? Otherwise, I'll just adjust my phrasing here to be more explicit and not link anything.

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
RyanCavanaugh: I don't like interrupting the flow of the Handbook with caveats that you need to read the entire page you're currently reading; the entire Handbook is intended to be consumed linearly and all sections include generalizations that are later fleshed out in more detail.

packages/documentation/copy/en/handbook-v2/Type Manipulation/Conditional Types.md
RyanCavanaugh: The Handbook is intended as a linear document, and follows the [inverted pyramid structure](https://en.wikipedia.org/wiki/Inverted_pyramid_(journalism)) of narrative wherein general information is given first, followed by more specific information later. It's also carefully structured so that a reader that doesn't understand anything about TypeScript can read it linearly, which means we can't reference terms we haven't introduced yet.

What you're describing is more of a reference page format, where all the possible relevant details are spelled out in a topically-organized manner. The reason we don't use that format for the Handbook is that you need semantic grounding for all the terms you use before it can make sense to a reader - if we're introducing the primitive types but haven't talked about union types or literal types yet, saying "boolean is a union of two literal types" is a *conceptual forward reference* to something the reader hasn't learned about yet, thus likely to confuse them.

PR #2969, opened 2023-10-30
Update TypeScript 3.2.md (fibonacci typo)

yairkukielka:
Fixed a typo in the fibonacci example


jakebailey:
This just changes the numbers to be 1 indexed rather than 0 indexed, so I'm not sure it's an actual improvement (and sort of not the point of the example, which is just to show code that blows up).

PR #2975, opened 2023-11-11
feat: using source maps for better stack traces

patricktree:
In the section about `I’m writing a library`, regarding `sourceMap: true` and `declarationMap: true`, the docs currently state:

> These are only useful if the library also ships its source (`.ts`) files.

IMO source maps for the output JavaScript files can also be useful for _better stack traces_.  
I added a sentence for that and removed the hint quoted above (because for better stack traces the source maps alone are sufficient; the source files don't need to be distributed also).


jakebailey (changes requested):
I don't mind mentioning this, but the current wording seems to be pretty repetitive; it's three sentences that say "By shipping ...".

patricktree:
@jakebailey I improved the text a little bit!

packages/documentation/copy/en/modules-reference/guides/Choosing Compiler Options.md
```diff
@@ -184,7 +184,9 @@ Let’s examine why we picked each of these settings:
...
+- **`sourceMap: true`** emits source maps for the output JavaScript files. By shipping source maps, when consumers of the library execute the code with source maps enabled (e.g. via [`node --enable-source-maps`](https://nodejs.org/docs/latest-v18.x/api/cli.html#--enable-source-maps)), stack traces of errors will point to the original source TypeScript files (instead of the emitted output JavaScript files). When the source files are also shipped with the library, source maps allow consumers of the library to debug the library code somewhat more easily.
+- **`declarationMap: true`** emits source maps for the output type declaration files. By shipping declaration maps and source files, consumers will be able to see the original TypeScript sources when they run Go To Definition on imports from the libraries.
+- Note that **`sourceMap: true`** and **`declarationMap: true`** represent a tradeoff between developer experience and library size, so it’s up to you whether to include them.
```
jakebailey, line 189: The pattern here has been for each bullet to be about a specific compiler option, but this one is clarifying the previous lines. I'm not sure how to do better, though.

PR #2976, opened 2023-11-14
[moduleResolution bundler] List restrictions from initial PR in documentation

Ayc0:
In the v5.0, when the mode `bundler` was introduced, in the PR's description (see https://github.com/microsoft/TypeScript/pull/51669), there was a list of restrictions:

> ## Module syntax restrictions
>
> `--moduleResolution bundler` does not support resolution of `require` calls. In TypeScript files, this means the `import mod = require("foo")` syntax is forbidden; in JavaScript files, `require` calls are not errors but only ever return the type `any` (or whatever an ambient declaration of a global `require` function is declared to return).

But those restrictions aren't listed in the doc:

image

I added those back to the main documentation.

Maybe this isn't the best place for it, and maybe it should live in the cookbook.

This is my 1st contribution to this repo so Idk if I followed every guideline.


Ayc0:

andrewbranch:
Thanks @Ayc0!

Issue #2807, opened 2023-04-26
Playground vNext Wishlist


RyanCavanaugh:
Things we've found useful that should continue to work, *very* roughly in priority order:

 * [ ] "Obvious" stuff: completion, quick info, error squiggles, error list
 * [ ] "Share code" links should be backward compatible
 * [ ] Configuration of commonly-used options, preferably via tsconfig
 * [ ] Version selection
 * [ ] See JS and .d.ts output

Things that would be nice, but likely below a MVP cutline if push comes to shove:

 * [ ] Automatic type acquisition
 * [ ] Multiple files support

Feel free to post other things you'd like to see.

ssalbdivad:
`//  ^?` queries are used in many if not most of the playground links I see.

I'm sure some of that is individual bias, but with access to the VSCode extension ecosystem, would the plan be to enable [vscode-twoslash-queries](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-twoslash-queries) by default?

jcalz:
I'd think the only way a "share code" link could be backward compatible in the face of arbitrary changes to the playground is if it actually brings you to the old playground.  Like, without automatic type acquisition, a lot of existing shared code would just break.  But I don't think you want a copy of the old playground hanging around, so... I'm worried?  Do I have reason to worry?

jakebailey:
We're working on ATA in vscode.dev anyhow, so if we go that direction, it should come "for free".

curran:
I'm wondering if it is likely a light lift or a heavy lift to allow ATA to work with TypeScript v5?

At the most basic level we can make the peer dependency accept v5 like this:

```
"peerDependencies": {
  "typescript": ">=4.4.4 <6.0.0"
}
```

I'm not sure if it would actually work though. Worth a shot!

I'm working on a project where we are using TypeScript v5, running in the browser, and would like to augment our project with ATA.

Related: https://github.com/vizhub-core/vzcode/issues/352

jakebailey:
I'm pretty sure that's just an oversight that can be fixed.

PR #2977, opened 2023-11-15
Allow v5 as peerDependency in ata package

curran:
One small step towards #2807

See comment https://github.com/microsoft/TypeScript-Website/issues/2807

Should in theory solve downstream issue https://github.com/vizhub-core/vzcode/issues/352


packages/ata/package.json
```diff
@@ -26,6 +26,6 @@
     "jest": "^29.5.0"
   },
   "peerDependencies": {
-    "typescript": "^4.4.4"
+    "typescript": ">=4.4.4 <6.0.0"
```
jakebailey: I wouldn't bother setting an upper bound; TypeScript doesn't generally follow semver in this way so 6 is just going to follow 5.9.

curran: Oh nice! Happy to change it.

curran: Updated

curran (commented):
Thanks for reviewing! I removed the upper bound.

curran:
CI failed.
image

It looks like I need to `yarn install` to update the lockfile, as part of the PR. Sorry about that!

jakebailey (approved):

curran:

curran:
Thanks for taking this over the finish line! Much appreciated.

PR #2994, opened 2023-12-09
Update `compilerOptions.jsx` to reflect latest preferences

eps1lon:
`react-jsx` makes the most sense as a default.

[Preview](https://github.com/eps1lon/TypeScript-Website/blob/jsx-default/packages/tsconfig-reference/copy/en/options/jsx.md)

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/jsx.md
+++ b/packages/tsconfig-reference/copy/en/options/jsx.md
@@ -6,11 +6,11 @@ oneline: "Specify what JSX code is generated."
 Controls how JSX constructs are emitted in JavaScript files.
 This only affects output of JS files that started in `.tsx` files.
 
-- `react`: Emit `.js` files with JSX changed to the equivalent `React.createElement` calls
-- `react-jsx`: Emit `.js` files with the JSX changed to `_jsx` calls
-- `react-jsxdev`: Emit `.js` files with the JSX changed to `_jsx` calls
+- `react-jsx`: Emit `.js` files with the JSX changed to `_jsx` calls for production
+- `react-jsxdev`: Emit `.js` files with the JSX changed to `_jsx` calls for development
 - `preserve`: Emit `.jsx` files with the JSX unchanged
 - `react-native`: Emit `.js` files with the JSX unchanged
+- `react`: Emit `.js` files with JSX changed to the equivalent `React.createElement` calls
 
 ### For example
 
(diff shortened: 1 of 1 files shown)


jakebailey:
> TODO: PR changing default in TypeScript actual.

The default in TS is undefined, so I'm not sure there's an actual code change needed there besides potentially changing a diagnostic to suggestion one choice over another.

(Additionally, other runtimes than React I believe still use that setting so I doubt it's going to be deprecated in TS)

jakebailey:
Are the discussions for this somewhere that can be linked?

eps1lon:
>The default in TS is undefined, so I'm not sure there's an actual code change needed there besides potentially changing a diagnostic to suggestion one choice over another.

Yeah, I tried it on a fresh project and you need to specify a value. You can't just run `tsc --jsx`. So all we need to do is document that there is no default and then we discuss in the React team whether people should prefer `react-jsx` or `react-jsxdev` if they have a single tsconfig.json.

packages/tsconfig-reference/copy/en/options/jsx.md
```diff
@@ -75,11 +76,12 @@ declare module JSX {
...
 // @showEmit
 // @noErrors
-// @jsx: react-jsx
+// @jsx: react-native
 export const HelloWorld = () => <h1>Hello world</h1>;
 ```
 
-React 17 dev transform: `"react-jsxdev"`<sup>[[1]](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)</sup>
+
+Legacy runtime: `"react"`
```
eps1lon: Make sure this is language we want to use.

packages/tsconfig-reference/copy/en/options/jsx.md
```diff
@@ -6,11 +6,11 @@ oneline: "Specify what JSX code is generated."
 Controls how JSX constructs are emitted in JavaScript files.
 This only affects output of JS files that started in `.tsx` files.
 
-- `react`: Emit `.js` files with JSX changed to the equivalent `React.createElement` calls
-- `react-jsx`: Emit `.js` files with the JSX changed to `_jsx` calls
-- `react-jsxdev`: Emit `.js` files with the JSX changed to `_jsx` calls
+- `react-jsx`: Emit `.js` files with the JSX changed to `_jsx` calls for production
+- `react-jsxdev`: Emit `.js` files with the JSX changed to `_jsx` calls for development
```
eps1lon:
```suggestion
- `react-jsx`: Emit `.js` files with the JSX changed to `_jsx` calls optimized for production
- `react-jsxdev`: Emit `.js` files with the JSX changed to `_jsx` calls for development only
```

that makes more sense since `react-jsx` can be used in dev as well. It's just that `react-jsxdev` will crash in prod.

jakebailey:
I think #3064 composes with this?

eps1lon:
This one is more focused and can land separately though, no?

jakebailey (approved):

jakebailey:
Close-reopening to verify CI after changes in the repo.

PR #3002, opened 2023-12-17
fix: ata return type

antfu:
It actually returns a Promise. Would be nice to provide that info so users could await for it to resolve.

https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/packages/ata/src/index.ts#L56-L60


orta (approved):

jakebailey (approved):
LGTM but will require a package release.

PR #3003, opened 2023-12-22
fixed typo

DamanjeetSingh1:
In the Generic Types section 'so' is typed instead of 'as', the sentence should be "We could also have used a different name for the generic type parameter in the type, as long as the number of type variables and how the type variables are used line up."


jakebailey:
"so long as" is a perfectly correct phrase; this is fine as-is.

DamanjeetSingh1:
> "so long as" is a perfectly correct phrase; this is fine as-is.

Okay Jake, thanks for the review.

PR #3009, opened 2024-01-03
Added a new example to the top of Template Literal Types.md

DanKaplanSES:
I thought this new example was valuable because I wasn't sure if Template Literal Types could define, "A string that *starts with*, *ends with*, or *contains* another string." 

The reason I put this example so early is because I think it gives the reader a taste of what Template Literal Types can do and I _think_ it's a compelling example that many developers can relate to?

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Type Manipulation/Template Literal Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Type Manipulation/Template Literal Types.md
@@ -12,7 +12,7 @@ They have the same syntax as [template literal strings in JavaScript](https://de
 ```ts twoslash
 // @errors: 2322
 type PopularTopLevelDomains = 'com' | 'net' | 'org' | 'edu' | 'gov';
-type TypicalEmail = `${string}@${string}.${PopularTopLevelDomains}`;
+type TypicalEmail = `${string}@${string}.${PopularTopLevelDomains}`; // A template literal type
 
 // Valid:
 const simpleEmail: TypicalEmail = 'foo@bar.com';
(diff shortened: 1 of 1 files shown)


DanKaplanSES:

andrewbranch:
I'm not sure about this. I think the existing order of the document is pretty good: it starts with the simplest possible example (making a new string literal type via concatenation), then shows what happens with string literal unions, then builds to a complex and powerful example. The example you're adding at the beginning combines several concepts that aren't explained until later, and one concept that is never explained and is missing from this document (`string` placeholders). I think the example is a good way to bring all those concepts together, but I think there should be an explanation of `string` placeholders after the union expansion examples, and I'm not sure I like putting the big example synthesizing several different concepts before the introduction of those concepts.

DanKaplanSES:
@andrewbranch Thanks for your feedback! I agree with everything you're saying.

Instead of discussing the PR, I'd like to explain what motivated me to create it. Actually, you already mentioned it in your reply: `string` placeholders. I didn't know they existed until yesterday (from reading [this stackoverflow post](https://stackoverflow.com/a/68847237/61624)). 

When I read that answer, it seemed like `string` placeholders would be a common use case for Template Literal Types and should be included in the official documentation; I'd like to hear your thoughts. 

Either way, I'll be closing this PR because it's not a good way to teach `string` placeholders anyway.

andrewbranch:
Yeah, I definitely think non-literal types in placeholders are a glaring docs hole.

DanKaplanSES:
@andrewbranch great! Unless it's taking up too much of your time, I might give this another shot in a few days. Thanks again.

PR #3011, opened 2024-01-10
Reorder Pick, Omit, Extract, and Exclude sections

DanKaplanSES:
- The original order of these four sections was: 

   `Pick, Omit, Exclude, Extract, `

   Or, in terms of Keeping and Deleting key/union members, the original order was:

   `Keep, Delete, Delete, Keep`

   I rearranged it to be this:

   `Pick, Omit, Extract, Exclude, `

   Or, in terms of Keeping and Deleting key/union members:

   `Keep, Delete, Keep, Delete`
- I plan to add another commit to change the order to: 

    `Pick, Extract, Omit, Exclude, `

   Or, in terms of Keeping and Deleting key/union members:

   `Keep, Keep, Delete, Delete`

   I think this order is even more useful, because I am often looking at this page trying to figure out which "Keep" utility type I'm looking for, or which "Delete" utility type, and now they will be side-by-side. The reason I'm not putting this in a single commit is because I think this next "improvement" is more subjective than the current, and I want to make it easier to `git reset` a portion

### DIFF
--- a/packages/documentation/copy/en/reference/Utility Types.md
+++ b/packages/documentation/copy/en/reference/Utility Types.md
@@ -188,6 +188,34 @@ todo;
 // ^?
 ```
 
+## `Extract<Type, Union>`
+
+<blockquote class=bg-reading>
+
+Released:  
+[2.8](/docs/handbook/release-notes/typescript-2-8.html#predefined-conditional-types)
+
+</blockquote>
+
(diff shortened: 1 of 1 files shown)


jakebailey:
I am not convinced this order is better; right now, they're roughly in the order of "what are people likely to use". Pick/Omit are the most widely used and useful ones, while Extract is honestly often misused.

PR #3014, opened 2024-01-12
Added Dockerfile for TypeScriptLang-Org

MatanelGordon:
This changes consists of 3 files:
Dockerfile - an actual working dockerfile
.dockerignore - a complementary file to deal with Dockerfile
README - added a sentence under `yarn start` that this deployment is also available via Docker


jakebailey:
I don't know why we need this. The website is solely published via Azure Static Web Apps, by our team.

MatanelGordon:
@jakebailey I needed to run the docs locally on my VM since I am in a closed network. Using docker seems to be the right choice for a case like this.

MatanelGordon:

jakebailey:
I don't think we're going to take this; the right thing to do here is to update https://github.com/microsoft/TypeScript-Website/tree/v2/.devcontainer to have the right node version (like https://github.com/microsoft/TypeScript/tree/main/.devcontainer)

jakebailey (changes requested):
See above.

PR #3017, opened 2024-01-26
Include dates and links to release blog posts in release-notes

danvk:
So far as I can tell, the only difference between the [release blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-2-3/) and the [release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html) is that the former includes a release date.

If you're looking at the release notes for an old release, that's useful information! So this adds it to the docs pages.

I only added dates and blog post links to five releases in this PR. If the maintainers like the idea I'm happy to add them to the remaining releases.

### DIFF
--- a/packages/documentation/copy/en/release-notes/TypeScript 2.0.md
+++ b/packages/documentation/copy/en/release-notes/TypeScript 2.0.md
@@ -5,6 +5,8 @@ permalink: /docs/handbook/release-notes/typescript-2-0.html
 oneline: TypeScript 2.0 Release Notes
 ---
 
+Release date: [September 22, 2016](https://devblogs.microsoft.com/typescript/announcing-typescript-2-0/)
+
 ## Null- and undefined-aware types
 
 TypeScript has two special types, Null and Undefined, that have the values `null` and `undefined` respectively.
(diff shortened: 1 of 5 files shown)


jakebailey:
The markdown files on the website is the same markdown we author the blog post in, while the date on devblogs is inserted by WordPress directly; I'm not totally sure we'd want to make getting content over to the website more manual (it's already a manual process, unfortunately).

DanielRosenwasser:
If I can add these to the front-matter, would that be helpful in the future?

danvk:
@DanielRosenwasser Not sure if that's a question for me or Jake :) Having that information in the Markdown in any form would be helpful for me.

jakebailey:
You could try putting it in the frontmatter, but then it's a question of whether or not Gatsby is going to do anything with that info. Typically you'd put a date in frontmatter because the template is a blog and that sets some sort of date, but I doubt that it will work for this website, which treats these pages as docs.

jakebailey:
Tested locally, and adding `date: "2022-11-24"` to one of the pages doesn't seem to do anything, so there would need to be code changes to `createDocumentationPages` or something to make use of that info and inject it into the generated page.

PR #3019, opened 2024-02-01
fix(sandbox): switch alwaysStrict default to false

JoshuaKGoldberg:
Per my interpretation of [a brief discussion on `alwaysStrict` in Discord]( https://discord.com/channels/508357248330760243/640177465490407425/1196309864260046858), changes the default value for that setting to off.

Note that this doesn't imply anything about TypeScript's recommendations around strict mode. It just changes the playground to not emit a `"use strict";` on top of emitted JS. That string has been a thorn in my side trying to teach TypeScript in workshops for years.

### DIFF
--- a/packages/playground-examples/copy/en/Playground/Config/New Compiler Defaults.ts
+++ b/packages/playground-examples/copy/en/Playground/Config/New Compiler Defaults.ts
@@ -28,7 +28,7 @@ const defaultCompilerOptions = {
   noImplicitThis: true,
   noImplicitReturns: true,
 
-  alwaysStrict: true,
+  alwaysStrict: false,
   allowUnreachableCode: false,
   allowUnusedLabels: false,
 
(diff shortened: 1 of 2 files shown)


orta (approved):

jakebailey:
I am still not entirely convinced that this is a good idea, but, in what conditions do we currently actually emit it? The default settings on the playground don't show `"use strict";` in the JS output.

JoshuaKGoldberg:
If the current file is a script (rather than a module). https://www.typescriptlang.org/play?#code/Q:

Screenshot of the TypeScript playground with no input code. The output is "use strict";

jakebailey:
Ah, I had some stuff sitting in my browser.

jakebailey:
Per https://github.com/microsoft/TypeScript/issues/62213, we're going in the opposite direction on this such that all code will be forcibly strict (and get this prologue if not ESM).

I think better would be to modernize the playground's options to not target CJS.

PR #3021, opened 2024-02-02
Fix sidebar header tree creation

oddmj:
I don't think the depth structure of the document's sidebar represents the depth of the headers in the document very well.

This makes it a little difficult to understand the content.

For example, for "Narrowing" in the handbook, the header in the document looks like this

```md
## `typeof` type guards
# Truthiness narrowing
## Equality narrowing
## The `in` operator narrowing
## `instanceof` narrowing
## Assignments
## Control flow analysis
## Using type predicates
## Assertion functions
# Discriminated unions
# The `never` type
# Exhaustiveness checking
```

In the sidebar on the current page, it looks like this

image

I think this makes more sense. because I think all the h1s should be 1-depth, and the first h2 should be 1-depth.

image

I've also found cases like this with "template-literal-types", "conditional-types", etc.


oddmj:

jakebailey (approved):
Thanks, this definitely looks better.

PR #3022, opened 2024-02-03
Fix `@typescript/vfs` Critical dependency: the request of a dependency is an expression

dimaMachina:
I am getting warnings and after errors due these `require` statements

```
../node_modules/.pnpm/@typescript+vfs@1.5.0/node_modules/@typescript/vfs/dist/vfs.esm.js
Critical dependency: the request of a dependency is an expression

Import trace for requested module:
../node_modules/.pnpm/@typescript+vfs@1.5.0/node_modules/@typescript/vfs/dist/vfs.esm.js
../node_modules/.pnpm/twoslash@0.1.0_typescript@5.3.3/node_modules/twoslash/dist/index.mjs
../node_modules/.pnpm/shikiji-twoslash@0.10.2_typescript@5.3.3/node_modules/shikiji-twoslash/dist/index.mjs
../packages/nextra/dist/server/compile.js
./app/docs/guide/ssg/page.mdx
./.next/static/chunks/nextra-page-map-.mjs
./app/layout.tsx
```

### DIFF
--- a/packages/typescript-vfs/src/index.ts
+++ b/packages/typescript-vfs/src/index.ts
@@ -641,9 +641,9 @@ export function createVirtualLanguageServiceHost(
 }
 
 const requirePath = () => {
-  return require(String.fromCharCode(112, 97, 116, 104)) as typeof import("path")
+  return require('path') as typeof import("path")
 }
 
 const requireFS = () => {
-  return require(String.fromCharCode(102, 115)) as typeof import("fs")
+  return require('fs') as typeof import("fs")
 }


orta:
I think I had to do this to ensure a bunch of bundlers or typescript aren't pulling in node, without addressing that issue (e.g. check out the blame history on that line, and also see what the build steps say around this) - I don't think this should probably get changed until that's been looked at

packages/typescript-vfs/src/index.ts
```diff
@@ -641,9 +641,9 @@ export function createVirtualLanguageServiceHost(
 }
 
 const requirePath = () => {
-  return require(String.fromCharCode(112, 97, 116, 104)) as typeof import("path")
+  return require('path') as typeof import("path")
```
dimaMachina, line 644: @orta thanks for your reply, can it at least be refactored to ?
```suggestion
  return typeof process === 'undefined' ? require(String.fromCharCode(112, 97, 116, 104)) : require('path') as typeof import("path")
```

jakebailey:
This code was added in #1746, but I don't see a particular reason why. This is not the only place in the change where a require was converted into an expression; see also packages/ts-twoslasher/src/index.ts.

orta:
Yeah, must be some side-effect of the Gatsby update - perhaps it doesn't build the playground correctly when removed? Likely that Gatsby updated webpack to v5 (see the addition of node-polyfill-webpack-plugin) and that came with issues

I think you'd want to have at least tested this locally to not be problematic before allowing a change in that space

dimaMachina:
@orta can you send how to reproduce current working behaviour?

currently in `v2` branch, on page http://localhost:8000/dev/twoslash I added `console.log` at the bottom of this file https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescript-vfs/src/index.ts

```
console.log('here', requireFS())
```

and received the following error in the browser

image

dimaMachina:
current behaviour pollutes my `next build` https://github.com/shuding/nextra/actions/runs/7846393108/job/21412984305?pr=2653
image

dimaMachina:
any updates about fixing it?

PR #3027, opened 2024-02-14
Be a bit less kind to javascript

lcarilla:


DanielRosenwasser:
All I'll say is that I'm not a fan of overridable `instanceof` either.

PR #3028, opened 2024-02-15
fix typo

hanyujie2002:


jakebailey:
This is a regional style thing; the text as-written is not incorrect and is used by many US publications.

(I much prefer `.` outside of `"` but that that's not a reason to change it.)

PR #3037, opened 2024-03-03
chore: Update ts-jest to support Typescript 5

xiaoxiyao:
see https://github.com/kulshekhar/ts-jest/issues/4048


jakebailey:
This has already been done in a different PR.

PR #3038, opened 2024-03-03
Fix two issues in the `typescript/vfs` module.

### LINKED ISSUE #2713: Module(vfs): throws an error when the file content is empty string
`@typescript/vfs`

```ts
import ts, { ScriptTarget } from 'typescript'
import { createDefaultMapFromCDN, createSystem, createVirtualTypeScriptEnvironment } from '@typescript/vfs'

;(async () => {
  const compilerOptions = {
    target: ScriptTarget.ES2015,
  }
  const fsMap = await createDefaultMapFromCDN(compilerOptions, ts.version, true, ts)
  fsMap.set('index.ts', '') // error
  const system = createSystem(fsMap)
  const env = createVirtualTypeScriptEnvironment(system, ['index.ts'], ts, compilerOptions)
})()
```

Error:
```
Uncaught (in promise) Error: error TS6053: File 'index.ts' not found.
  The file is in the program because:
    Root file specified for compilation

    at Module.createVirtualTypeScriptEnvironment (index.ts:58:11)
```

It may be because of this:
```typescript
// "@typescript/vfs" line 553:
    getScriptSnapshot: fileName => {
      const contents = sys.readFile(fileName)
      if (contents) { // <=
        return ts.ScriptSnapshot.fromString(contents)
      }
      return
    },
```

xiaoxiyao:
1. 2fa00c539f94f8fb613acdd669486582ab136409 Fix the exception when file content is empty.
 fix #2713
2. 3458b37e7ff419c337bd92cb160aced73ca58058 Fix `moduleDetection` compiler option is not working.


tonivj5:
@xiaoxiyao thank you so much for the PR! ❤️ this commit [3458b37e7ff419c337bd92cb160aced73ca58058](https://github.com/microsoft/TypeScript-Website/pull/3038/commits/3458b37e7ff419c337bd92cb160aced73ca58058) help me a lot undestanding why the code wasn't being transpile as expected. I can confirm it works like a charm 😄

packages/typescript-vfs/src/index.ts
```diff
@@ -555,14 +555,14 @@ export function createVirtualCompilerHost(sys: System, compilerOptions: Compiler
...
-      getSourceFile: fileName => {
+      getSourceFile: (fileName, languageVersionOrOptions) => {
         return (
           sourceFiles.get(fileName) ||
           save(
             ts.createSourceFile(
               fileName,
               sys.readFile(fileName)!,
-              compilerOptions.target || defaultCompilerOptions(ts).target!,
+              languageVersionOrOptions,
```
jakebailey: I suspect this should be this for backwards compat:

```suggestion
              languageVersionOrOptions ?? compilerOptions.target ?? defaultCompilerOptions(ts).target!,
```

jakebailey: (The website does old versions of TS where this may matter)

jakebailey (approved):

PR #3052, opened 2024-04-04
Use explicit file types for module examples

shanefontaine:
In "The `module` compiler option" section of the TS docs, it is difficult to understand that each pair of codeblocks represents a source file and built file. It is obvious in the markdown code due to the fact that the language is specified in the code block, but that is not transparent to the reader of the site. For example, the `ES...` example section has an identical input and output which is confusing for most even given the context of the section it is in or even the page they are on. 

This PR attempts to fix that by adding filename comments above each pair of code blocks that specify that it is an input file (`.ts`) or a built file (`.js`). The syntax chosen matches the syntax from the "Type-only imports and exports" section on the same page.

As an example, here is the existing `ES...` section.

```ts
import x, { y, z } from "mod";
import * as mod from "mod";
const dynamic = import("mod");
console.log(x, y, z, mod, dynamic);

export const e1 = 0;
export default "default export";
```

```js
import x, { y, z } from "mod";
import * as mod from "mod";
const dynamic = import("mod");
console.log(x, y, z, mod, dynamic);

export const e1 = 0;
export default "default export";
```


shanefontaine:

jakebailey (approved):

PR #3053, opened 2024-04-08
Document NoInfer

### LINKED ISSUE #58121: NoInfer introduced in 5.4 is undocumented
### 🔎 Search Terms

doc, NoInfer

- Since 5.4, [NoInfer](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#the-noinfer-utility-type) was introduced, but it remains undocumented.

### ⏯ Playground Link

_No response_

### 💻 Code

_No response_

### 🙁 Actual behavior

`NoInfer` is undocumented.

### 🙂 Expected behavior

`NoInfer` should be documented, may be in the [Utility Types Page](https://www.typescriptlang.org/docs/handbook/utility-types.html). 

### Additional information about the issue

Without the official documentation, it remains unclear what we can and we cannot do with `NoInfer`. For example, must `NoInfer` be applied directly to a generic param? In other words, is `NoInfer<Array<T>>` allowed?

xuhdev:
Not using twoslash for the example here because the document is built with TS 5.0, which would complain that `NoInfer` is an unknown symbol.

Fix microsoft/TypeScript#58121

### DIFF
--- a/packages/documentation/copy/en/reference/Utility Types.md
+++ b/packages/documentation/copy/en/reference/Utility Types.md
@@ -452,6 +452,32 @@ type T4 = InstanceType<Function>;
 //    ^?
 ```
 
+## `NoInfer<Type>`
+
+<blockquote class=bg-reading>
+
+Released:  
+[5.4](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#the-noinfer-utility-type)
+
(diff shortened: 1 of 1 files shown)


xuhdev:

andrewbranch:
Thanks!

PR #3064, opened 2024-04-18
Update JSX documentation

remcohaszing:
This adds some missing information about using JSX in TypeScript.

- Add the `react-jsx` and `react-jsxdev` modes in the introduction. Also mention they are called the classic and automatic runtime.
- Specify where frameworks need to specify the `JSX` namespace.
- Clarify which `jsx` options use which type definitions.
- Specify that JSX runtimes must use package exports in order to support ESM.
- Fix syntax highlighting in JSX code blocks.
- Document that users may use the `@jsxRuntime` comment to override the `jsx` option on a per-file basis.


jakebailey:
Is there anyone who might be able to help verify this? @weswigham maybe? @DanielRosenwasser?

orta:
I can at least confirm that this is how I model the JSX support in my head (I think people are re-looking at this because React 19 is starting to come out) after  giving it a look though. 

The only bits I'm not certain of are the ESM support docs, because I've not had to interact with that system. Feels reasonable to me though.

remcohaszing:
Missing package exports are the reason why the automatic runtime in `react@<18` didn’t work with ESM.

packages/documentation/copy/en/reference/JSX.md
```diff
@@ -70,14 +69,73 @@ This is important for two reasons:
 TypeScript uses the [same convention that React does](http://facebook.github.io/react/docs/jsx-in-depth.html#html-tags-vs.-react-components) for distinguishing between these.
 An intrinsic element always begins with a lowercase letter, and a value-based element always begins with an uppercase letter.
 
+### The `JSX` namespace
+
+JSX in TypeScript is typed by the `JSX` namespace. The `JSX` namespace may be defined in various places, depending on the `jsx` compiler option.
+
+The `jsx` options `preserve`, `react`, and `react-native` use the type definitions for classic runtime. This means a variable needs to be in scope that’s determined by the `jsxFactory` compiler option. The `JSX` namespace needs to specified on the top-most identifier of the JSX factory. For example, React uses the default factory `React.createElement`. This means its `JSX` namespace should be defined as `React.JSX`.
```
eps1lon:
```suggestion
The `jsx` options `preserve`, `react`, and `react-native` use the type definitions for classic runtime. This means a variable needs to be in scope that’s determined by the `jsxFactory` compiler option. The `JSX` namespace needs to be specified on the top-most identifier of the JSX factory. For example, React uses the default factory `React.createElement`. This means its `JSX` namespace should be defined as `React.JSX`.
```

eps1lon: > The `JSX` namespace needs to specified on the top-most identifier of the JSX factory.

global is also still an option I believe. So the "must" would be misleading.

packages/documentation/copy/en/reference/JSX.md
```diff
@@ -70,14 +69,73 @@ This is important for two reasons:
...
+Then in `jsx-runtime.d.ts` and `jsx-dev-runtime.d.ts`:
+
+```ts
+export namespace JSX {
+  // …
+}
+```
+
+Note that while exporting the `JSX` namespace is sufficient for type checking, the production runtime needs the `jsx`, `jsxs`, and `Fragment` exports at runtime, and the development runtime needs `jsxDEV` and `Fragment`. Ideally you add types for those too.
+
```
eps1lon: Since this is all specific to a single library, I don't think we need docs for that. The docs are equal to the current implementation in `@tyes/react` or `react`.

It was already problematic that these were introduced without working with React types maintainers. Writing docs make it seem like the TS team just need to update docs and React types maintainers just need to pay attention which is not (and [was not) healthy for the ecosystem](https://github.com/microsoft/TypeScript/pull/39199#issuecomment-731090590).

packages/documentation/copy/en/reference/JSX.md
```diff
@@ -386,11 +444,30 @@ You can customize the type by specifying the `JSX.Element` interface.
 However, it is not possible to retrieve type information about the element, attributes or children of the JSX from this interface.
 It is a black box.
 
+## The JSX function return type
+
+By default, JSX functions must return the `JSX.Element` type. However, this doesn’t always represent runtime behaviour. As of TypeScript 5.1, you can specify `JSX.ElementType` to override what is a valid JSX component type. The default looks something like this.
```
eps1lon:
```suggestion
By default, function components must return `JSX.Element | null`. However, this doesn’t always represent runtime behaviour. As of TypeScript 5.1, you can specify `JSX.ElementType` to override what is a valid JSX component type. The default looks something like this:
```

eps1lon: We should mention here that the type of the props is not extracted from this type at the moment. Props type inference still only looks at the first argument of functions or constructors, otherwise looks up props from `JSX.IntrinsicElements`

packages/documentation/copy/en/reference/JSX.md
remcohaszing: I don’t understand. The word _“must”_ isn’t used here. The global `JSX` namespace is documented a few lines down on line 129.

packages/documentation/copy/en/reference/JSX.md
remcohaszing: How is this specific to a single library? This applies to all libraries that support JSX.

packages/documentation/copy/en/reference/JSX.md
eps1lon: "needs" was used not "must", sorry for the confusion. Which effectively means the same. Correcting this later is not sufficient. If TypeScript has a fallback, you don't *need* to have it scoped.

packages/documentation/copy/en/reference/JSX.md
eps1lon: `react-jsx` is specific to libraries other than React?

remcohaszing: None of the possible options are specific to React, but most have the word `react` in it.

| Name | Meaning |
| --- | --- |
| `preserve` | Preserve JSX, emit `.jsx` extension |
| `react-native` | Preserve JSX, emit `.js` extension |
| `react` | JSX classic runtime |
| `react-jsx` | JSX automatic runtime |
| `react-jsxdev` | JSX automatic development runtime |

packages/documentation/copy/en/reference/JSX.md
remcohaszing: Would _“should”_ be better in your opinion?

packages/documentation/copy/en/reference/JSX.md
eps1lon: I understand that for classic because that's old. No idea why this was continued with `react-jsx`

jakebailey (approved):

jakebailey:
Gave this a reread, and everything seems to check out to me.

PR #3066, opened 2024-04-19
Make String, Number and Boolean types more clear.

udohjeremiah:
The current documentation lacks clarity in its explanation of `String`, `Number`, and `Boolean`. Rather than providing a clear definition, the document simply states that these are special built-in types that are rarely used in code.

It would be more helpful to explain what these types are when they are typically used, and why they are not commonly seen in code. Additionally, it adds links to the sources where users can learn more about these types.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -24,7 +24,7 @@ As you might expect, these are the same names you'd see if you used the JavaScri
 - `number` is for numbers like `42`. JavaScript does not have a special runtime value for integers, so there's no equivalent to `int` or `float` - everything is simply `number`
 - `boolean` is for the two values `true` and `false`
 
-> The type names [`String`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String), [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number), and [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) (starting with capital letters) are legal and refer to the object wrapper for their respective data types. You should rarely find yourself using them except when you need to work with methods or properties specific to these data types. However, _Always_ use `string`, `number`, or `boolean` for types.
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v2/Everyday Types.md
```diff
@@ -24,7 +24,7 @@ As you might expect, these are the same names you'd see if you used the JavaScri
 - `number` is for numbers like `42`. JavaScript does not have a special runtime value for integers, so there's no equivalent to `int` or `float` - everything is simply `number`
 - `boolean` is for the two values `true` and `false`
 
-> The type names `String`, `Number`, and `Boolean` (starting with capital letters) are legal, but refer to some special built-in types that will very rarely appear in your code. _Always_ use `string`, `number`, or `boolean` for types.
+> The type names [`String`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String), [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number), and [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) (starting with capital letters) are legal and refer to the object wrapper for their respective data types. You should rarely find yourself using them except when you need to work with methods or properties specific to these data types. However, _always_ use `string`, `number`, or `boolean` for types.
```
jakebailey, line 27: I can't say I like the new prose; the first part gives a load of info and says you may want to use them, but then says "but don't use them". The whole point is to draw attention away from these bad types. If we're going to explain anything, we should probably be explaining why it's undesirable to use them, not restate that they aren't good.

udohjeremiah: You're right @jakebailey but I sometimes see this as a pattern in the docs, where stuff is explained and then we're told not to use them, stating the reason why.

The point of stating when they are used is that since TypeScript is **"JavaScript's runtime"**, it isn't best to just shun not using those types, since as the PR said, they are the most obvious way to use methods or properties when trying to work with these data types. For example, doing:

```ts
let a = "3.14259";
Number(a).toFixed(a);
```

What better and more concise code could we use to achieve the above? So that's the point. State what they are, when they are used, and then tell readers not to use them in types. Instead of just saying: "Hey, this is bad don't touch it."

PR #3072, opened 2024-04-23
Fix missing typescript peer dependency package.json

KiranJKurian:
Typescript is being used from src as a dependency but is only listed as a dev dependencies in the package.json. Package.json needs to be updated to include it as a peer dependency so package managers like pnpm can correctly resolve the dependency.


jakebailey:
This PR now requires a CLA since it now touches too many files; CLAs are normally not required for small changes like this was originally. Since the changeset is just procedure, I'll probably revert the CLA change and add the changeset manually elsewhere, unless you have the time to send the message listed above (which would be much easier).

KiranJKurian:
@microsoft-github-policy-service agree [company="Robinhood Markets"]

KiranJKurian:
@microsoft-github-policy-service agree company="Robinhood Markets"

jakebailey:
Awesome, thank you!

PR #3073, opened 2024-04-23
Removing app insights

navya9singh:


packages/typescriptlang-org/src/templates/scripts/setupLikeDislikeButtons.ts
```diff
@@ -4,14 +4,6 @@ export const setupLikeDislikeButtons = (slug: string, i: any) => {
   if (!likeButton || !dislikeButton) return
 
   const clicked = (eventName: string) => () => {
-    // @ts-ignore
```
jakebailey, line 7: Honestly, I'm wondering if we should be ditching the buttons altogether for now, at least until we can move to something different (or the same but recreated). Otherwise we'll have thumbs up/down which just don't do anything. But maybe that's a little too invasive for the short term. But we definitely can't leave them.

jakebailey: Thankfully removing it doesn't seem too hard; happy to push that up unless you want to do it.

jakebailey: Or, actually, we should just make them not show up; that's a lot easier to undo.

jakebailey: (so, ctrl-f `dislike` and comment out, more or less)

navya9singh: I've commented out the buttons. `setupLikeDislikeButtons.ts` creates these, so I can comment that out too if needed.

jakebailey: Looks good; that's what I had tried locally too.

jakebailey (approved):
LGTM; can revert later once we figure out what's next.

PR #3076, opened 2024-04-24
for differences between interfaces and type aliases, add note about c…

btoo:
…ompiler performance when using interfaces with extends vs type aliases with intersections


orta:
We wanted to avoid putting performance recommendations in this list before, that opinion could be worth re-examining but I'd probably recommend not merging

jakebailey:
Given this is a direct link to the wiki, I'm not against including it.

btoo:
sorry for closing. i force pushed in https://github.com/microsoft/TypeScript-Website/pull/3076#event-12629106864 and the PR was auto-closed and now prevents me from re-opening so I'm moving this to https://github.com/microsoft/TypeScript-Website/pull/3087

PR #3079, opened 2024-04-25
Migrate to pnpm and changesets

jakebailey:
This switches the repo to pnpm/changesets, the monorepo and deployment method we use for dt-tools and attw (and me personally).

Notably:

- Internal deps are now marked as `workspace:*`, so we don't need to do the bumping manually.
- Publishing is now via `pnpm publish -r`, which rewrites the specifiers automatically.
- Changesets auto-bumps the version numbers themselves.
- Builds are a lot faster (multiple minutes better).

TODO:

- [x] Figure out the changesets private package setting; see also https://github.com/microsoft/DefinitelyTyped-tools/pull/1003
- [x] Perhaps it should be `workspace:^` to be less strict.
	- For now, no, as we used explicit deps previously.
- [x] Docs on how to use changesets
- [x] Make CI expectations happy


.changeset/config.json
```diff
@@ -0,0 +1,15 @@
...
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "v2",
+  "updateInternalDependencies": "patch",
+  "ignore": [],
+  "privatePackages": {
+    "tag": false,
+    "version": false
+  }
```
jakebailey, line 14: I didn't know about these but they prevent changeset from caring about private packages.

jakebailey:
Hey, it works.

.changeset/config.json
jakebailey: Doesn't seem to work: https://github.com/microsoft/DefinitelyTyped-tools/pull/1003

jakebailey:
I enabled testing on Windows which has exposed that the VFS doesn't pass. May need to re-disable tests there.

jakebailey:
Yay, got it working on all OSs.

andrewbranch (approved):
Amazing. What’s the remaining CI status thing?

jakebailey:
> Amazing. What’s the remaining CI status thing?

GitHub is configured to require the CI check named "tests" but that doesn't work when it turns into a matrix as the name changed. Which is silly. I plan to use this trick: https://github.com/microsoft/typescript-benchmarking/blob/7371a75e3d5862eaffabaca749b94cfa9b934bc9/.github/workflows/ci.yml#L132

jakebailey:
This is looking good to go, but I'm going to do one last check and pack each package and compare it to what's on npm to ensure that the build is actually working as intended.

navya9singh (approved):

jakebailey:
Diffing was fruitful and found a few issues. The packages that can be published after this PR should be correct, plus or minus that the other gatsby 5 and such PR updated some deps.

PR #3081, opened 2024-04-25
Update Object Types.md

adriancuadrado:
Added a note about the `in` keyword in Index Signatures

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Object Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Object Types.md
@@ -297,6 +297,23 @@ interface NotOkay {
 
 </details>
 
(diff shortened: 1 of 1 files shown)


adriancuadrado:

jakebailey:
This file is generated; I don't think it should be updated by hand like this. This file is sourced from ShemaStore, so would need to be fixed there and revendored. (There are scripts to do this IIRC)

adriancuadrado:
@jakebailey
> This file is generated; I don't think it should be updated by hand like this. This file is sourced from ShemaStore, so would need to be fixed there and revendored. (There are scripts to do this IIRC)

I don't know how to do that :(

jakebailey:
I think I was wrong and this isn't actually generated from schemastore anymore.

PR #3091, opened 2024-05-08
Move preview to new deployment, pin dependencies, misc cleanup

jakebailey:
Closes #3062


.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,45 @@
+name: Publish packages
```
jakebailey, line 1: This is just .github/workflows/v2-merged-staging.yml without the staging publish (which no longer worked).

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,44 @@
...
+          yarn workspace typescriptlang-org setup-playground-cache-bust
+          yarn build
+        env:
+          YARN_CHECKSUM_BEHAVIOR: ignore
+
+      - name: Build website v2
+        run: |
+          yarn build-site
+          cp -r packages/typescriptlang-org/public site
+
```
jakebailey, line 38: I'm not sure how much of this is needed to get valid packages, but I'm not taking any chances until we can rework this.

navya9singh (approved):

PR #3093, opened 2024-05-09
Adding back app insights

navya9singh:


.github/workflows/deploy-prod-static.yml
```diff
@@ -71,3 +71,10 @@ jobs:
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4
+
+  # Update the site stats
+      - run: node packages/typescriptlang-org/scripts/updateAppInsightsGitHubIssue.js
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          APP_INSIGHTS_API_KEY: ${{ secrets.APP_INSIGHTS_API_KEY }}
+          APP_INSIGHTS_ID: ${{ secrets.APP_INSIGHTS_ID }}
```
jakebailey: We should just move this to a new workflow on a schedule; having it be in this deploy environment won't work as we're explicitly not cloning the reop here.

jakebailey: And also it'll need `contents: read` and `comments: write`, I believe.

jakebailey (changes requested):

.github/workflows/weekly-stats.yml
```diff
@@ -0,0 +1,40 @@
...
+      
+      - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # v4.0.0
+
+      - name: Build website
+        run: |
+          pnpm install
+          pnpm docs-sync pull microsoft/TypeScript-Website-localizations#main 1
+          pnpm bootstrap
+          pnpm run --filter=typescriptlang-org setup-playground-cache-bust
+          pnpm build
```
jakebailey:
```suggestion
      - run: pnpm install
```

.github/workflows/weekly-stats.yml
```diff
@@ -0,0 +1,40 @@
...
+  comments: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+        with:
+          # Fetch the full history, to build attribution.json
+          fetch-depth: 0
```
jakebailey:
```suggestion
```

jakebailey (commented):
Seems good, though no need do any special builds I don't think,

jakebailey (approved):

Issue #2804, opened 2023-04-25
Next Steps for the TypeScript Website (April 2023 edition)


RyanCavanaugh:
# Next Steps on the TypeScript Website

### TL;DR

We're taking this time to deliver a smaller, higher-quality, lower-cost version of the website that serves our needs while still leaving us time to do the work that we're best suited to do.

Key points:
 * Declaring "issue bankruptcy"
 * Shutting down the localization effort
 * PRs will be handled

## Background

For transparency: our team does not have any dedicated headcount for website content creation, design, or maintenance. Effectively, this means any work done for the website is in lieu of [other work](https://github.com/microsoft/TypeScript/issues/), much of which requires very specialized expertise. Conversely, a lot of the content and functionality on the website today is *not* differentiated in terms of the value it delivers. The TypeScript Playground, for example, largely replicates functionality found in VS Code or in other "Code Sandbox" websites, and it's slower and more difficult for us to deliver that functionality than it would be for someone who does web development as their full-time trade.

Microsoft web properties are rightfully held to high standards for compliance on *many* different axes, and we do not have adequate resourcing to maintain the current scope of the website while meeting those standards without taking away time from other higher-impact work. We plan to reduce the scope and scale of the website down to the minimum required to fulfill our needs.

#### Localizations

We were heartened to see the enthusiasm for the community-driven localization of the website and Handbook. However, the current state of localization leaves much to be desired for other reasons:

 * We have no mechanism for keeping localizations up-to-date
 * We have no mechanism for validating that localizations are correct
 * Localization creates certain novel compliance obligations that are difficult or impossible to address without a fluent speaker of those languages
 * Many pages are sitting "half-localized" with no actionable path toward 100%
 * Observed traffic data for most localizations is *extremely* low

Mitigations:
 * There have been previous language-specific efforts to localize Handbook content and these can now continue
 * Machine translation should be "good enough" in the vast majority of cases, since we strive to write in simple sentences in our documentation. PRs to simplify language for the sake of more accurate machine translations will be accepted
 * Our content has appropriate licensing to allow for this work to continue elsewhere, if desired

#### Playground Configurability

The goal of the TypeScript Playground is to allow TS users to easily try out TypeScript features and quickly examine the behavior of small examples.

The Playground isn't really intended for use for some of the other cases that it's currently being used for:
 * Complex bug investigation
 * Plugin hosting environment
 * Comprehensive configuration testing
 * Complex code sharing scenarios

[...]

PR #3100, opened 2024-05-12
Add context about when to use type predicates and assertion functions

Masstronaut:
Updated based on discussion in [#57436](https://github.com/microsoft/TypeScript/issues/57436)

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -415,7 +415,9 @@ function example() {
 
 ## Using type predicates
 
-We've worked with existing JavaScript constructs to handle narrowing so far, however sometimes you want more direct control over how types change throughout your code.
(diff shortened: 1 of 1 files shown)


jakebailey:
"supersede" doesn't seem like the right term to me; "bypass" or "override" seem more accurate.

Masstronaut:
@jakebailey great suggestion. I wasn't thrilled with "supersede" but wasn't coming up with a better term. I think "bypass" does a much better job conveying the behavior.

Masstronaut:
@jakebailey anything else I need to do here? It's my first time contributing to the TS docs so I'm blissfully unaware of any other process/requirements to get this PR merged in.

jakebailey:
See https://github.com/microsoft/TypeScript-Website/issues/2804, we don't have much bandwidth to review all but critical changes. I wouldn't want to merge this without someone like @RyanCavanaugh or @DanielRosenwasser giving it a look.

PR #3104, opened 2024-05-16
Remove old preview site comments on new comment

jakebailey:
Right now, the action comments every time a build succeeds. If we keep pushing to a PR, that's basically one comment per push.

We can either only ever have one comment, or just remove old comments and leave the final one. I think the latter is better as it'll clearly show for which commits a preview site applies.

While here, ensure preview site jobs don't race.


andrewbranch (approved):
Thanks!

PR #3108, opened 2024-05-20
Always deploy preview if site artifact is present

jakebailey:
Rather than checking for success, just always deploy when the artifact is present. It's okay if tests fail as long as the site actually built.


jakebailey:
Oops, added the wrong "a" name

andrewbranch (approved):

PR #3118, opened 2024-05-22
Fix type-only import link to point to the reference document

pcdevil:
I noticed the [Enum's documentation](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls) point to a non-existing part of the Modules Reference page, so after cross-checking the reference I updated it to the right one.


pcdevil:
updated link is available here:
https://victorious-plant-05c166c10-3118.centralus.5.azurestaticapps.net/docs/handbook/enums.html#const-enum-pitfalls
(the search redirects to the live site)

jakebailey (approved):
Thanks!

PR #3119, opened 2024-05-22
Temporarily ignore private packages in changeset config

jakebailey:
Until changesets is fixed to also ignore private packages for tagging, releasing, etc, manually list out all private packages we currently have so they are never looked at.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,7 +7,17 @@
   "access": "public",
   "baseBranch": "v2",
   "updateInternalDependencies": "patch",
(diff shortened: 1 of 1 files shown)


jakebailey:
Hm, this is not sufficient to prevent tagging.

jakebailey:
Ah, no, that doesn't work without https://github.com/changesets/changesets/pull/1361/files#diff-07a259f2798961cade28a82496fe9953ebc6985cea7c2cd172f025cf36bcf08a

PR #3122, opened 2024-05-23
Disable playground plugin script in bootstrap

jakebailey:
This script changes the list of plugins on bootstrap into a committed file, which causes differences when building. This is annoying for local dev, but also fails CI because I made it error on file diffs like that.

The fixes could be:

- gitignore the file; builds are less deterministic
- Stop updating it

I'm going to opt for the latter for now. There do not seem to be new plugins being created these days. We can manually rerun the script later or figure something better out.

### DIFF
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -9,7 +9,7 @@
     "dist"
   ],
   "scripts": {
-    "bootstrap": "node scripts/getListOfPluginsFromNPM.js",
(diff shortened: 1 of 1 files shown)


navya9singh (approved):

jakebailey:
Then again, builds are already non-deterministic thanks to the releases. Maybe I do just gitignore it.

jakebailey:
Eh, let's just do this for now to unblock other changes.

PR #3128, opened 2024-05-30
Testing new CDN endpoint changes

jakebailey:
This change is temporary; this CDN URL will be deleted after testing.


jakebailey:
Cool, that works.

PR #3133, opened 2024-06-03
Set trailingSlash to try and fix zip path

jakebailey:


packages/typescriptlang-org/gatsby-config.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
   flags: {
     DEV_SSR: false,
   },
+  trailingSlash: "ignore",
```
jakebailey, line 28:
```suggestion
  trailingSlash: "never",
```

packages/typescriptlang-org/gatsby-config.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
   flags: {
     DEV_SSR: false,
   },
+  trailingSlash: "never",
```
jakebailey:
```suggestion
  trailingSlash: "ignore",
```

jakebailey:
#3138

PR #3144, opened 2024-06-05
Fix: Remove irrelevant errors from example

foxcaulfield:
I think errors like these aren't appropriate for this example since it's about different matters.
Screenshot from 2024-06-05 22-11-28


jakebailey (approved):
Oops

PR #3149, opened 2024-06-08
correct the example of const keyword

hanyujie2002:
`a.push(102);` is legitimate, n `D:` is a misleading expression

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for Functional Programmers.md
@@ -541,8 +541,8 @@ immutable. The referent is still mutable:
 
 ```js
 const a = [1, 2, 3];
-a.push(102); // ):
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/get-started/TS for Functional Programmers.md
```diff
@@ -541,8 +541,8 @@ immutable. The referent is still mutable:
 
 ```js
 const a = [1, 2, 3];
-a.push(102); // ):
-a[0] = 101; // D:
```
jakebailey, line 545: The original code is trying to show that even though `a` is const, it can still be modified through methods and array access. The emoji are trying to imply that this is "bad" in that even when const, it's still writable.

If anything, we should just write a comment to the side that says like `// no error` for each line, not change the code itself.

jakebailey (changes requested):

PR #3156, opened 2024-06-12
Cleanup "ts=Nightly", aka "ts=next"

rotu:
1. Fix grammar and clarity in documentation
2. Clean up broken conditional (`tsVersionParam && tsVersionParam === "Nightly" || tsVersionParam === "next"` has a useless truthy check which is ignored in the case of "next")
3. ~~Describe as "next" in the redirect page.~~


rotu:

packages/typescriptlang-org/src/templates/play.tsx
```diff
@@ -111,7 +111,7 @@ const Play: React.FC<Props> = (props) => {
           div.style.webkitAnimation = ""
         })
 
-        document.getElementById("loading-message")!.innerHTML = `This version of TypeScript <em>(${tsVersion?.replace(/</g, "-")})</em><br/>has not been prepared for the Playground<br/><br/>Try <a href='/play?ts=${latestRelease}${document.location.hash}'>${latestRelease}</a> or <a href="/play?ts=next${document.location.hash}">Nightly</a>`
+        document.getElementById("loading-message")!.innerHTML = `This version of TypeScript <em>(${tsVersion?.replace(/</g, "-")})</em><br/>has not been prepared for the Playground<br/><br/>Try <a href='/play?ts=${latestRelease}${document.location.hash}'>${latestRelease}</a> or <a href="/play?ts=next${document.location.hash}">next</a>`
```
jakebailey: Although the tag is "next", we pretty much always refer to it as the nightly build. I would suggest not changing this.

rotu: Okay. Should `ts=Nightly` instead be the default? I'm unclear why the two names in the first place.

jakebailey: For the URLs, I would always use `next` as that's the tag on npm, it's just the human language stuff would be better staying as "nightly".

jakebailey (approved):

PR #3157, opened 2024-06-14
Update Type Compatibility.md

ComplexPlane:
I believe `undefined` and `null` are not assignable to `never` when strict null checks is on.


jakebailey (approved):
This looks right; matches the table above too.

PR #3158, opened 2024-06-15
Remove a segment of meaningless HTML code

hanyujie2002:


jakebailey:
Not clear this is meaningless; this can be used to link to this section by ID, so may have been out out there for compat.

PR #3162, opened 2024-06-20
Correct a code example

hanyujie2002:
In the original code example, `g()` is called before `f()` being done.


jakebailey (changes requested):
The code is correct as-witten; functions are hoisted and can be called before they appear to be declared.

hanyujie2002:
https://github.com/microsoft/TypeScript-Website/blob/26343c8b9d6e5a11501796ad0f44e5877d7645c8/packages/documentation/copy/en/reference/Variable%20Declarations.md?plain=1#L56-L75

True the code is legitimate. But it is not in line with the text above. In the code example, `g()` was supposed to be called after `f()` being done.

PR #3165, opened 2024-06-25
`Host` uses `getDirectories` from `System`

edemaine:
I found that a `Host` created from `createVirtualCompilerHost`/`createVirtualTypeScriptEnvironment` ignores the `getDirectories` provided by the `System`, despite `getDirectories` being a required component of `System`. Instead it forces a function that returns `[]` in all cases.

This is problematic for fs-based systems. The end result in my application is that TypeScript fails to preload all the types in `node_modules/@types` because it's unable to list that directory.

This one-line fix works for my application, and I added a somewhat realistic test (via existing `jest` definitions in `node_modules/@types`) that works only with this change.

### DIFF
--- a/packages/typescript-vfs/src/index.ts
+++ b/packages/typescript-vfs/src/index.ts
@@ -572,7 +572,6 @@ export function createVirtualCompilerHost(sys: System, compilerOptions: Compiler
       getCanonicalFileName: fileName => fileName,
       getDefaultLibFileName: () => "/" + ts.getDefaultLibFileName(compilerOptions), // '/lib.d.ts',
       // getDefaultLibLocation: () => '/',
-      getDirectories: () => [],
       getNewLine: () => sys.newLine,
       getSourceFile: (fileName, languageVersionOrOptions) => {
         return (
--- a/.changeset/many-avocados-learn.md
+++ b/.changeset/many-avocados-learn.md
@@ -1,0 +1,5 @@
+---
+"@typescript/vfs": patch
+---
+
+Use System's getDirectories if it's provided when constructing Host
--- a/packages/ts-twoslasher/test/fixtures.test.ts
+++ b/packages/ts-twoslasher/test/fixtures.test.ts
@@ -9,6 +9,11 @@ expect.extend({ toMatchFile })
 // To add a test, create a file in the fixtures folder and it will will run through
 // as though it was the codeblock.
 
+const defaultCompilerOptions = {
+  // avoid extra annotations from @types/node
+  types: []
+}
+
 describe("with fixtures", () => {
   // Add all codefixes
   const fixturesFolder = join(__dirname, "fixtures")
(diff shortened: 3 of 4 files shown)


packages/typescript-vfs/src/index.ts
```diff
@@ -572,7 +572,6 @@ export function createVirtualCompilerHost(sys: System, compilerOptions: Compiler
       getCanonicalFileName: fileName => fileName,
       getDefaultLibFileName: () => "/" + ts.getDefaultLibFileName(compilerOptions), // '/lib.d.ts',
       // getDefaultLibLocation: () => '/',
-      getDirectories: () => [],
```
jakebailey, line 575: My recollection is that this is a newish call; I would suggest putting it before the spread so that older runtimes won't crash. The TS website in particular supports very old versions.

edemaine: I tried putting it before, and TypeScript type checking failed because it claimed `getDirectories` was specified twice. I'm all for backward-compatibility; is it OK to use `// @ts-ignore`?

Also, speaking of backward compatibility, as far as I can tell, uses of this object already check for its existence. For example:

https://github.com/microsoft/TypeScript/blob/e70904a24fba0e17c6a8acc9f592c80d41543110/src/compiler/moduleNameResolver.ts#L816

But it might be safer to provide the default.

jakebailey: Hm, I guess it's complaining that it always overwrites; it's definitely optional in older versions of TS, I'm just thinking about whether or not this was added to make some other behavior work. I can't imagine why it not being provided _wouldn't_ be equivalent to an empty array, though.

jakebailey: Easy way to tell, though; I'll have this PR build a preview and we can play around with the playground there.

edemaine: Thanks!

If we leave it this way, would it be a good idea to make `getDirectories` an *optional* property of `System` and `Host`? It seems to be getting treated that way, and this would enforce checking whether it's defined... (Of course, that's an issue for the main TS repo.)

jakebailey: On TS? It's optional in some, required in others. Your linked code is `ModuleResolutionHost` where it's explicitly optional.

edemaine: Ah, and it seems to already be optional in `CompilerHost` (via `ModuleResolutionHost`) which is what this host-building function claims to return anyway. So I think we'd be OK even if the `System` didn't have one as it's supposed to.

jakebailey:
The playground still works, though the test failures have shown the effect of this on twoslash; now it's falling back to the disk and we're getting more info, it seeems?

edemaine:
Yes, it seems it was getting [extra documentation for `console` from `@types/node`]( https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/console.d.ts). I think for repeatability independent of that library, it's best to add `types: []` to avoid any of this reading. I've modified accordingly.

edemaine:
@jakebailey Let me know if there's anything I should do on this PR. Happy to make any changes you'd like!

jakebailey:
I think the main concern is breaking people who depended on this behavior (e.g. shiki?) but I serious doubt it's a problem.

edemaine:
Understood. FWIW, this shouldn't affect any browser uses; it's only in Node where additional type definitions would get picked up, which could break type checking but could be fixed with a `types: []` specifier.

On the other hand, from our perspective (building [our own CLI for typechecking](https://civet.dev/getting-started#typechecking) in the style of `tsc`, but for [another language](https://civet.dev/)), the current system is pretty broken. Of course, I could work around the issue by overriding the output, but this PR seems to bring expected behavior.

But I understand compatibility is a tricky issue; let me know if there's anything I can do to help investigate.

jakebailey (approved):

PR #3167, opened 2024-06-28
chore: use `debug` package for debug logging

kermanx:
Previously, `!!process.env.DEBUG` was used to check whether to debug in `@typescript/vfs` and `@typescript/twoslash`. However, most of the packages are using [the `debug` package](https://npmjs.com/debug) to debug. When we set `DEBUG=vite:hmr`, `@typescript/vfs` will output **lots** of things, which immediately fill the terminal and prevent us from watching logs from Vite.

Also, the `debug` package has been installed as a dependency in `@typescript/vfs`, but seems unused: https://github.com/microsoft/TypeScript-Website/blob/c341935c7f3b7b34812a7438b1b0de5c5cc42e04/packages/typescript-vfs/package.json#L44

And the README also said it's using the `debug` package: https://github.com/microsoft/TypeScript-Website/blob/c341935c7f3b7b34812a7438b1b0de5c5cc42e04/README.md?plain=1#L33

### DIFF
--- a/.changeset/chilled-bulldogs-cross.md
+++ b/.changeset/chilled-bulldogs-cross.md
@@ -1,0 +1,6 @@
+---
+"@typescript/vfs": patch
+"@typescript/twoslash": patch
+---
+
+Use `debug` package for debug logging
--- a/.changeset/chilled-bulldogs-cross.md
+++ b/.changeset/chilled-bulldogs-cross.md
@@ -1,5 +1,6 @@
 ---
 "@typescript/vfs": patch
+"@typescript/sandbox": patch
 "@typescript/twoslash": patch
 ---
 
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^25.0.7",
+    "@types/debug": "^4.1.12",
     "@types/jest": "^29.5.12",
     "dts-cli": "^2.0.5",
     "jest": "^29.5.0",
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -45,6 +45,7 @@
   "dependencies": {
     "@typescript/ata": "workspace:*",
     "@typescript/vfs": "workspace:*",
+    "debug": "^4.3.4",
     "lz-string": "^1.5.0"
   },
   "devDependencies": {
--- a/packages/ts-twoslasher/src/index.ts
+++ b/packages/ts-twoslasher/src/index.ts
@@ -10,7 +10,7 @@ import { createSystem, createVirtualTypeScriptEnvironment, createFSBackedSystem 
 
 import Debug from 'debug'
 
-const log = Debug("typescript:vfs")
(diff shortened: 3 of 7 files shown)


packages/ts-twoslasher/src/index.ts
```diff
@@ -15,7 +8,9 @@ import { validateInput, validateCodeForErrors } from "./validation"
 
 import { createSystem, createVirtualTypeScriptEnvironment, createFSBackedSystem } from "@typescript/vfs"
 
-const log = shouldDebug ? console.log : (_message?: any, ..._optionalParams: any[]) => ""
+import Debug from 'debug'
+
+const log = Debug("typescript:vfs")
```
jakebailey: This looks like a copy paste error.

kermanx: My bad. Resolved.

jakebailey:
CI will fail here because this change is missing a changeset.

packages/sandbox/package.json
```diff
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^25.0.7",
+    "@types/debug": "^4.1.12",
```
jakebailey, line 53: This seems to imply that `sandbox` needs a dep on `debug`; I think that's true given we vendor the vfs into this package?

kermanx: Yes. The DTS build of `@typescript/sandbox` would fail unless `@types/debug` is installed, because `@typescript/vfs` is vendored.

jakebailey: Right but I mean that this package should also depend on "debug" too so the code doesn't crash.

.changeset/chilled-bulldogs-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@typescript/vfs": patch
+"@typescript/twoslash": patch
```
jakebailey:
```suggestion
"@typescript/vfs": patch
"@typescript/sandbox": patch
"@typescript/twoslash": patch
```

jakebailey:
Checking out the vendored code, it does now import from debug. I'm not 100% certain if that's desirable, though; I believe this code is intended to be loaded in the browser directly so this code may break there now that it's not dependency free.

What confuses me though is why `lzstring` is bundled and not `debug`. I guess we locally vendor lzstring? Maybe we need to tweak our config to ensure that these packages are still self contained?

@orta do you recall the restrictions here?

orta:
I just didn't think that debug was important enough of a dependency to add it to the playground's web bundle - given how little I was using of it, could replicate the check for the substring in this package to copy the behavior they are looking for?

jakebailey:
The debug module does a lot of stuff, especially on web; I'm not sure it would be exactly trivial to copy that bit.

But, I think maybe we could just change the original code to check `DEBUG=typescript:vfs` (and similar) instead, and just not document that we use the `debug` module?

(sorry for this reversal, I just looked at the output and noticed that this would break someone)

orta:
Yeah, I think that's the right call

PR #3171, opened 2024-06-30
fix example imports

frankie303:


jakebailey:
Why change this? Extensions are legal, and even required in newer module modes.

frankie303:
@jakebailey I didn't know that js files can include types (in example, type is imported from js file). So I thought that was a mistake.

jakebailey:
TypeScript doesn't rewrite extensions; an import may say .js but point to a TS file, or that file may be .js but contain JSDoc or inferred types under allowJs.

frankie303:
Thank you so much for the explanation!

PR #3173, opened 2024-07-03
Adds VFS deleteFile method

### LINKED ISSUE #2799: missing `deleteFile` function in VirtualTypeScriptEnvironment
the VirtualTypeScriptEnvironment interface has two file-related functions `createFile` and `updateFile`, but seemingly no `deleteFile`.

https://github.com/microsoft/TypeScript-Website/blob/b89d197063b6849bbfc2edfb033f6082be950636/packages/typescript-vfs/src/index.ts#L22-L23

is this intended?
i've tried using the `VirtualTypeScriptEnvironment.sys.deleteFile` function, but it is undefined.  

my current workaround is to set the file contents to a single-space string (completely empty strings currently throw errors as reported here: #2713).  
it's not too big an issue for me right now since my virtual files are not directly visible and their names are auto-incremented, but i figured this would be worth posting here.

marklundin:
This PR adds a `deleteFile` function to `typescript/vfs` which works alongside the `createFile` and `updateFile` methods.

A test for is this has also been included.

Fixes #2799


marklundin:

marklundin:
> @marklundin please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

kpal81xd (approved):

.changeset/tame-garlics-shop.md
```diff
@@ -0,0 +1,2 @@
+---
```
jakebailey, line 1: You shouldn't use an empty changeset here; this is a change that requires a package release with a bump.

jakebailey (changes requested):
See previous comment; this isn't a dev only change and requires a proper versions changset and description.

.changeset/tame-garlics-shop.md
marklundin: Yep thanks @jakebailey. Have updated now

packages/typescript-vfs/src/index.ts
```diff
@@ -99,6 +100,13 @@ export function createVirtualTypeScriptEnvironment(
 
       updateFile(newSourceFile)
     },
+    deleteFile(fileName) {
+      const sourceFile = languageService.getProgram()!.getSourceFile(fileName)
+      if (!sourceFile) {
+        throw new Error("Did not find a source file for " + fileName)
+      }
+      deleteFile(sourceFile)
```
jakebailey: Not sure either way, but does it make sense to throw on delete of a file that doesn't exist? Or should this just be a noop? The canonical implementation in TS itself catches errors, so never observes Node's throws for a nonexistent file.

marklundin: I don't have a strong opinion either way. For updateFile it make sense but a nooop here is probably acceptable

jakebailey: Yeah, I would probably make this operation not throw.

marklundin: Removed 👍

jakebailey: Thanks!

jakebailey (approved):

PR #3175, opened 2024-07-08
Bump to 5.5 and update compiler options

jakebailey:


packages/sandbox/src/compilerOptions.ts
```diff
@@ -85,7 +85,7 @@ export const getCompilerOptionsFromParams = (
       let toSet = undefined
       if (val === "true" && playgroundDefaults[key] !== true) {
         toSet = true
-      } else if (val === "false" && playgroundDefaults[key] !== false) {
+      } else if (val === "false" && (playgroundDefaults[key] as any) !== false) { // TODO(jakebailey): remove as any, check undefined above
```
jakebailey, line 88: I think this may have been hiding a bug, but I don't want to change this PR to modify the behavior in this PR.

packages/tsconfig-reference/copy/en/options/isolatedDeclarations.md
```diff
@@ -0,0 +1,6 @@
+---
+display: "isolatedDeclarations"
+oneline: "Does something"
```
jakebailey: These need to be updated.

andrewbranch (approved):

PR #3178, opened 2024-07-09
noUnusedParameters: add note about underscore prefix exception

jamonholmgren:
In the [release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals) for noUnusedParameters, there's a note about prefixing the parameter name with an underscore to bypass this error.

This PR adds that same note and example to the [noUnusedParameters](https://www.typescriptlang.org/tsconfig/#noUnusedParameters) documentation.

(thanks to @lindboe for noticing this and pointing me to the release docs!)

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
+++ b/packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
@@ -13,3 +13,12 @@ const createDefaultKeyboard = (modelID: number) => {
   return { type: "keyboard", modelID: defaultModelID };
 };
 ```
+
+Parameters declaration with names starting with an underscore _ are exempt from the unused parameter checking. e.g.:
+
+```ts twoslash
+function returnNull(_a) {
+  // OK
+  return null;
+}
+```
--- a/packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
+++ b/packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
@@ -14,7 +14,7 @@ const createDefaultKeyboard = (modelID: number) => {
 };
 ```
 
-Parameters declaration with names starting with an underscore _ are exempt from the unused parameter checking. e.g.:
+Parameters declaration with names starting with an underscore (`_`) are exempt from the unused parameter checking. e.g.:
 
 ```ts twoslash
 // @noUnusedParameters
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
```diff
@@ -13,3 +13,12 @@ const createDefaultKeyboard = (modelID: number) => {
   return { type: "keyboard", modelID: defaultModelID };
 };
 ```
+
+Parameters declaration with names starting with an underscore _ are exempt from the unused parameter checking. e.g.:
+
+```ts twoslash
+function returnNull(_a) {
```
jakebailey: This needs a type annotation to prevent the error: https://victorious-plant-05c166c10-3178.centralus.5.azurestaticapps.net/tsconfig/#noUnusedParameters

jakebailey: Also, set `// @noUnusedParameters` here. Best to copy the above example and just rename the parameter, actually (but without the `@errors`)

jamonholmgren: Will do!

jamonholmgren:
Finally got back to this. @jakebailey thanks for the feedback, should be ready to go...hopefully!

jamonholmgren:
Actually, now that I think about it ... isn't the underscore enough? Do we really need the `// @noUnusedParameters` annotation? The underscore should make it exempt from that check.

packages/tsconfig-reference/copy/en/options/noUnusedParameters.md
```diff
@@ -13,3 +13,12 @@ const createDefaultKeyboard = (modelID: number) => {
   return { type: "keyboard", modelID: defaultModelID };
 };
 ```
+
+Parameters declaration with names starting with an underscore _ are exempt from the unused parameter checking. e.g.:
```
jakebailey:
```suggestion
Parameters declaration with names starting with an underscore (`_`) are exempt from the unused parameter checking. e.g.:
```

jakebailey (changes requested):
Minor nit, given `_` is special in markdown.

jamonholmgren:
Just remembered this issue and committed that change, @jakebailey. Anything else?

jakebailey (approved):

jamonholmgren:
Thank you!!

PR #3183, opened 2024-07-16
Explanation on how TypeScript handles conflicts between interfaces and intersection types

### LINKED ISSUE #3071: Should elaborate more on the way conflicts are handled in Interfaces vs. Intersections
**Page URL:** https://www.typescriptlang.org/docs/handbook/2/objects.html#interfaces-vs-intersections

**Issue:** It says the following:

> The principal difference between the two is how conflicts are handled, and that difference is typically one of the main reasons why you’d pick one over the other between an interface and a type alias of an intersection type.

But then it doesn't elaborate on how the conflicts are handled. A link to any article explaining it would be ok.

merintheressa:
Explanation with examples on the way conflicts are handled in Interfaces vs. Intersections

Closes https://github.com/microsoft/TypeScript-Website/issues/3071.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Object Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Object Types.md
@@ -591,7 +591,7 @@ In the case of intersection types, properties with different types will be merge
 
 For example, the following code will throw an error because the properties are incompatible:
 
-```
+```ts
 interface Person {
   name: string;
 }
@@ -603,7 +603,7 @@ interface Person {
 
 In contrast, the following code will compile, but it results in a `never` type:
 
-```
+```ts twoslash
 interface Person1 {
   name: string;
 }
@@ -613,6 +613,10 @@ interface Person2 {
 }
 
 type Staff = Person1 & Person2
+
+declare const staffer: Staff;
+staffer.name;
+//       ^?
 ```
 In this case, Staff would require the name property to be both a string and a number, which results in property being of type `never`.
 
--- a/packages/documentation/copy/en/handbook-v2/Object Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Object Types.md
@@ -601,7 +601,7 @@ interface Person {
 }
 ```
 
-In contrast, the following code will compile, but it results in an invalid type:
+In contrast, the following code will compile, but it results in a `never` type:
 
 ```
 type Person = {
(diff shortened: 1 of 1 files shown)


merintheressa:

packages/documentation/copy/en/handbook-v2/Object Types.md
```diff
@@ -585,10 +585,34 @@ We just looked at two ways to combine types which are similar, but are actually
...
+type Person = {
+  name: string;
+};
 
-TODO -->
+type Staff = Person & {
+  name: number;
+};
+```
+In this case, Staff would require the name property to be both a string and a number, which is impossible and thus problematic.
```
jakebailey: I would avoid saying "problematic" or saying "invalid"; the type is "valid", it's just that the prop is `never`.

merintheressa: Added a commit to clarify this.

packages/documentation/copy/en/handbook-v2/Object Types.md
```diff
@@ -585,10 +585,34 @@ We just looked at two ways to combine types which are similar, but are actually
...
+
+```
+type Person = {
+  name: string;
+};
 
-TODO -->
+type Staff = Person & {
+  name: number;
+};
```
SaeWooKKang: Oh, that's cool. When I first read the example, I wondered if it was a matter of `type` and `interface`. But since it's not a matter of keywords, I think it will help to control variables by writing it in `interface` as follows.

```ts
interface Person1 {
  name: string;
}

interface Person2 {
  name: number;
}

type Staff = Person1 & Person2
```

merintheressa: That's a good idea 👍

packages/documentation/copy/en/handbook-v2/Object Types.md
```diff
@@ -585,10 +585,36 @@ We just looked at two ways to combine types which are similar, but are actually
...
+}
 
-TODO -->
+interface Person2 {
+  name: number;
+}
+
+type Staff = Person1 & Person2
+```
+In this case, Staff would require the name property to be both a string and a number, which results in property being of type `never`.
```
jakebailey, line 621: I think this example would benefit from a twoslash comment / example showing that the prop type was changed, like

````md
```ts twoslash
interface Person1 {
  name: string;
}

interface Person2 {
  name: number;
}

type Staff = Person1 & Person2

declare const staffer: Staff;

staffer.name;
//       ^?
```
````

Additionally, both code blocks should be marked as `ts` code.

jakebailey (approved):

PR #3188, opened 2024-07-24
Update Modules.md

izadi-abraham:
The import was incorrectly importing `math.js` instead of `math.ts` file.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Modules.md
+++ b/packages/documentation/copy/en/handbook-v2/Modules.md
@@ -174,7 +174,7 @@ You can import a file and _not_ include any variables into your current module v
 export var pi = 3.14;
 // ---cut---
 // @filename: app.ts
-import "./maths.js";
+import "./maths.ts";
 
 console.log("3.14");
 ```


jakebailey:
This code is correct as-is. TypeScript does not rewrite import paths so this code would be unlikely to work if emitted.

Note the other imports below that code block which also use the `.js` extension.

Consider taking a look at https://www.typescriptlang.org/docs/handbook/modules/theory.html.

izadi-abraham:
Ok, so then this desctription in the line 182 should be modified and be `math.js`, because it is referring to this code, right?:<br>
image

jakebailey:
I'm not seeing the inconsistency; a `.ts` file compiles to a `.js` file, the sentence is just talking about the code as it was written in originally source

izadi-abraham:
> I'm not seeing the inconsistency; a `.ts` file compiles to a `.js` file, the sentence is just talking about the code as it was written in originally source

You are right, it was all my own confusion. Sorry for the inconvenience.

PR #3190, opened 2024-07-27
Update allowUnreachableCode.md

valerii15298:
Fix/Update grammar
The word `provably` is deprecated and was only used in the Middle English period (1150—1500) https://www.oed.com/dictionary/provably_adj

This PR substitutes the obsolete word `provably` with `probably` which serves a better purpose according to the context.


jakebailey:
This word is most definitely not obsolete in the parlance of programming languages; it's specifically intended to indicate that it we can only make determinations about "provable" things.

"Probably" is far too weak of a replacement; it indicates that we may not be correct about what we report.

PR #3191, opened 2024-07-27
ATA: `node:` module resolution

Renegade334:
- There are certain modules that Node doesn't expose without the `node:` prefix, to avoid clashing with existing packages. These are now resolved.
- Ensure that `node:` is only handled if it's a prefix (and not `"dnode:ns"` _etc._)
- Add `inspector/promises` to the module list.

### DIFF
--- a/.changeset/plenty-deers-visit.md
+++ b/.changeset/plenty-deers-visit.md
@@ -2,4 +2,4 @@
 "@typescript/ata": patch
 ---
 
-Improve `node:` module resolution
+Always treat `node:` modules as Node, expand list of known Node modules
--- a/.changeset/plenty-deers-visit.md
+++ b/.changeset/plenty-deers-visit.md
@@ -1,0 +1,5 @@
+---
+"@typescript/ata": patch
+---
+
+Improve `node:` module resolution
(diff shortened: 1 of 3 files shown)


packages/ata/src/edgeCases.ts
```diff
@@ -57,7 +58,15 @@ export const mapModuleNameToModule = (moduleSpecifier: string) => {
     "zlib",
   ]
 
-  if (builtInNodeMods.includes(moduleSpecifier.replace("node:", ""))) {
+  const builtInNodeModsWithMandatoryPrefix = [
```
jakebailey: Every new module node adds is going to be locked behind the node: scheme; I think it'd be better to hardcode the current list of modules that don't need the prefix, then always load of the prefix is present in an import.

.changeset/plenty-deers-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@typescript/ata": patch
+---
+
+Improve `node:` module resolution
```
jakebailey:
```suggestion
Always treat `node:` modules as Node, expand list of known Node modules
```

jakebailey (approved):
LGTM after the changeset tweak to be more descriptive.

amethystens (approved):

PR #3192, opened 2024-07-27
Add missing line break between Privacy and Terms of Use links in footer

nevalenti:
The Privacy and Terms of Use links are joined together
image


packages/typescriptlang-org/src/components/layout/SiteFooter.tsx
```diff
@@ -303,6 +303,7 @@ export const SiteFooter = (props: Props) => {
             >
             { lang === "ko" ? "개인정보처리방침 및 위치정보이용약관" : "Privacy"}
             </a>
+            <br />
```
jakebailey, line 306: I would just add a space, not a line break.

PR #3197, opened 2024-08-03
Adjusted description of the Partial type

StockerGaming:
Subsets of a given type cannot have less properties than the original type (which is not the case with partial types), so the original description was inconsistent with established type operations


jakebailey:
Honestly, I'd just remove the second sentence. The first sentence to me fully explains the utility.

PR #3201, opened 2024-08-13
Fix a typo in Template Literal Types.md

artemious7:
Fix a typo


jakebailey:
This is not a typo; the types "can't be found" because they are ones that are only built-in to the compiler. They are written like `type Uppercase<T> = intrinsic`.

Issue #3204, opened 2024-08-17
PDFs: support PDFs for the website


a0m0rajab:
I did not find a better place to share this (would appreciate to point me to where should I publish it if it's not the correct place) 

I was trying to find the PDF of the handbook for couple of reasons: 
- Offline reading
- Reading from ereader
- Focus reading (by using ereader to not get distracted)

Yet I found that it was removed from the website for simplicity reasons. Is there any new decision about this? 

Related issue: https://github.com/microsoft/TypeScript-Website/pull/2619#issuecomment-1331128079

a0m0rajab:
Note in the comment it's suggested that the print function would be a good usage, yet it does not show the full text (which was the same issue with playwright) 

Here is the related page: https://www.typescriptlang.org/docs/handbook/2/modules.html

The print generated from the CTRL + P (on chrome browser, Version 127.0.6533.120 (Official Build) (64-bit), windows) 

![image](https://github.com/user-attachments/assets/4491926e-944c-4725-b40b-d25316d2d868)

The missing text: The JavaScript specification declares that any JavaScript files without an import declaration, export, or top-level await should be considered a script and not a module.

The docs: 

![image](https://github.com/user-attachments/assets/fb554b46-1a00-4432-9fe9-62e55d9e0587)

I think that search docs and some other tags should not be visible on the print document.

a0m0rajab:
Related issue: https://github.com/microsoft/TypeScript-Website/issues/620

a0m0rajab:
Workaround: I did remove the navbar, and extra information from the page which helped to have a better look for the printing and get the text here is the result: 

After: 
![image](https://github.com/user-attachments/assets/1bcc32b2-192a-46dc-8c58-7f8b0a0c902e)
![image](https://github.com/user-attachments/assets/24c17991-3c19-4f1c-8534-67ed3223002e)


Before: 
![image](https://github.com/user-attachments/assets/2e5481d1-6019-4eaa-9253-9ce49eaa362a)
![image](https://github.com/user-attachments/assets/afd925eb-a4f9-43b2-a83f-bb35a1e5d0ab)

a0m0rajab:
Javascript code to remove the extra elements from the webpage: 
```js
document.querySelectorAll("header, #page-helpful-popup, #site-footer, .whitespace-tight.raised").forEach(e => {
    e.remove()
})
```

a0m0rajab:
Using print query and break points would improve the experience: 

[...]

PR #3205, opened 2024-08-17
feat: add print query to the website

a0m0rajab:
This PR adds print query to the website and show a minimum example of using pagebreak feature. 

The no-print class hide the element on print, and the pagebreak add new page before the related div. 
Here is the first issue about this: https://github.com/microsoft/TypeScript-Website/issues/3204

This PR improve printing a webpage and hide the unnecessary details here is an example: 

| Before| After |
|--------|--------|
| image  | image |
| image | image | 

Page breaks should be added and tested manually which requires a lot of work that I decided to do only if this approach were approved.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Narrowing.md
+++ b/packages/documentation/copy/en/handbook-v2/Narrowing.md
@@ -62,8 +62,6 @@ function padLeft(padding: number | string, input: string): string {
 
 There are a couple of different constructs TypeScript understands for narrowing.
 
-<div class="pagebreak"></div>
-
 ## `typeof` type guards
 
 As we've seen, JavaScript supports a `typeof` operator which can give very basic information about the type of values we have at runtime.
@@ -165,8 +163,6 @@ This at least prevents us from dreaded errors when we run our code like:
 TypeError: null is not iterable
 ```
 
-<div class="pagebreak"></div>
-
 Keep in mind though that truthiness checking on primitives can often be error prone.
 As an example, consider a different attempt at writing `printAll`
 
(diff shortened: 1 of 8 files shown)


a0m0rajab:

a0m0rajab:
@jakebailey hey, is there anything I am missing in this pr? I would love to get your thoughts about it

jakebailey:
It's not clear to me how this scales; surely other pages would need this? These are explicit page breaks; how would we keep them in the right place and not forget some or add some where they aren't needed?

It just feels like there should be a better solution to this problem than manually breaking pages.

a0m0rajab:
Thanks for the answer. If you want, we can split this PR into two different PRs, one to fix the headers that hide the text behind it and the other to discuss the page breaks, 

The main issue here is that we have extra pages and hidden texts because of the header and footers that are available on the website. 

As for scaling I just did a double check and found another property that we can use for the code, I will check it and ping you.

a0m0rajab:
For future reference I need to check: 
https://tympanus.net/codrops/css_reference/break-inside/
https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside
https://dev.to/amruthpillai/avoiding-awkward-element-breaks-in-print-html-5goe
https://stackoverflow.com/questions/907680/css-printing-avoiding-cut-in-half-divs-between-pages

a0m0rajab:
@jakebailey Thanks for your suggestion, we have a better and cleaner code now. 
The code I wrote avoids the breaks for code blocks and works well.

a0m0rajab:
@jakebailey is there anything I need to do for this?

a0m0rajab:
@jakebailey hey, it seems the checks has passed, is there anything I should do from my side?

jakebailey:
This seems fine, though I think this also needs to be applied to the "is this page helpful" thing:

image

PR #3208, opened 2024-08-20
Fix default `maxNodeNodeJsDepth`

louwers:
Verify here that this is correct: https://github.com/microsoft/TypeScript/blob/533ed3d665c4bb0ecb08dd2eae95fe0aa2e1c973/src/testRunner/unittests/tsserver/maxNodeModuleJsDepth.ts#L16


louwers:

louwers:
Apparantly `jsconfig.json` and `tsconfig.json` have different defaults. Closing this!

PR #3221, opened 2024-09-09
Update for TS 5.6

jakebailey:
I had most of this change locally already while testing a 5.7 PR.


packages/tsconfig-reference/tsconfig.tsbuildinfo
jakebailey, line 1: Ah, these files are now generated

packages/documentation/copy/en/project-config/Compiler Options.md
```diff
@@ -54,75 +54,66 @@ tsc app.ts util.ts --target esnext --outfile index.js
...
   <td><p><code>boolean</code></p>
 </td>
 </tr>
-<tr class="option-description even"><td colspan="3">
+<tr class="option-description odd"><td colspan="3">
 <p>Initializes a TypeScript project and creates a tsconfig.json file.</p>
 </td></tr>
 
-<tr class='odd' name='listFilesOnly'>
+<tr class='even' name='listFilesOnly'>
```
DanielRosenwasser, line 75: Not in this PR but one day https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

DanielRosenwasser (approved):

packages/tsconfig-reference/scripts/schema/result/schema.json
```diff
@@ -834,12 +840,13 @@
                       "esnext.decorators",
                       "esnext.object",
                       "esnext.regexp",
+                      "esnext.iterator",
                       "decorators",
                       "decorators.legacy"
                     ]
                   },
                   {
```
navya9singh, line 848: Just for my understanding, could you tell me what these pattern changes mean?

jakebailey: These options are case insensitive, so this is just the absolutely ridiculous regex to verify the union of a bunch of strings

jakebailey: In a better world, the regex syntax would accept syntax like `(?i)`, but JS doesn't have that syntax, it puts it _outside_ the regex expression and a load of consuming tools do not parse out strings like `"/blah/i"`

navya9singh (approved):

PR #3224, opened 2024-09-10
Document `satisfies`

RyanCavanaugh:
ref https://github.com/microsoft/TypeScript/issues/59930

### DIFF
--- a/packages/documentation/copy/en/handbook-v1/Basic Types.md
+++ b/packages/documentation/copy/en/handbook-v1/Basic Types.md
@@ -367,7 +367,7 @@ TypeScript assumes that you, the programmer, have performed any special checks t
 
 Type assertions have two forms.
 
-One is the `as`-syntax:
+One is the `as` operator:
 
 ```ts twoslash
 let someValue: unknown = "this is a string";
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/handbook-v1/Basic Types.md
```diff
@@ -384,7 +384,48 @@ let strLength: number = (<string>someValue).length;
 ```
 
 The two samples are equivalent.
-Using one over the other is mostly a choice of preference; however, when using TypeScript with JSX, only `as`-style assertions are allowed.
+Using one over the other is mostly a choice of preference; however, when using TypeScript with JSX, only `as` assertions are allowed.
+
+## `satisfies` checks
+
+While the `as` operator *asserts* that an expression is of a given type, the `satisfies` operator *checks` the type instead.
```
jakebailey, line 391:
```suggestion
While the `as` operator *asserts* that an expression is of a given type, the `satisfies` operator *checks* the type instead.
```

jakebailey:
This is the v1 handbook; is there a v2 place where this is documented?

RyanCavanaugh:
StarTrekPatrickStewartGIF

devonhk:
Will the `satisfies` operator ever be documented in the reference section? It's only mentioned in old release [notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html).

PR #3225, opened 2024-09-11
The separator of union members is allowed before the first element

adriancuadrado:


packages/documentation/copy/en/handbook-v2/Everyday Types.md
```diff
@@ -252,6 +252,18 @@ printId("202");
 printId({ myID: 22342 });
 ```
 
+> The separator of the union members is allowed before the first element, so you could also write this:
+> ```ts twoslash
+> function printTextOrNumberOrBool(
+>   textOrNumberOrBool:
+>   | string
+>   | number
+>   | boolean
```
jakebailey:
```suggestion
>     | string
>     | number
>     | boolean
```

jakebailey (approved):

PR #3229, opened 2024-09-14
Fixing accessibility bug

navya9singh:
Fixing https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2187524
The PR fixes the incorrect name provided for "Handbook" button as "Handbook Close" Button.

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
+++ b/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
@@ -102,7 +102,7 @@ export const Sidebar = (props: Props) => {
         classes.push("closed")
       }
 
-      const opened = { "aria-expanded": "true", "aria-label": item.title + " close" }
+      const opened = { "aria-expanded": "true", "aria-label": item.title }
       const closed = { "aria-label": item.title + " expand" }
       const aria = hostsSelected ? opened : closed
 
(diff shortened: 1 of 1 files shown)


packages/typescriptlang-org/src/components/layout/Sidebar.tsx
```diff
@@ -102,7 +102,7 @@ export const Sidebar = (props: Props) => {
         classes.push("closed")
       }
 
-      const opened = { "aria-expanded": "true", "aria-label": item.title + " close" }
```
jakebailey, line 105: The naming implies this button can "close" things (because it's opened), the closed variable below has "expand" because it can expand things. Where does the bug come into play?

navya9singh: It mentions that the Screen reader users will get incorrect information about the "Handbook button" as "Handbook Close Button

jakebailey: What happens if we add `aria-expanded: false` to `closed`? Does that provide a distinction in the screen reader?

navya9singh: Yes, so I checked in the office hours, and they said that "aria-label" is not needed here since the button already has a name. Just making sure that "aria-expanded" is correctly set to true or false will be enough for the screen reader to produce the expected result.

jakebailey (approved):

PR #3230, opened 2024-09-14
Fixing accessibility bug

navya9singh:
Fixing https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2188102
This pr adds different landmark names on the same page.


jakebailey (approved):

navya9singh:
Adding `aria-label` instead of `role` actually fixes the issue.

PR #3232, opened 2024-09-17
Fix type on JSX docs

titoBouzout:
I believe this is a typo.  That, or I misunderstand.

### DIFF
--- a/packages/documentation/copy/en/reference/JSX.md
+++ b/packages/documentation/copy/en/reference/JSX.md
@@ -452,7 +452,7 @@ By default, function components must return `JSX.Element | null`. However, this 
 namespace JSX {
     export type ElementType =
         // All the valid lowercase tags
-        keyof IntrinsicAttributes
+        keyof IntrinsicElements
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/JSX.md
```diff
@@ -452,7 +452,7 @@ By default, function components must return `JSX.Element | null`. However, this
 namespace JSX {
     export type ElementType =
         // All the valid lowercase tags
-        keyof IntrinsicAttributes
+        keyof IntrinsicElements
```
jakebailey: I think this is correct, but, this code's syntax is definitely already wrong; this should be:

```ts
export type ElementType =
    | keyof IntrinsicElements
    | ((props: any) => Element)
    | (new (props: any) => ElementClass)
    ;
```

Or something.

titoBouzout: oh right, I missed that

jakebailey (approved):

PR #3235, opened 2024-09-26
Fixing accessibility bug

navya9singh:
Adding unique navigation landmarks. Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2175671

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
+++ b/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
@@ -122,7 +122,7 @@ export const Sidebar = (props: Props) => {
   }
 
   return (
-    <nav aria-label = "sidebar" id="sidebar">
+    <nav aria-label="sidebar" id="sidebar">
       <ul>
         {props.navItems.map(item => <RenderItem key={item.id} item={item} openAllSectionsExceptWhatsNew={props.openAllSectionsExceptWhatsNew} selectedID={props.selectedID} />)}
       </ul>
--- a/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
+++ b/packages/typescriptlang-org/src/components/layout/Sidebar.tsx
@@ -122,7 +122,7 @@ export const Sidebar = (props: Props) => {
   }
 
   return (
-    <nav id="sidebar">
+    <nav aria-label = "sidebar" id="sidebar">
       <ul>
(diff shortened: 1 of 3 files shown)


packages/typescriptlang-org/src/components/layout/Sidebar.tsx
```diff
@@ -122,7 +122,7 @@ export const Sidebar = (props: Props) => {
   }
 
   return (
-    <nav id="sidebar">
+    <nav aria-label = "sidebar" id="sidebar">
```
jakebailey:
```suggestion
    <nav aria-label="sidebar" id="sidebar">
```

packages/typescriptlang-org/src/components/layout/TopNav.tsx
```diff
@@ -98,7 +98,7 @@ export const SiteNav = (props: Props) => {
             <span className="hide-small">TypeScript</span>
           </IntlLink>
 
-          <nav role="navigation">
+          <nav aria-label = "top menu">
```
jakebailey:
```suggestion
          <nav aria-label="top menu">
```

packages/typescriptlang-org/src/templates/documentation.tsx
```diff
@@ -159,7 +159,7 @@ const HandbookTemplate: React.FC<Props> = (props) => {
             </div>
             {showSidebar &&
               <aside className="handbook-toc">
-                <nav className={deprecationURL ? "deprecated" : ""}>
+                <nav aria-label = "table of contents">
```
jakebailey:
```suggestion
                <nav aria-label="table of contents">
```

packages/typescriptlang-org/src/templates/documentation.tsx
```diff
@@ -159,7 +159,7 @@ const HandbookTemplate: React.FC<Props> = (props) => {
             </div>
             {showSidebar &&
               <aside className="handbook-toc">
-                <nav className={deprecationURL ? "deprecated" : ""}>
```
jakebailey, line 162: Does removing this have any effect? It seems like what this is saying is that we should hide the ToC on deprecated pages?

navya9singh: I don't really see a difference as such, and I'm not sure if we should actually hide the ToC on deprecated pages, but I can add it back if you think that's needed?

jakebailey (approved):

PR #3236, opened 2024-09-27
Fixing accessibility bug

navya9singh:
This pr fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2175667 and https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2175655. I've tested it out in both scenarios and this seems to solve the issue.

### DIFF
--- a/packages/typescriptlang-org/src/templates/documentation.scss
+++ b/packages/typescriptlang-org/src/templates/documentation.scss
@@ -45,7 +45,12 @@
 
   article {
     display: flex;
+    flex-direction: row;
     width: 100%;
+
+    @media (max-width: 999px) {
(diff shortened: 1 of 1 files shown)


jakebailey:
Hiding the links on the right is designed behavior, as far as I know. Old:

image

New:

image

Is there no way to satisfy the report another way? I know some sites will instead put the aside on top or something...

navya9singh:
> Is there no way to satisfy the report another way? I know some sites will instead put the aside on top or something...

Added some changes to make it put aside on top. How does this look?

jakebailey (approved):
Seems good!

PR #3240, opened 2024-10-04
Docs: `noImplicitReturns` is `true` when `strictNullChecks` is `true`

alexilyaev:
## Why?

Looking at the docs for [noImplicitReturns](https://www.typescriptlang.org/tsconfig/#noImplicitReturns), it seems to yield the same error whether the option is `true` or `false`.

Searching further I found https://github.com/microsoft/TypeScript/issues/51312 and tested with `strictNullChecks` relation.
Indeed, when setting `strictNullChecks: true`, it's as if we set `noImplicitReturns: true`.

So, I suggest adding this to the docs to avoid confusion.


alexilyaev:
Is anything blocking this from being merged?

jakebailey:
I'm a bit wary of this change, because there is no explicit code in the compiler which outlines this implication, unlike some of the other inferred options.

alexilyaev:
Well, I'm just pointing out that right now it's confusing that `noImplicitReturns: false` has no effect when `strictNullChecks: true`. And there's no way to understand that from the docs.
So, I would expect the connection between them to be explicit in the code and in the docs.
Or, if it's not, then it's probably a bug.

PR #3243, opened 2024-10-05
Explain keyof before it first appears in a snippet

natepmay:
In the [Using Type Parameters in Generic Constraints](https://www.typescriptlang.org/docs/handbook/2/generics.html#using-type-parameters-in-generic-constraints) section of the Generics chapter of the handbook, an example is introduced that uses `keyof`. However, the `keyof` operator isn't introduced until the next chapter. As a result, I was confused by the example as I was working through the handbook in sequence.

I've kept the example but briefly introduced `keyof` beforehand to make the example comprehensible.


natepmay:

PR #3248, opened 2024-10-08
updated the type name

patichandana:
although the variable and the type having the same name works, to avoid confusion among the users, I've changed the type name


packages/documentation/copy/en/reference/JSX.md
```diff
@@ -41,7 +41,7 @@ You can specify this mode using either the [`jsx`](/tsconfig#jsx) command line f
 Recall how to write a type assertion:
 
 ```ts
-const foo = <foo>bar;
+const foo = <Foo>bar;
 ```
 
 This asserts the variable `bar` to have the type `foo`.
```
jakebailey: The text down here is still referring to the type by the old name.

jakebailey (approved):

PR #3253, opened 2024-10-15
Fixing formatting bug

navya9singh:
Fixing https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2201000


jakebailey:
I don't know that this will do anything; IIRC the website build formats all code on the page, so the line break shouldn't persist.

navya9singh:
> I don't know that this will do anything; IIRC the website build formats all code on the page, so the line break shouldn't persist.

Yeah, I wanted to check if it was giving a different output here. Seems like the the code mentioned in the bug is generated from this markdown file. So I'm not sure how to fix this issue now.

PR #3255, opened 2024-10-15
Fixing Accessibility bug

navya9singh:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2201745


jakebailey (approved):
I can't really reproduce the problem, but the page seems ok with the change so if you find that it fixes the issue, seems fine to me.

PR #3256, opened 2024-10-16
Fixing accessibility bug

navya9singh:
Removing code samples from footers to fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2187474


packages/typescriptlang-org/src/components/layout/SiteFooter.tsx
```diff
@@ -82,10 +81,6 @@ const useTypeScriptLinks = [
     title: "TSConfig Ref",
     url: "/tsconfig",
   },
-  {
-    title: "Code Samples",
-    url: "/play/#show-examples",
-  },
```
jakebailey: Can we just remove the popover and leave it as a link here? The actual playground page has a version of this.

navya9singh: Sure, added it back.

jakebailey (approved):

PR #3257, opened 2024-10-16
Patch gatsby-remark-shiki-twoslash to add in tabindex

jakebailey:
Modern shiki does this automatically, but we're using an old version without it. Patch it in for now until we can wholesale upgrade

closes #3253


navya9singh (approved):

navya9singh:
This fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2201000, tested it! 🎉

PR #3258, opened 2024-10-17
Fixing accessibility bug

navya9singh:
Removes anchor for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2201723

### DIFF
--- a/packages/typescriptlang-org/src/templates/markdown.scss
+++ b/packages/typescriptlang-org/src/templates/markdown.scss
@@ -20,6 +20,7 @@
     &[id]::before {
       content: "";
       display: block;
+      visibility: hidden;
       height: 30px;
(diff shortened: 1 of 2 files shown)


packages/typescriptlang-org/src/templates/markdown.scss
```diff
@@ -20,7 +20,6 @@
     &[id]::before {
       content: "";
       display: block;
-      visibility: hidden;
```
jakebailey: Is this still required?

jakebailey (approved):

packages/typescriptlang-org/src/templates/markdown.scss
navya9singh: It's not. Changin it back

PR #3265, opened 2024-10-29
Update triple-slash directive docs for preserve=true

### LINKED ISSUE #3261: Documentation for `/// <reference types="..." />` is outdated/incorrect
**Page URL:** https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-

**Issue:**

The page states:
> For declaration files generated during compilation, the compiler will automatically add `/// <reference types="..." />` for you; A `/// <reference types="..." />` in a generated declaration file is added _if and only if_ the resulting file uses any declarations from the referenced package.

This is no longer correct as of TypeScript 5.5; see https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#simplified-reference-directive-declaration-emit.

The documentation should be updated with the new behavior, including coverage of `preserve="true"`.

jakebailey:
Fixes #3261

### DIFF
--- a/packages/documentation/copy/en/reference/Triple-Slash Directives.md
+++ b/packages/documentation/copy/en/reference/Triple-Slash Directives.md
@@ -13,7 +13,7 @@ Triple-slash directives are **only** valid at the top of their containing file.
 A triple-slash directive can only be preceded by single or multi-line comments, including other triple-slash directives.
 If they are encountered following a statement or a declaration they are treated as regular single-line comments, and hold no special meaning.
 
-As of TypeScript 5.5, the compiler does not generate reference directives, and does not _not_ emit handwritten triple-slash directives to output files unless those directives are marked as [`preserve="true"`](#preservetrue).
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Triple-Slash Directives.md
```diff
@@ -13,6 +13,8 @@ Triple-slash directives are **only** valid at the top of their containing file.
 A triple-slash directive can only be preceded by single or multi-line comments, including other triple-slash directives.
 If they are encountered following a statement or a declaration they are treated as regular single-line comments, and hold no special meaning.
 
+As of TypeScript 5.5, the compiler does not generate reference directives, and does not _not_ emit handwritten triple-slash directives to output files unless those directives are marked as [`preserve="true"`](#preservetrue).
```
DanielRosenwasser:
```suggestion
As of TypeScript 5.5, the compiler does not generate reference directives, and does _not_ emit handwritten triple-slash directives to output files unless those directives are marked as [`preserve="true"`](#preservetrue).
```

PR #3266, opened 2024-11-01
Replace type aliases with interfaces in "Handbook -> More On Functions"

sebastiancarlos:
This change aligns with the recommendation in "Handbook -> Everyday Types":

> For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type.

Based on this guidance, all examples in the documentation should use interfaces where possible.


sebastiancarlos:

PR #3274, opened 2024-11-12
Fix: wrong typescript announcement link for release version

arijit192:
**Issue:**
The link it is redirecting to is wrong.

image
is redirecting to
image
Here the minor version number is wrong.

**Fix:**
Fixed this issue by changing the wrong 'rc' variable to 'stable' to build the url from stable version instead of the release candidate one.


arijit192:
Hey @jakebailey can you please have a look on it? I think this is a simple yet required fix.

PR #3277, opened 2024-11-20
Update TS for the New Programmer.md

Ibmidquantum:


jakebailey (changes requested):
Sorry, but this doesn't actually do anything in practice; these empty lines are ignored by the tooling.

PR #3282, opened 2024-12-02
Fix z-index conflict between aside and sidebar in mobile view

aqibshabir:
This PR fixes a CSS z-index conflict in mobile view where the aside is overlapping the sidebar. 
- The aside had a z-index of 1000, which caused it to appear above the sidebar (z-index: 99).
- Changed the aside z-index to 98, ensuring it stays behind the sidebar in mobile view.

### DIFF
--- a/packages/typescriptlang-org/src/templates/documentation.scss
+++ b/packages/typescriptlang-org/src/templates/documentation.scss
@@ -1,13 +1,11 @@
 @import "../style/globals.scss";
 
 #beta {
-  background: repeating-linear-gradient(
-    45deg,
-    #9b9a4e,
-    #9b9a4e 10px,
-    #6e6a33 10px,
-    #6e6a33 20px
-  );
+  background: repeating-linear-gradient(45deg,
+      #9b9a4e,
+      #9b9a4e 10px,
+      #6e6a33 10px,
+      #6e6a33 20px);
   background-color: "#c63131";
   text-align: center;
   color: white;
@@ -32,7 +30,7 @@
   }
 
   // Page title
-  & > h1 {
+  &>h1 {
     font-size: 3.5rem;
     line-height: 3.5rem;
     font-weight: 400;
@@ -136,6 +134,7 @@
     }
 
     .like-dislike-subnav {
+
       #like-button,
       #dislike-button {
         cursor: pointer;
@@ -478,6 +477,6 @@
   overflow: auto;
   max-height: 80vh;
   ul {
-    padding-left: 0.5em !important;
+    padding-left: .5em !important;
   }
 }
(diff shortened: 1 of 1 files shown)


aqibshabir:

packages/typescriptlang-org/src/templates/documentation.scss
```diff
@@ -1,11 +1,13 @@
 @import "../style/globals.scss";
 
 #beta {
-  background: repeating-linear-gradient(45deg,
-      #9b9a4e,
-      #9b9a4e 10px,
-      #6e6a33 10px,
-      #6e6a33 20px);
+  background: repeating-linear-gradient(
```
jakebailey: Please undo any unrelated formatting changes.

aqibshabir: I’ve reverted the unrelated formatting changes and kept only the necessary updates.

jakebailey:
This does seem like an improvement but https://victorious-plant-05c166c10-3282.centralus.5.azurestaticapps.net/docs/handbook/declaration-files/templates/module-class-d-ts.html shows that the sidebar still overlaps content 😞

aqibshabir:
> This does seem like an improvement but https://victorious-plant-05c166c10-3282.centralus.5.azurestaticapps.net/docs/handbook/declaration-files/templates/module-class-d-ts.html shows that the sidebar still overlaps content 😞

The solution I provided (almost 2 months ago) deals with the z-index conflict with the content appearing within the sidebar when the sidebar is active.

Having taken a look at the link provided, I'm slightly puzzled. You mentioned that the sidebar overlaps the content - isn't that the desired outcome when you select the hamburger menu - for the sidebar to be on top of the content?

Could you clarify what the expected behavior should be? I want to ensure we’re aligned and address any remaining concerns effectively. I appreciate your follow-up on this.

jakebailey:
Screenshot_20250126-023043

What I see is this on loading that page

jakebailey:
Now, this issue is certainly unrelated to your change since I'm pretty sure this is a regression introduced in some other change, but it is a little unfortunate to fix the z-index and then it still covers content 😄

jakebailey:
This PR affects code added in https://github.com/microsoft/TypeScript-Website/pull/3236; I don't see it breaking the accessibility fix in my test, so that's good.

jakebailey (approved):

aqibshabir:
> Now, this issue is certainly unrelated to your change since I'm pretty sure this is a regression introduced in some other change, but it is a little unfortunate to fix the z-index and then it still covers content 😄

Thank you for pointing that out! I did notice the issue, and I agree—it’s not ideal to leave it as is. To address this, I’ll work on making the site more mobile-friendly and will submit a pull request in the coming days with a proposed improvement. Looking forward to your feedback once it’s ready! 😊

PR #3286, opened 2024-12-16
Resolve transitive dependencies based on package.json

seanoch:
When the newly added `resolveDependenciesFromPackageJson` flag is set to true, ATA will resolve the versions of any transitive dependency according to the `package.json` file of the same module (unless any `// types:` comment is specified).

### DIFF
--- a/packages/ata/src/index.ts
+++ b/packages/ata/src/index.ts
@@ -29,9 +29,22 @@ export interface ATABootstrapConfig {
   fetcher?: typeof fetch
   /** If you need a custom logger instead of the console global */
   logger?: Logger
+  /** Wheather to use package.json as the source of truth for transitive dependencies' versions */
+  usePackageJson?: boolean
 }
 
-type ModuleMeta = { state: "loading" }
+type PackageJsonDependencies = {
+  [packageName: string]: string;
+};
+
+type PackageJson = { 
+  dependencies?: PackageJsonDependencies;
+  devDependencies?: PackageJsonDependencies;
+  peerDependencies?: PackageJsonDependencies;
+  optionalDependencies? :PackageJsonDependencies;
+}
+
+type ModuleMeta = { state: "loading", typesPackageJson?: PackageJson }
 
 /**
  * The function which starts up type acquisition,
@@ -60,8 +73,8 @@ export const setupTypeAcquisition = (config: ATABootstrapConfig) => {
     })
   }
 
-  async function resolveDeps(initialSourceFile: string, depth: number) {
-    const depsToGet = getNewDependencies(config, moduleMap, initialSourceFile)
+  async function resolveDeps(initialSourceFile: string, depth: number, parentPackageJson?: PackageJson) {
+    const depsToGet = getNewDependencies(config, moduleMap, initialSourceFile, parentPackageJson)
 
     // Make it so it won't get re-downloaded
     depsToGet.forEach(dep => moduleMap.set(dep.module, { state: "loading" }))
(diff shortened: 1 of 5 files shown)


seanoch:
@microsoft-github-policy-service agree company="wix"

packages/ata/src/index.ts
```diff
@@ -29,9 +29,22 @@ export interface ATABootstrapConfig {
   fetcher?: typeof fetch
   /** If you need a custom logger instead of the console global */
   logger?: Logger
+  /** Wheather to use package.json as the source of truth for transitive dependencies' versions */
```
tomenden:
```suggestion
  /** Whether to use package.json as the source of truth for transitive dependencies' versions */
```

packages/ata/src/index.ts
```diff
@@ -124,7 +143,11 @@ export const setupTypeAcquisition = (config: ATABootstrapConfig) => {
           }
 
           // Recurse through deps
-          await resolveDeps(dtsCode, depth + 1)
+          let typesPackageJson
+          if (config.usePackageJson){
+            typesPackageJson = moduleMap.get(dts.moduleName)?.typesPackageJson
+          }
```
tomenden:
```suggestion
          const typesPackageJson = config.usePackageJson && moduleMap.get(dts.moduleName)?.typesPackageJson
```

packages/ata/src/index.ts
```diff
@@ -29,9 +29,22 @@ export interface ATABootstrapConfig {
   fetcher?: typeof fetch
   /** If you need a custom logger instead of the console global */
   logger?: Logger
+  /** Wheather to use package.json as the source of truth for transitive dependencies' versions */
+  usePackageJson?: boolean
```
tomenden: Maybe a name like `resolveDependenciesFromPackageJson` is a bit more descriptive, although this is actually the strategy to resolve dependencies _of dependencies_.

tomenden (approved):
See small comments regarding naming.
I suggest adding a bit more context in the PR description to explain the motivation. I think an example will help the typescript team understand what you did here

jakebailey:
To be clear, is the intent here to ensure that all dependencies are loaded transitively?

Should this flag be set for the website? Should it be disabled at all?

seanoch:
@jakebailey The intent is to allow resolution of transitive (indirect) dependencies according to their module's package.json file. Before this change, any indirect dependency gets resolved to latest, since the only way to control versioning is via the `// types:` comment, which is controllable only in the first input file.

For example, if the input file says `import a from 'PackageA'`, and some of `PackageA`'s types imports from `PackageB` -
* Before this change, any d.ts from `PackageB` will be resolved to `PackageB@latest`
* After this change, any d.ts from `PackageB` will be resolved to the version of `PackageB` that is specified in `PackageA`'s package.json.

seanoch:
@jakebailey Hi, is any additional explanation needed before a review?

seanoch:
@jakebailey Hi again, any updates here? Let me know if any additional explanation is needed :)

jakebailey:
The main thing I would like to see is this flag be enabled for ATA on the playground, then be tested there (which I can deploy from this PR).

seanoch:
@jakebailey Hi Jake, I enabled the flag on the playground. For better clarification, this is the new functionality:
1. user added import from 'sean-test-2'
2. 'sean-test-2' has a dependency on 'sean-test-1@1.1.0'
3. ATA used to fetch 'sean-test-1@latest'. Now it fetches 'sean-test-1@1.1.0'

As for tests - I didn't find any ATA related tests under the `playground` package, can you guide me to where you want to add tests?
Screenshot 2025-01-26 at 10 53 52
Screenshot 2025-01-26 at 10 54 20

jakebailey:
There aren't any tests, really, not that I'm aware of.

jakebailey:
TBH I feel like the website would be better served by https://www.npmjs.com/package/@vscode/ts-package-manager / https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/filesystems/autoInstallerFs.ts which actually uses a package manager to do resolution and place files into the VFS, but that's a larger undertaking. I'd just worry that this PR is potentially wrong for things like version constraints (not that the current code is particularly great on that front either).

If you weren't trying to plug this into the playground originally, why were you updating the ATA package? Where are you using it?

seanoch:
We aim to use ATA in our monaco-based code editor. I'm aware of the fact that the solution is far from perfect but it's more accurate then the existing resolution to latest versions of transitive dependencies

PR #3298, opened 2025-01-10
Rename for consistent example snippet

lazyrainynight:
`User` has `admin` property in other example snippets so as in the playground

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/More on Functions.md
+++ b/packages/documentation/copy/en/handbook-v2/More on Functions.md
@@ -601,7 +601,7 @@ This pattern is common with callback-style APIs, where another object typically 
 // @errors: 7041 7017
 interface User {
   id: number;
-  isAdmin: boolean;
+  admin: boolean;
 }
 declare const getDB: () => DB;
 // ---cut---


packages/documentation/copy/en/handbook-v2/More on Functions.md
```diff
@@ -601,7 +601,7 @@ This pattern is common with callback-style APIs, where another object typically
 // @errors: 7041 7017
```
jakebailey, line 601: It's possible these errors may need to go away?

lazyrainynight: The current error showing on the page is not the point. The error is supposed to be there when using the arrow function
but if the error doesn't go away even when you change it to normal function statement because `User` doesn't have a property named 'isAdmin`. It's `admin`.

If you click on the `Try` button and go to the Playground, or scroll up a bit to check the existing `User` definition and the property `admin`.

That was my point with the "consistency"

jakebailey: I was just trying to figure out if either of these errors were "this property does not exist", but looking at the site, they are actually two different errors.

jakebailey (approved):

PR #3299, opened 2025-01-12
Add legality update on Special Static Names in ES2022

lazyrainynight:


lazyrainynight:

PR #3307, opened 2025-01-25
Fix default value of esModuleInterop

TomiBelan:
Based on https://github.com/microsoft/TypeScript/blob/v5.7.3/src/compiler/utilities.ts#L9001-L9006

Should I also mention `"node18"`? It seems it is not released yet.


TomiBelan:

jakebailey:
> Should I also mention "node18"? It seems it is not released yet.

No, only after 5.8 in at least beta.

jakebailey (approved):

PR #3308, opened 2025-01-25
Reword strictBuiltinIteratorReturn for consistency

TomiBelan:
From: `false, unless strict is set` to: `true if strict; false otherwise.` matching the other strict options.


jakebailey (approved):

jakebailey:
Thanks!

PR #3311, opened 2025-01-26
ATA: Call delegate.errorMessage upon Errors

seanoch:
This PR adds calls to the user's `delegate.errorMessage()`
The API exists but it currently does nothing

### DIFF
--- a/.changeset/green-bulldogs-vanish.md
+++ b/.changeset/green-bulldogs-vanish.md
@@ -1,0 +1,5 @@
+---
+"@typescript/ata": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/real-falcons-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@typescript/ata": minor
```
jakebailey: I almost wonder if it should just be a patch, since it should have worked before?

seanoch:
@jakebailey Changed to `patch`, let me know if any additional change is needed

PR #3312, opened 2025-01-27
Fixing bug: Unsafe HTML constructed from library input

navya9singh:
This PR fixes https://liquid.microsoft.com/codeql/issues/71221708-d80d-49f5-b177-1b536b3b3e22 where unsafe HTML was constructed from library input. Fix is according to the recommendation provided by https://liquid.microsoft.com/Web/Object/Read/ScanningToolWarnings/Requirements/CodeQL.SM03892#Zguide.

### DIFF
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -17,7 +17,8 @@
   "dependencies": {
     "@typescript/playground-worker": "workspace:*",
     "@typescript/sandbox": "workspace:*",
-    "esbuild": "^0.17.8"
+    "esbuild": "^0.17.8",
+    "striptags": "^3.2.0"
   },
   "devDependencies": {
     "@types/jest": "^29.5.12",
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -17,8 +17,7 @@
   "dependencies": {
     "@typescript/playground-worker": "workspace:*",
     "@typescript/sandbox": "workspace:*",
-    "esbuild": "^0.17.8",
-    "striptags": "^3.2.0"
+    "esbuild": "^0.17.8"
   },
   "devDependencies": {
     "@types/jest": "^29.5.12",
--- a/packages/playground/src/index.ts
+++ b/packages/playground/src/index.ts
@@ -284,7 +284,9 @@ export const setupPlayground = (
 
   // Set up the label for the dropdown
   const versionButton = document.querySelectorAll("#versions > a").item(0)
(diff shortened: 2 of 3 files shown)


packages/playground/src/index.ts
```diff
@@ -284,7 +284,9 @@ export const setupPlayground = (
 
   // Set up the label for the dropdown
   const versionButton = document.querySelectorAll("#versions > a").item(0)
-  versionButton.innerHTML = "v" + sandbox.ts.version + " <span class='caret'/>"
+  // Adding HTML sanitizer to remove unsafe content.
+  const striptags = require('striptags');
```
jakebailey: I don't see this dep in the repo; does it need to be added for this to work? And to be regular imported at the top?

navya9singh: The fix that was suggested in the bug report to make it safe was to use `striptags`. I'm trying to run it locally to see if that works, but nothing shows up :( so I'm using the deploy-preview label here to actually see if anything changed in the UI. However, using the new static site link, the playground just gets stuck.

jakebailey: image


The console error seems to imply it can't find things; I'm not sure if we do anything special to bundle deps on this page or what, gatsby is weird...

orta:
I'd recommend switching the text for that button to use `innerText` instead of `innerHtml` and then manually adding the span, it's got a problem with the DOM insert, which innerText protect against naturally

( But no idea what's going on with that dependency, Gatsby normally bundles deps, maybe it makes a networking request inside its codebase? )

jakebailey:
Yeah, the code I am suggesting is:

```js
const versionButton = document.querySelectorAll("#versions > a").item(0)
versionButton.textContent = "v" + sandbox.ts.version + " "
const caret = document.createElement("spam")
caret.classList.add("caret")
versionButton.appendChild(caret)
```

or similar

packages/playground/package.json
```diff
@@ -17,7 +17,8 @@
   "dependencies": {
     "@typescript/playground-worker": "workspace:*",
     "@typescript/sandbox": "workspace:*",
-    "esbuild": "^0.17.8"
+    "esbuild": "^0.17.8",
+    "striptags": "^3.2.0"
```
jakebailey: The changes to package.json / pnpm-lock can now be reverted.

jakebailey (approved):
Seems to work.

PR #3318, opened 2025-01-30
Fixing codeQL bugs :Updating hash algorithm

navya9singh:
This pr fixes <https://liquid.microsoft.com/codeql/issues/554838ed-6abd-48f1-bfa9-03546781e3c1?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058> and <https://liquid.microsoft.com/codeql/issues/df74a67a-14d7-41b4-9ed2-b8e9c2a184cb?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058>

Reference: https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.Cryptography.10021#Zguidance

### DIFF
--- a/packages/documentation/scripts/generateAttribution.js
+++ b/packages/documentation/scripts/generateAttribution.js
@@ -39,7 +39,7 @@ const gravatarURLForAuthor = (email) => {
       return "https://avatars.githubusercontent.com/u/49038?s=100&u=0b9ac5bf42a8ea2543a05191e150e0213456744e&v=4";
 
     default:
-      return crypto.createHash("sha512").update(email).digest("hex");
+      return crypto.createHash("sha256").update(email).digest("hex");
   }
 };
 
--- a/packages/documentation/scripts/generateAttribution.js
+++ b/packages/documentation/scripts/generateAttribution.js
@@ -39,7 +39,7 @@ const gravatarURLForAuthor = (email) => {
       return "https://avatars.githubusercontent.com/u/49038?s=100&u=0b9ac5bf42a8ea2543a05191e150e0213456744e&v=4";
 
     default:
-      return crypto.createHash("md5").update(email).digest("hex");
+      return crypto.createHash("sha512").update(email).digest("hex");
   }
 };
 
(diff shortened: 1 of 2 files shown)


packages/documentation/scripts/generateAttribution.js
```diff
@@ -39,7 +39,7 @@ const gravatarURLForAuthor = (email) => {
       return "https://avatars.githubusercontent.com/u/49038?s=100&u=0b9ac5bf42a8ea2543a05191e150e0213456744e&v=4";
 
     default:
-      return crypto.createHash("md5").update(email).digest("hex");
```
navya9singh, line 42: I'm not a 100% sure if anything else is needed to make them work. It seems like this should be enough according to the `createHash` function description in the `crypto` package.

jakebailey: The question is whether or not this breaks anything, but maybe it doesn't; I can't imagine a md5 hash of an email address ever worked as a URL?

packages/playground-examples/scripts/generateTOC.js
```diff
@@ -96,7 +96,7 @@ langs.forEach(lang => {
         .replace(/\+/g, "-"),
 
       sortIndex: index,
-      hash: crypto.createHash("md5").update(contents).digest("hex"),
+      hash: crypto.createHash("sha512").update(contents).digest("hex"),
```
jakebailey, line 99: Checking this, the website still works fine here. It's put into `data-hash` in `ShowExamples` and that's just used to cache loading them.

packages/documentation/scripts/generateAttribution.js
jakebailey: Yeah, so this actually can be sha256:

- https://docs.gravatar.com/api/avatars/images/
- https://docs.gravatar.com/api/avatars/hash/

That should pass testing and also actually work.

packages/documentation/scripts/generateAttribution.js
```diff
@@ -39,7 +39,7 @@ const gravatarURLForAuthor = (email) => {
...
+      return crypto.createHash("sha512").update(email).digest("hex");
```
jakebailey:
```suggestion
      return crypto.createHash("sha256").update(email).digest("hex");
```

jakebailey (changes requested):

jakebailey (approved):

PR #3319, opened 2025-02-01
docs: remove the word 'duck typing' in structural-type-system section.

mohammednumaan:
In the [Structural Type System](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html#structural-type-system) section, the documentation states that TypeScript’s type checking can be referred to as `duck typing` because it compares the shape of values. However, this is incorrect due to the following reasons:

-  `duck typing` compares types based on **behavior**.
- `structural typing` compares types based on their **structure** (or) the **members** they contain.

This PR resolves the above issue.

### DIFF
--- a/packages/documentation/copy/en/get-started/TS for JS Programmers.md
+++ b/packages/documentation/copy/en/get-started/TS for JS Programmers.md
@@ -208,7 +208,7 @@ backpack.add(23);
 
 ## Structural Type System
 
-One of TypeScript's core principles is that type checking focuses on the _shape_ that values have. This is sometimes called "duck typing" or "structural typing".
+One of TypeScript's core principles is that type checking focuses on the _shape_ that values have. This is sometimes called "structural typing".
 
 In a structural type system, if two objects have the same shape, they are considered to be of the same type.
 


mohammednumaan:
@jakebailey hey, if you don't mind, can you review this PR? Thanks!

jakebailey:
I can't say I agree with the change; TS has interfaces, function properties, etc, which to me very much qualify.

mohammednumaan:
> I can't say I agree with the change; TS has interfaces, function properties, etc, which to me very much qualify.

@jakebailey yes, but at its core ts is `structurally typed` code, not duck typed. 

> One of TypeScript’s core principles is that type checking focuses on the shape that values have. This is sometimes called “duck typing” or “structural typing”.

Imo, in this context, `or duck typing` shouldn't be included.

orta:
I thought the gains in comparing to duck typing were valuable originally, but I found myself agreeing with the PR in general after reading the [duck typing wikipedia page](https://en.wikipedia.org/wiki/Duck_typing) and [c2](https://wiki.c2.com/?DuckTyping=) 

JavaScript is duck-typing in the sense that its "try it and see", but I've never heard a JS person refer to JS as a duck-typed language (and that is who the doc is aimed at) so dropping it as unnecessary jargon is probably a win. 

Personally, I'd go a bit harder on the structural typing reference:

> One of TypeScript's core principles is that type checking focuses on the _shape_ that values have. This is commonly called "structural typing".

mohammednumaan:
@orta yep. as far as i know, there is another important distinction: 
- in duck typing, the type compatibility is determined at **runtime**.
- in structural typing, the type compatibility is determined at **compile time**.

and since ts at its core is structurally typed, imo its better to remove the `duck typing` to not cause any confusion :thinking:.

mohammednumaan:
@jakebailey any updates on this PR?

PR #3328, opened 2025-02-09
add oneline description to rewriteRelativeImportExtensions

lukekarrys:
I noticed the short description of `rewriteRelativeImportExtensions` was a placeholder (I think?) on the playground.

Screenshot 2025-02-08 at 9 55 30 PM

I tracked it down to this file and I made it a slightly shorter version of the full description to try and match the length of the other `oneline` properties. Please edit as needed 😄

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/rewriteRelativeImportExtensions.md
+++ b/packages/tsconfig-reference/copy/en/options/rewriteRelativeImportExtensions.md
@@ -1,6 +1,6 @@
 ---
 display: "rewriteRelativeImportExtensions"
-oneline: "Does something"
+oneline: "Rewrite TypeScript extensions of relative import paths to their JavaScript equivalent."
 ---
 Rewrite .ts, .tsx, .mts, and .cts file extensions in relative import paths to their JavaScript equivalent in output files.
  


packages/tsconfig-reference/copy/en/options/rewriteRelativeImportExtensions.md
```diff
@@ -1,6 +1,6 @@
 ---
 display: "rewriteRelativeImportExtensions"
-oneline: "Does something"
+oneline: "Rewrite TypeScript extensions of relative import paths to their JavaScript equivalent."
```
jakebailey, line 3: Normally this is a copy of https://github.com/microsoft/TypeScript/blob/34ea32f8ce2771fb3b063f66ec0bc7b0c017d9e4/src/compiler/commandLineParser.ts#L1207 which is actually just the string below.

lukekarrys:
This isn't needed anymore after #3373 🎉

jakebailey:
Shoot, sorry, I entirely forgot about this.

lukekarrys:
No worries! So did I 😁 I finally remembered about it today and was happy it already got fixed.

PR #3333, opened 2025-02-12
Fix a punctuation error in Functions chapter

mtsfer:


leroyanders (approved):

packages/documentation/copy/en/handbook-v2/More on Functions.md
```diff
@@ -870,7 +870,7 @@ const dst = [0];
 src.forEach((el) => dst.push(el));
 ```
 
-There is one other special case to be aware of, when a literal function definition has a `void` return type, that function must **not** return anything.
```
jakebailey, line 873: I think a better fix would be to just swap the comma out for a colon.

PR #3335, opened 2025-02-15
Consistently style code blocks in tsconfig reference

mkantor:
Use `json tsconfig` language identifier for all tsconfig snippets for uniform syntax highlighting on <https://www.typescriptlang.org/tsconfig>.

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/customConditions.md
+++ b/packages/tsconfig-reference/copy/en/options/customConditions.md
@@ -8,7 +8,7 @@ These conditions are added to whatever existing conditions a resolver will use b
 
 For example, when this field is set in a `tsconfig.json` as so:
 
-```jsonc
+```json tsconfig
 {
     "compilerOptions": {
         "target": "es2022",
--- a/packages/tsconfig-reference/copy/en/options/disableFilenameBasedTypeAcquisition.md
+++ b/packages/tsconfig-reference/copy/en/options/disableFilenameBasedTypeAcquisition.md
@@ -7,7 +7,7 @@ TypeScript's type acquisition can infer what types should be added based on file
 
 You can disable this via `disableFilenameBasedTypeAcquisition`.
 
-```json
+```json tsconfig
 {
   "typeAcquisition": {
     "disableFilenameBasedTypeAcquisition": true
--- a/packages/tsconfig-reference/copy/en/options/enable.md
+++ b/packages/tsconfig-reference/copy/en/options/enable.md
@@ -5,7 +5,7 @@ oneline: "Disable the type acquisition for JavaScript projects."
 
 Disables automatic type acquisition in JavaScript projects:
 
-```json
+```json tsconfig
 {
   "typeAcquisition": {
     "enable": false
--- a/packages/tsconfig-reference/copy/en/options/explainFiles.md
+++ b/packages/tsconfig-reference/copy/en/options/explainFiles.md
@@ -16,7 +16,7 @@ example
 
 Using a `tsconfig.json` which has `explainFiles` set to true:
 
-```json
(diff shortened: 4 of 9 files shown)


jakebailey:
There's not really a need for this, all code blocks are formatted during build.

mkantor:
@jakebailey I was inspired to suggest these changes when I noticed that code blocks on the live <https://www.typescriptlang.org/tsconfig> site are inconsistently highlighted.

For example, this one from [`explainFiles`](https://www.typescriptlang.org/tsconfig/#explainFiles):

Screenshot 2025-02-15 at 10 39 00 AM

Uses different colors than this one from [`listEmittedFiles`](https://www.typescriptlang.org/tsconfig/#listEmittedFiles):

Screenshot 2025-02-15 at 10 39 06 AM

I do see now that indentation is irrelevant to what's actually displayed on the page. I could scrub 9ae2236 from this branch.

mkantor:
> I do see now that indentation is irrelevant to what's actually displayed on the page.

Actually, some of the four-space indentations are visible, like these in [`resolveJsonModule`](https://www.typescriptlang.org/tsconfig/#resolveJsonModule):

Screenshot 2025-02-15 at 10 46 20 AM

(Probably because only code blocks with certain language identifiers get auto-formatted?)

Still, I'd be happy to omit the indentation changes (or submit them as a separate pull request; I think I missed some anyway).

mkantor:
I've removed the indentation tweaks, leaving only the changes that will affect syntax highlighting. @jakebailey mind taking another look when you have a chance?

jakebailey:
Looking at the preview, I think this may actually be a regression; the colors appear to have less contrast which may mean that this is somehow undoing some accessibility work we had to do to ensure that contrast is high enough. e.g. here, the top is `json tsconfig`, the bottom is `jsonc`:

image

To my eye, everything should really be `jsonc`. But I don't know that things are fully consistent yet with this PR either, unfortunately.

mkantor:
I assumed the `json tsconfig` identifier was the preferred one because it's more specific, but aesthetically (and readability-wise) I agree with you and prefer the lighter keys. I'd be happy to switch them all to `jsonc` or whatever else is preferred (or maybe the color scheme for `json tsconfig` ought to be changed?).

jakebailey:
To be honest I'm not sure who's in charge of the coloring, other than https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/lib/themes/typescript-beta-dark.json which doesn't actually mention json at all, so it's entirely possible that the colors are a fluke.

orta:
probably a side-effect of the hover implementation on the keys of the tsconfig, likely because you've seen the page it links to (as it would be linking back to the tsconfig reference page)

mkantor:
I hadn't even noticed they were links!

The darker blue comes from here: https://github.com/microsoft/TypeScript-Website/blob/3173c39c72e3405c9c7f313fad2c7ab2f07e5bb8/packages/typescriptlang-org/src/components/layout/main.scss#L138-L140

Locally I hacked in this override and now I see the lighter blue in those code blocks:
```diff
--- a/packages/typescriptlang-org/src/templates/markdown-twoslash.scss
+++ b/packages/typescriptlang-org/src/templates/markdown-twoslash.scss
@@ -137,6 +137,10 @@ pre code a {
   text-decoration: none;
 }
 
+.raised pre code a {
+  color: inherit;
+}
+
 pre .query {
   margin-bottom: 10px;
   color: #137998;
```

jakebailey:
Oh, wow, I didn't know that either. Yeah, we should not let those links change based on clicked status.

mkantor:
I've opened #3337 to make these links use syntax highlighter colors.

PR #3336, opened 2025-02-19
Clearer `NoInfer<Type>` example

dieulot:
For people (such as myself two hours ago) who don’t know that the multiple-arguments union inference (described as “undesirable” in the release notes) occurs in TypeScript, the current example, and thus `NoInfer<Type>`’s usefulness, is incomprehensible.


dieulot:

PR #3337, opened 2025-02-22
Make links in code blocks inherit syntax highlighter colors

mkantor:
See https://github.com/microsoft/TypeScript-Website/pull/3335#issuecomment-2669857953 and preceding discussion.

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/main.scss
+++ b/packages/typescriptlang-org/src/components/layout/main.scss
@@ -138,6 +138,10 @@ html.font-open-dyslexic {
   a {
     color: var(--link-color);
   }
+  pre code a {
+    // Don't interfere with syntax highlighting in code blocks
+    color: inherit;
+  }
 
   h1,
   h2,


packages/typescriptlang-org/src/components/layout/main.scss
```diff
@@ -138,6 +138,10 @@ html.font-open-dyslexic {
   a {
     color: var(--link-color);
   }
+  pre code a {
+    // Don't interfere with syntax highlighting in code blocks
+    color: inherit;
+  }
```
mkantor, line 144: Please let me know if there's a better place/way to make this exception. I couldn't tell if there were any particular SASS conventions I should be following.

mkantor:
With this change links in code are effectively invisible other than a cursor change. They weren't especially noticeable before (despite looking at the docs many times over the years I'd never noticed these links), so this isn't exactly a regression, but I'll happily make further style changes[^1] if desired.

[^1]: My first thought is a `:hover` effect (like adding an underline). That only helps after the link has been pointed at, but makes it more apparent that they're clickable.

orta:
They are supposed to give a hover of the description, but it looks like that isn't happening - its probably not worth blocking the PR but I'll see if I get some time next week to bring that back up to working

PR #3339, opened 2025-02-24
[TS Workbench] Correctly HTML escape rendered code

MichaelMitchell-at:
This has been bugging me for a while

Before:
before

After:
after

### DIFF
--- a/packages/playground/src/ds/createDesignSystem.ts
+++ b/packages/playground/src/ds/createDesignSystem.ts
@@ -154,7 +154,7 @@ export const createDesignSystem = (sandbox: Sandbox) => {
       createCodePre.setAttribute("tabindex", "0")
       const codeElement = document.createElement("code")
 
-      codeElement.innerHTML = code
+      codeElement.textContent = code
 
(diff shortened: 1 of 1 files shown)


packages/playground/src/ds/createDesignSystem.ts
```diff
@@ -154,7 +155,7 @@ export const createDesignSystem = (sandbox: Sandbox) => {
       createCodePre.setAttribute("tabindex", "0")
       const codeElement = document.createElement("code")
 
-      codeElement.innerHTML = code
+      codeElement.innerHTML = htmlEscape(code)
```
jakebailey: Can this just be:

```suggestion
      codeElement.textContent = code
```

Then there's no need to write any escaping code.

MichaelMitchell-at: Good call

MichaelMitchell-at: Updated!

jakebailey (approved):

PR #3355, opened 2025-03-10
Add ./ to package.json conditional export values

dsherret:
> All paths defined in the "exports" must be relative file URLs starting with ./.

https://nodejs.org/api/packages.html#exports


jakebailey:
LGTM but maybe we should fix the devblog site too. @DanielRosenwasser

PR #3358, opened 2025-03-22
Fix getScriptSnapshot for file with empty content

nanokind:
After applying `env.updateFile` to reset a file's content to empty, then use any env.languageService.* method, it would throw an error message stating "Could not find source file: 'index.ts'"

```ts
import {
  createDefaultMapFromCDN,
  createSystem,
  createVirtualTypeScriptEnvironment,
} from '@typescript/vfs';
import ts from 'typescript';

const fsMap = await createDefaultMapFromCDN(
  { target: ts.ScriptTarget.ES2018 },
  '5.7.2',
  false,
  ts
);

const system = createSystem(fsMap);

const compilerOpts = {};
const env = createVirtualTypeScriptEnvironment(system, [], ts, compilerOpts);

env.createFile('/index.ts', 'console.logx()');

env.updateFile('/index.ts', '');

// throws error: Could not find source file: 'index.ts'.
env.languageService.getSemanticDiagnostics('index.ts');
```

Here is the repro: https://stackblitz.com/edit/vitejs-vite-obbb75xc?file=src%2Findex.ts

This PR fixes this

### DIFF
--- a/.changeset/spicy-swans-work.md
+++ b/.changeset/spicy-swans-work.md
@@ -1,0 +1,5 @@
+---
+"@typescript/vfs": patch
+---
+
+Fix getScriptSnapshot for file with empty content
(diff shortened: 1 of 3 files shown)


nanokind:

PR #3365, opened 2025-04-07
Document auto-accessors in classes

mrazauskas:
As I was mentioning on Discord, currently the handbook does not document auto-accessors. This feature was shipped with TypeScript 4.9, but it is still in Stage 3.

To put the text together, I used these sources:
- release notes, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html?#auto-accessors-in-classes
- pull request description, https://github.com/microsoft/TypeScript/pull/49705
- proposal, https://github.com/tc39/proposal-decorators?tab=readme-ov-file#class-auto-accessors

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Classes.md
+++ b/packages/documentation/copy/en/handbook-v2/Classes.md
@@ -285,15 +285,15 @@ By adding the `accessor` keyword in front of a class field, you can define a get
 
 ```ts twoslash
 class Person {
-  accessor name: string | undefined;
+  accessor name = "";
 }
 ```
 
-The above roughly desugars to:
+The above is roughly equivalent to:
 
-```js
+```ts twoslash
 class Person {
-  #name;
+  #name = "";
 
   get name() {
     return this.#name;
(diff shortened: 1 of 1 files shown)


jakebailey:
Should this mention "why" someone might want to do this? I.e. I think the relationship to decorators? Not that those are very documented, either.

mrazauskas:
Ah.. Right! Perhaps simply: less boiler plate code? This is exactly what is visible in the examples. Or?

jakebailey:
Well, the immediate question with this example would be "if it generates me a getter and a setter, why not use a prop"? So I would figure that there's some other reason for them.

mrazauskas:
> why not use a prop

A prop is not a readonly field.

Of course, do I understand this isn’t a real world example. The aim is slightly different. Last week I saw the `accessor` keyword and I did not know what it is. Let’s read the docs, is the first thought. But there was nothing about that.

mrazauskas:
On second thought, this indeed sounds mostly useful in decorators context. So leaving it for the future.

PR #3380, opened 2025-05-04
fix: mobile styles on documentation page and missing react key

### LINKED ISSUE #3270: Mobile friendly website for Handbook and Documentation
The content of the Typescript Handbook that I read was pretty good, but it is not possible to read on mobile (especially in portrait view).

It would be great to read about Typescript when I am stranded away from my laptop. However, when loading or rotating a page to portrait, the text does not re-justify and flows off the screen. When attempting to pinch to zoom out, the page throws an error and cannot even load the page.

It seems like having a web page readable on mobile screens is the most basic requirement in 2024, and it's a bit infuriating trying to simply review the docs.

It also makes one question the validity/stability of the language itself, if the documentation website can't even work correctly (as it seems most would assume the docs website is built with Typescript).

I remember having this issue with the site a few years ago, surprised the problem persists.

Image
Image

connorb08:
fixes: #3270

### DIFF
--- a/packages/typescriptlang-org/src/templates/documentation.scss
+++ b/packages/typescriptlang-org/src/templates/documentation.scss
@@ -48,6 +48,19 @@
     flex-direction: row;
     width: 100%;
 
+    aside.handbook-toc {
+      @media (max-width: $screen-xs) {
+        padding: 0 0.5rem;
+        box-sizing: border-box;
+      }
+    }
+
+    .whitespace.raised {
+      width: fill-available;
+      width: -moz-available;
+      width: -webkit-fill-available;
+    }
+
     @media (max-width: 999px) {
       flex-direction: column;
     }
--- a/packages/typescriptlang-org/gatsby-ssr.js
+++ b/packages/typescriptlang-org/gatsby-ssr.js
@@ -38,5 +38,5 @@ const CustomColorSwitcherCode = () => {
 }
 
 exports.onRenderBody = ({ setPreBodyComponents }) => {
-  setPreBodyComponents(<CustomColorSwitcherCode />)
+  setPreBodyComponents(<CustomColorSwitcherCode key="CustomColorSwitcherCode" />)
 }


connorb08:
> @connorb08 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

packages/typescriptlang-org/src/templates/documentation.scss
```diff
@@ -48,6 +48,19 @@
     flex-direction: row;
     width: 100%;
 
+    aside.handbook-toc {
+      @media (max-width: 999px) {
```
jakebailey: Where did 999 come from? Are you sure this shouldn't be one of the size constants from https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/src/style/globals.scss?

999 does work on my phone (yay), but I'm trying to understand if it'll work for everyone.

connorb08: I used 999 because that is the breakpoint in the stylesheet where it turns from a flex row into a column. You are right though, it definitely makes more sense to use the xs breakpoint, I updated the PR to use that instead. 

It would work either way because the fix is in the `.whitespace.raised` style, but adding the padding just makes it look better / more readable on a mobile screen so the text doesn't go all the way to the edge.

jakebailey: Not sm, like the others here? xs is 480px, so I'm just trying to figure out what is and isn't right as (honestly) not a CSS expert

connorb08: Yep, the media query on [lines 28-30](https://github.com/microsoft/TypeScript-Website/blob/9e33d1d50a695052bb7ddcbaca176faa9fb40880/packages/typescriptlang-org/src/templates/documentation.scss#L28-L30) uses the xs breakpoint to remove the padding for the content. 

At 500px, the page has padding around the content: 
image

Without the padding, when you shrink the page, the text goes all the way to the edge:
image

Adding the breakpoint at xs adds some padding to the page header:
image

Since the content is already padded at anything larger than xs, it makes more sense for that to be the breakpoint.

jakebailey (approved):

PR #3386, opened 2025-05-14
Clarify allowed syntax in tsconfig.json

grillonbleu:
It would be helpful to explicitly state the syntax tsconfig.json affords. The most official reference I’ve found about comments being allowed is [TypeScript issue 4987](https://github.com/Microsoft/TypeScript/issues/4987) being tagged "fixed". For trailing comments, the best I’ve found is [a comment](https://github.com/microsoft/TypeScript/issues/20384#issuecomment-348552936).

Comments are quite valuable in config files and that possibility deserves to be better known. Furthermore this undocumented leniency has tripped up IDEs in the past, validating `tsconfig.json` as strict JSON.


grillonbleu:

PR #3389, opened 2025-05-16
Remove all AppInsights code

jakebailey:
In prep for using something else.


RyanCavanaugh (approved):

packages/typescriptlang-org/gatsby-browser.js
jakebailey, line 1: This will have to be brought back for new analytics when we add it.

PR #3391, opened 2025-05-21
[ATA] Remove duplicate modules from getReferencesForModule

ChaseMalik:
Why?

Go to the playground and try to use the following code:

```ts
import { Client } from "@modelcontextprotocol/sdk/client/index.js";
```

[Link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgYQDbAKYDt4F84BmUEIcARAAIgQAmGqAxhDhgB4xjEwROoD0AZxoBrPg3TYYfYFjqsAdACsBZANwAoIA)

This can trigger 1000s of calls to jsdelivr because imports to zod are not de-duped, resulting in ERR_INSUFFICENT_RESOURCES errors.

This fix removes duplicate modules within the same file (they are already handled cross file).

### DIFF
--- a/packages/ata/src/index.ts
+++ b/packages/ata/src/index.ts
@@ -173,19 +173,21 @@ export const getReferencesForModule = (ts: typeof import("typescript"), code: st
     .filter(f => !isDtsFile(f.fileName))
     .filter(d => !libMap.has(d.fileName))
 
-  return references.map(r => {
-    let version = undefined
-    if (!r.fileName.startsWith(".")) {
-      version = "latest"
-      const line = code.slice(r.end).split("\n")[0]!
-      if (line.includes("// types:")) version = line.split("// types: ")[1]!.trim()
-    }
+  return references
+    .map(r => {
+      let version = undefined
+      if (!r.fileName.startsWith(".")) {
+        version = "latest"
+        const line = code.slice(r.end).split("\n")[0]!
+        if (line.includes("// types:")) version = line.split("// types: ")[1]!.trim()
+      }
 
-    return {
-      module: r.fileName,
-      version,
-    }
-  })
+      return {
+        module: r.fileName,
+        version,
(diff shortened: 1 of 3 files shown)


ChaseMalik:
@microsoft-github-policy-service agree [company="Google"]

ChaseMalik:
@microsoft-github-policy-service agree company="Google"

packages/ata/src/index.ts
```diff
@@ -173,19 +173,21 @@ export const getReferencesForModule = (ts: typeof import("typescript"), code: st
...
-      module: r.fileName,
-      version,
-    }
-  })
+      return {
+        module: r.fileName,
+        version,
+      }
+    })
+    .filter((r, index, self) => self.findIndex(m => m.module === r.module && m.version === r.version) === index)
```
jakebailey, line 190: This is quadradic; how big are these inputs typically?

ChaseMalik: I don't have a scientific answer, but I would think most files don't import more than on the order of 10s of packages - but in the case of @modelcontextprotocol it is >1000. (I also don't think one would ever specify a different module versions for the same module? but just did the safest option for now, but happy to update) - I can also switch to a Set if preferred.

jakebailey: Well, you had said there were 1000s of calls; does that occur without 1000s of entries in this list?

ChaseMalik: Sorry, I updated my response (probably while you were responding, sorry about that), but I would not expect more than 10s of unique packages. So the array would not actually loop through 1000s of entries, but yes it can have 1000s.

jakebailey: Ah, this is just in a single file. Gotcha.

jakebailey (approved):

PR #3393, opened 2025-05-23
fix: improve accessibility of "Subtype vs Assignment" table

rvedotrc:
Currently, the two styles of tick are rather hard to distinguish: two similar shades of blue / green, and two similar font weights. This PR addresses that.

Before:

image

After:

image

The "only if not strict" is now indicated by "✓*" (as opposed to just "✓").

The ✓* is now the same colour as the ✓, the "main blue", since the "muted green" was really hard to distinguish anyway, so served no purpose.

It _might_ even be better to have these marks changed to "✕*", and comment that it's not compatible as long as `strictNullChecks` is `true`. Even though it's not the default, I think it's probably recognised as being good practice.


rvedotrc:

rvedotrc:
@jakebailey How can I help get this moving?

PR #3406, opened 2025-06-17
fix: incorrect deprecation link on Advanced Types page of old handbook

ryanoboril:
Fixes the "Go to New Page" buttons on the Advanced Types page of the old handbook so they redirect to the same subsection of the v2 handbook

### DIFF
--- a/packages/documentation/copy/en/reference/Advanced Types.md
+++ b/packages/documentation/copy/en/reference/Advanced Types.md
@@ -3,7 +3,7 @@ title: Advanced Types
 layout: docs
 permalink: /docs/handbook/advanced-types.html
 oneline: Advanced concepts around types in TypeScript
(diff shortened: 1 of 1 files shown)


packages/documentation/copy/en/reference/Advanced Types.md
```diff
@@ -3,7 +3,7 @@ title: Advanced Types
 layout: docs
 permalink: /docs/handbook/advanced-types.html
 oneline: Advanced concepts around types in TypeScript
-deprecated_by: /docs/handbook/2/types-from-types.html
```
RyanCavanaugh, line 6: The old page covers a ton of different kinds of types, but here you're linking to one very specific concept on a page that's mostly about narrowing and doesn't really serve the same purpose.

RyanCavanaugh (changes requested):

PR #3411, opened 2025-07-06
fix: add -p flag to mkdir to prevent build error on Unix systems

sw1tch3roo:
This PR fixes a build issue on Unix-like systems (e.g., macOS, Linux) where the command mkdir in packages/playground-examples/scripts/copyFiles.js fails because the parent directories may not exist.

Added a check to ensure the target directory exists before calling mkdir.

Replaced mkdir with mkdir -p to avoid errors when the directory tree doesn't exist.

Tested locally: the build now completes successfully on macOS.

Thank you for reviewing this fix!

P.S. This issue doesn't occur on Windows because mkdir there behaves like mkdir -p by default, but it fails on Unix-based systems without the -p flag.


jakebailey:
I exclusively develop this repo on Linux and have never hit this; when would the directory not exist?

Issue #1555, opened 2021-02-08
Module: twoslash does not support tsconfig.json "paths" which is an object


mmorearty:
I can't find any way to get twoslash to support the "paths" property of tsconfig.json. It seems like there is code that's intended to work with object; see [here](https://github.com/microsoft/TypeScript-Website/blob/2a0ad5736cde9181925c05e66825c87095472dbc/packages/ts-twoslasher/src/index.ts#L165-L169). However, when I put this in a twoslash document, I get an error:

```
// @paths: { "*": [ "*", "lib/*" ] }
```

E.g. if you try that on [bug workbench](https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEAcEMBcAsDOAuUBvUAiAVJ1BtLXAGiwBsBLAI2F1AF1QBfAWACgg), you get this:

```
TypeError: r.get is not a function
    at c (https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:1685)
    at i (https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:2360)
    at f (https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:2481)
    at https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:4204
    at g (https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:4244)
    at j (https://www.typescriptlang.org/component---src-pages-dev-bug-workbench-tsx-52a0293ef52c1fc9cf74.js:1:5437)
    at https://www.typescriptlang.org/component---src-pages-dev-bug-workbench-tsx-52a0293ef52c1fc9cf74.js:1:4882
    at n (https://www.typescriptlang.org/17821066377c4486a980330106160c805ea5b4a9-3746668446b11083d639.js:1:23019)

```

I tried every variation on the directive that I could think of. Looking at the code, I think it's probably a bug.

cc @d-fischer , who I think did some recent work on this code in [this commit](https://github.com/microsoft/TypeScript-Website/commit/02b5b3b4532a19749db28c791fe09e9d3d93e123)

orta:
[...]

PR #3425, opened 2025-07-31
Support twoslash directives for `TsConfigOnlyOption` and its list

64qq:
This PR adds support for twoslash directives (e.g., `@path`) for `TsConfigOnlyOption` and its list.

Currently, twoslasher doesn't correctly handle the cases where `type` is `"object"`. [This code example](https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEDMEsBsFMB2BDAtvAXKVB7AJgK4IC0yAdAC4DOAUPAB4AOOAThaAMY6JXsBuyFtGQAjBFkQFUI+C1ABeUAAYaNEBBgIU6LLkIkRlWuvCNkFABZUsAb1AAicMABU9rAG17ZF-YC6oAF8aaFRmNlA7ASFRBEDQSBYcVAcnPSJ4UnsAbhogA) triggers a runtime exception: `TypeError: optMap.get is not a function`
```ts
// @filename: module-a.ts
export const variable: number = 0

// @filename: module-b.ts
// @paths: { "@/*": ["./*"] }
import { variable } from "@/module-a";
```
This problem comes from missing internal API types (`ts.optionDeclarations` is `any`). 

I've added a patch to manage those type definitions in a single project (ts-internals) and added handling for missing cases based on them. Due to type consistency requirements, twoslasher now throws an error when an option with a list of booleans is passed, but that shouldn't be a problem since it's unreachable.

Related: #1555
Related: https://github.com/shikijs/twoslash/issues/197#user-content-fn-1-f77cbcb48e8118449f21b9fde2b4bc96


64qq:

PR #3426, opened 2025-07-31
Error in Modules section

thavguard:
Error in file names


thavguard:

64qq:
I think this is intentional, ref https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-bundlers-typescript-runtimes-and-nodejs-loaders

jakebailey:
Yes, I don't know what this PR is trying to fix here

PR #3427, opened 2025-08-01
Reflect changes from https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/

maschwenk:
https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/

Adds a bunch of stuff about how `--module node20` implies target! This is really cool as I've complained before that `--module` should configure more sensible defaults.

I had some issues running some of the commands in the readme so will followup once I get those sorted out

PR #3428, opened 2025-08-01
Update things for TS 5.9

jakebailey:
Prereq for #3427


packages/tsconfig-reference/scripts/schema/result/schema.json
jakebailey, line 1: Need to go make sure the jsonschema repo is up to date with these.

DanielRosenwasser (approved):

PR #3440, opened 2025-09-03
Modules Theory: require(ESM) is in Node 20

robpalme:
The world changed since this was written.

https://nodejs.org/en/blog/release/v20.19.0

cc @andrewbranch


andrewbranch (approved):
Thanks Rob!

PR #3447, opened 2025-10-07
"infer" rather than "as"

wuyuansushen:
TypeScript uses the assigned value to infer the variable’s type.


wuyuansushen:
@jakebailey

jakebailey:
IDK seems weird to use the word "infer" in the section used to define what "infer" means?

PR #3452, opened 2025-10-17
Correct type reference in Mapped Types example

TheUncertaintim:
Updated the type reference from 'Type' to 'Features' in the Mapped Types section of the TypeScript Handbook.


orta:
Screenshot 2025-10-26 at 09 34 53

It's referring to this "Type" in the code sample above, no need to make this change 👍🏻

PR #3466, opened 2025-11-30
added subpath support to ata

envokerid:
Previously ata didn't work for importing packages under a subpath. This requests fixes it. And Typescript is updated to latest.

**Before**
Screenshot 2025-11-30 171950
**Now**
Screenshot 2025-11-30 172021
Screenshot 2025-11-30 172042

**Before**
Screenshot 2025-11-21 233016
**Now**
Screenshot 2025-11-21 235601


envokerid:
> @microsoft-github-policy-service agree

packages/ata/package-lock.json
jakebailey, line 1: This file should not be here, this repo uses pnpm.

packages/ata/package.json
```diff
@@ -27,6 +27,6 @@
     "jest": "^29.5.0"
   },
   "peerDependencies": {
-    "typescript": ">=4.4.4"
+    "typescript": "^5.9.3"
```
jakebailey, line 30: why is this changing?

jakebailey (changes requested):

packages/ata/package.json
envokerid: This is changed because it gives type error to new projects using new typescript versions. See setupTypeAcquision() in the added screenshots. Should I revert it?

PR #3470, opened 2025-12-07
Bring back the note about type safety of any from the old docs

artur-klesun-st:
https://www.typescriptlang.org/docs/handbook/basic-types.html#any

> After all, remember that all the convenience of any comes at the cost of losing type safety. Type safety is one of the main motivations for using TypeScript and you should try to avoid using any when not necessary.

The new documentation lacks any kind of warning about the dangers of any and why using it is a bad practice which was concerning considering that this is the official documentation of TypeScript.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Everyday Types.md
+++ b/packages/documentation/copy/en/handbook-v2/Everyday Types.md
@@ -54,6 +54,8 @@ const n: number = obj;
 
 The `any` type is useful when you don't want to write out a long type just to convince TypeScript that a particular line of code is okay.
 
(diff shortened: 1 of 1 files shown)


artur-klesun-st:
Maybe even it would be better to add a link to [this section](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any) in Do's and Don'ts page:
> ❌ Don’t use any as a type unless you are in the process of migrating a JavaScript project to TypeScript. The compiler effectively treats any as “please turn off type checking for this thing”. It is similar to putting an @ts-ignore comment around every usage of the variable. This can be very helpful when you are first migrating a JavaScript project to TypeScript as you can set the type for stuff you haven’t migrated yet as any, but in a full TypeScript project you are disabling type checking for any parts of your program that use it.

PR #3482, opened 2025-12-29
Document the JSDoc `@overload` tag

kisaragi-hiu:
The tag is currently only documented in the 5.0 Release Note and not referenced in the JSDoc reference. That's hard to find: I only found it after stumbling onto it on typedoc's website.

This PR adds the tag to the reference. The description and example are mostly copied and rephrased from the 5.0 Release Note and [/docs/handbook/2/functions.html#function-overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). It is placed after `@param` and `@returns` because it also concerns functions.


kisaragi-hiu:

PR #3488, opened 2026-01-06
I updated the version dropdown logic in TypeScript-Website/packages/p…

samay-hash:
I finished adding release dates to each TypeScript release in the Playground.

What I changed,
I updated the version dropdown logic in TypeScript-Website/packages/playground/src/index.ts so that development builds now correctly show their release dates.

Instead of trying to load dates from JSON files, I added logic that:
Detects version strings with the pattern -dev.YYYYMMDD

Extracts the date from the version

Displays it in the format:
(YYYY-MM-DD)
the issues was The existing code attempted to fetch release dates from releases.json and pre-releases.json, and those files only contain version numbers but Development builds already include the date inside the version string but that information wasn’t being parsed or shown in the UI.


samay-hash:
agree

PR #3492, opened 2026-01-09
Moved "Parameter Properties" section above "this  Types"

adriancuadrado:
In the [`this-based type guards`](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-based-type-guards) subsection of the [`this Types`](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-types) section there is an example that uses a parameter property. This may confuse readers that are reading this article from top to bottom since you are using a feature that has not been explained yet.

### DIFF
--- a/packages/documentation/copy/en/handbook-v2/Classes.md
+++ b/packages/documentation/copy/en/handbook-v2/Classes.md
@@ -1079,6 +1079,29 @@ This method makes the opposite trade-offs of the arrow function approach:
 - Only one function per class definition gets allocated, rather than one per class instance
 - Base method definitions can still be called via `super`.
 
+## Parameter Properties
+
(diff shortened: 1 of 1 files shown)


adriancuadrado:
This is the example code that uses parameter properties:
```ts
class FileSystemObject {
  isFile(): this is FileRep {
    return this instanceof FileRep;
  }
  isDirectory(): this is Directory {
    return this instanceof Directory;
  }
  isNetworked(): this is Networked & this {
    return this.networked;
  }
  constructor(public path: string, private networked: boolean) {}
}
```

This code is used in [this-based type guards](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-based-type-guards), but it is in the [next section](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties), where they are actually explained.

PR #3502, opened 2026-02-13
Update for TS 6.0

jakebailey:
This should unbreak TS 6.0 on the website.

https://github.com/microsoft/typescript-go/issues/2767#issuecomment-3898269432


packages/documentation/copy/en/project-config/Compiler Options.md
```diff
@@ -791,7 +800,7 @@ tsc app.ts util.ts --target esnext --outfile index.js
   <td><code><a href='/tsconfig/#libReplacement'>--libReplacement</a></code></td>
   <td><p><code>boolean</code></p>
 </td>
-  <td><p><code>true</code></p>
+  <td><p><code>false</code></p>
```
jakebailey: Annoying that these defaults change because we use the nightly version here, but I can't not have nightly here because it's needed to make other stuff work.

it's not even right, because it doesn't correctly handle `strict`

packages/typescript-vfs/src/index.ts
```diff
@@ -463,7 +476,7 @@ const defaultCompilerOptions = (ts: typeof import("typescript")): CompilerOption
     suppressOutputPathCheck: true,
     skipLibCheck: true,
     skipDefaultLibCheck: true,
-    moduleResolution: ts.ModuleResolutionKind.NodeJs,
+    moduleResolution: ts.ModuleResolutionKind.Bundler,
```
jakebailey: This is technically a break, but things get weird otherwise :|

.changeset/poor-bears-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@typescript/vfs": patch
+---
+
+Update lib files list
```
jakebailey:
```suggestion
Update for compat with TS6
```

packages/tsconfig-reference/scripts/msbuild/generateMarkdown.ts
```diff
@@ -15,6 +15,7 @@ import matter from "gray-matter";
 import prettier from "prettier";
 
 import remark from "remark";
+// @ts-ignore - no types available
 import remarkHTML from "remark-html";
```
DanielRosenwasser, line 19: looks like it has types - I'm guessing you don't want to upgrade though

RyanCavanaugh (approved):

packages/tsconfig-reference/scripts/msbuild/generateMarkdown.ts
jakebailey: we are like 100000 versions behind, it's not worth it

DanielRosenwasser (approved):

PR #3514, opened 2026-02-23
docs: clarify path base for project references

skdas20:
References: https://github.com/microsoft/TypeScript/issues/50529

This clarifies what `references[].path` is resolved against in both places users look for project references docs:

- `tsconfig` option docs (`/tsconfig#references`)
- Handbook page (`/docs/handbook/project-references.html`)

Added explicit wording that each reference path is resolved relative to the `tsconfig.json` file where it appears.


skdas20:

PR #3521, opened 2026-03-03
Migrate away from `@import` in SCSS, add some JSDoc types

DanielRosenwasser:
This was mostly accomplished by runnning

```
pnpx sass-migrator module --migrate-deps [...]
```

on every scss file we have with an `@import`.

See more at https://sass-lang.com/documentation/breaking-changes/import/


jakebailey (approved):
Seems fine but I don't really get why `globals.` comes in. The linked page doesn't talk about `globals.` at all.

vskjt4m5vp-sketch (approved):

PR #3523, opened 2026-03-09
fix(sandbox): getInitialCode should not use URL hash without full prefix

Josh-Cena:
I'm hunting down https://github.com/typescript-eslint/typescript-eslint/issues/8780 (see more context in the issue discussion), and I finally got to `getInitialCode`. It turns out that the hash may start with `#code` but then not start with `#code/`, causing malformed initial code to be parsed.

### DIFF
--- a/.changeset/small-animals-switch.md
+++ b/.changeset/small-animals-switch.md
@@ -1,0 +1,5 @@
+---
+"@typescript/sandbox": patch
+---
+
+getInitialCode should not use URL hash without full prefix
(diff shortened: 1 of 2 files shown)


jakebailey:
Thanks; you'll need to `pnpm changeset`, though (patch release is fine).

Josh-Cena:
That's real speed ^^ I was trying to avoid a local clone.

Josh-Cena:
Hi can this be merged

jakebailey (approved):
Sorry, I didn't click the green button

PR #3542, opened 2026-04-09
Add declare global guidance to the declaration reference

### LINKED ISSUE #3409: `declare global` is not explained anywhere
**Page URL:** https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

**Issue:** It doesn't explain what `declare global` is, how to use it, how it works, and when and why you should use it.

I think [this](https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html) is the most appropriate place to explain this, but I am not even sure about that. Maybe it might fit somewhere else, I don't really know.

This construct is mentioned [here](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html), but only as an example and doesn't really get into details about its mechanics.

ShipItAndPray:
## Summary

This adds a new `declare global` section to the Declaration Reference handbook page.

The change explains:
- what `declare global` is used for
- how to describe a global-modifying module
- why `export {}` is needed in declaration files that only use `declare global`
- where to read more about global augmentation

This also improves searchability for people looking for `declare global`, since the term and explanation now appear directly on the page mentioned in the issue.

Closes #3409.

## Validation

- Verified the markdown diff with `git diff --check`
- Did not run the repo's `pnpm` checks in this environment because `pnpm` is not installed on this machine

### DIFF
--- a/packages/documentation/copy/en/declaration-files/By Example.md
+++ b/packages/documentation/copy/en/declaration-files/By Example.md
@@ -249,3 +249,35 @@ Use `declare function` to declare functions.
 declare function greet(greeting: string): void;
 ```
 
+## Global augmentation with `declare global`
+
+_Documentation_
+
+> The `magic-string-time` module adds a `startsWithHello()` method to every string when it is imported.
+
+_Code_
+
+```ts
+import "magic-string-time";
+
+"hello world".startsWithHello();
+```
+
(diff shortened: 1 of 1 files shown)


ShipItAndPray:

PR #3563, opened 2026-05-01
Update documentation for TypeScript 6.0 compiler option changes

5cover:
This PR updates TypeScript Website documentation to match the TypeScript 6.0 compiler option defaults, removals, and deprecations described in the 6.0 release notes.

The main goal is to remove stale guidance that describes TypeScript 5.x behavior as the current default. In particular, the docs now make clear that projects should explicitly opt into global `@types` packages, modern module resolution and emit settings, and supported ECMAScript targets.

## Rationale

TypeScript 6.0 changes several long-standing defaults and deprecates older compiler behaviors. Some existing docs still described the previous behavior as current, which could lead users to:

- expect all visible `node_modules/@types` packages to be loaded automatically;
- rely on `rootDir`, `module`, `target`, and interop defaults from older TypeScript versions;
- use deprecated module emit and resolution modes;
- compile command-line files while expecting `tsconfig.json` to be ignored;
- follow examples using `target: ES5`, `outFile`, legacy namespace `module` syntax, or `no-default-lib` directives.

This PR aligns the current reference, handbook, project config, and tutorial content with TypeScript 6.0 while preserving historical release notes.

## Changes

- Updated `types` and `typeRoots` documentation for the new `types: []` default, including the `types: ["*"]` escape hatch for the old visible-`@types` enumeration behavior.
- Updated `rootDir` and related `outDir` docs for the TypeScript 6.0 default of the directory containing the `tsconfig.json`.
- Refreshed default-value docs for `strict`, `module`, `target`, `noUncheckedSideEffectImports`, and `libReplacement`.
- Added ES2024/ES2025 library documentation and documented that `DOM` now includes the declarations previously split across `DOM.Iterable` and `DOM.AsyncIterable`.
- Updated ES5 target guidance to reflect its deprecation and adjusted active examples to use supported targets or describe ES5 behavior as historical.
- Marked `downlevelIteration` as deprecated because it only affects ES5 emit.
- Updated module resolution guidance for the deprecation of `node`/`node10`, `classic`, and `baseUrl`, including migration guidance toward `paths`, `nodenext`, and `bundler`.
- Updated module emit docs for deprecated/removed AMD, UMD, System, `none`, and `outFile` workflows, steering users toward external bundlers.
- Updated interop documentation for TypeScript 6.0 defaults around `esModuleInterop`, `allowSyntheticDefaultImports`, and `alwaysStrict`.
- Updated namespace documentation to mark legacy `module X {}` namespace syntax as deprecated while keeping `namespace X {}` documented.
- Updated triple-slash directive docs to mark `/// <reference no-default-lib="true"/>` as deprecated and recommend `noLib` instead.
- Updated project configuration docs for the new command-line-file behavior: specifying files is now an error when a `tsconfig.json` would otherwise be loaded unless `--ignoreConfig` is used.
- Updated generated compiler option pages where visible defaults or descriptions still reflected pre-6.0 behavior.

## Validation

- Performed targeted searches for stale current-doc references to the updated 6.0 behaviors.
- Historical release notes were left intact unless they were part of the TypeScript 6.0 release-note source material.


5cover:

packages/documentation/copy/en/handbook-v2/Basics.md
rdsalcedo: Replace these 6 lines with:
Notice that our `person` and `date` parameters no longer have type annotations.

rdsalcedo, line 370: Since this downleveling no longer happens by default, maybe this section should be removed. If not then it should at least explain that this only happens if you specify a target of ES5 and doing so results in error TS5107

rdsalcedo (commented):
I just started reading the handbook and noticed some of the out of date stuff in `Basics.md` . I found this pull request that addresses part of what I noticed, so that is the only file that I commented on.

5cover:
I implemented your suggestions. I rewrote the downleveling section to explain downleveling in general, with the template string behavior as an example for pre-ES5 targets.

PR #3570, opened 2026-05-20
docs: update `"target"` default value to `"es2025"`

zac1st1k:
References:

Previous change default `target` to `latestStandard`.
https://github.com/microsoft/TypeScript/commit/95e3aaa903

Later change updated `latestStandard` to `ES2025`.
https://github.com/microsoft/TypeScript/commit/ee09569714#diff-e9fd483341eea176a38fbd370590e1dc65ce2d9bf70bfd317c5407f04dba9560R7679-R7680

This pull request updates the default value for the TypeScript `target` compiler option to `ES2025` in both the documentation and the configuration scripts. Previously, the default depended on the value of the `module` option, but now it is always set to `ES2025`.

**Documentation update:**

* Updated the default value for `--target` in `Compiler Options.md` to always be `ES2025`, removing the conditional logic based on the `module` setting.

**Configuration script update:**

* Changed the `target` default in `tsconfigRules.ts` to `ES2025`, replacing the previous conditional default logic.


zac1st1k:

PR #3587, opened 2026-06-19
Remove innerHTML uses

jakebailey:
Stop using innerHTML across the board. After this, the only remaining uses are for prerendered content (like code highlighting or gatsby output).


packages/create-typescript-playground-plugin/template/src/vendor/utils.ts
```diff
@@ -9,6 +9,6 @@ export const requireURL = (path: string) => {
 /** Use this to make a few dumb element generation funcs */
 export const el = (str: string, el: string, container: Element) => {
   const para = document.createElement(el)
-  para.innerHTML = str
+  para.textContent = str
```
jakebailey: Maybe this should just be deleted from the template entirely

navya9singh (approved):

PR #3605, opened 2026-07-06
Rename local typescript dep to avoid package manager name bugs

jakebailey:
If you do `"typescript": "@typescript/typescript6"`, then some package managers seem to replace `@typescript/typescript6`'s own `typescript` dep with... `@typescript/typescript6`.


andrewbranch (approved):
Checks out in npm, pnpm, and yarn classic. Yarn berry tries (and of course fails) to patch this package when it's aliased as `typescript`, but of course that's not new to this PR.

Issue #3155, opened 2024-06-12
Playground code runner UX is confusing


rotu:
The TypeScript Playground is confusing when trying to run code. It would be great to update the [handbook](https://www.typescriptlang.org/play##handbook-3) and better yet to update the interface to make the expected behavior more clear.

1. If Settings -> "Run JavaScript in Browser" is disabled, the "Run" button is still enabled. Clicking it creates an error in the native browser console "can't access property "id", plugin is undefined". It's unclear what the "Run JavaScript in Browser" checkbox supposed to do (is there a way for the browser interface to run javascript *out of browser*? I think maybe it's from a previous version of the playground where JavaScript would auto-run instead of having a run button?)

2. Running code seems to fail if TSConfig Module is not "CommonJS", "None", or "Preserve".
It fails silently if Module is "AMD" or "UMD". There *is* a helpful error if you have top-level exports ("[WRN]: "Tip: Change the Module setting to "CommonJS" in TS Config settings to allow top-level exports to work in the Playground"). I expect to see a warning or error if the settings are incompatible with the runner, and it's not clear which settings are compatible with the runner (e.g. I *expect* ES modules to work but *not* CommonJS in the browser, but the opposite appears true)

3. Attempting to import a package from npm often shows type hints. But running the code results in a misleading error, which is contextually inappropriate. (there is no "dependency list" that I know of to check and the import declarations *are* at the top of the user code in the code displayed under the `.js` tab).

  > [ERR]: "Executed JavaScript Failed:" 
  > [ERR]: Check dependency list! Synchronous require cannot resolve module 'jest'. This is the first mention of this module!

Or:
  > [ERR]: "Executed JavaScript Failed:" 
  > [ERR]: import declarations may only appear at top level of a module

Issue #2541, opened 2022-09-22
Playground: Add ability to stop a running script


JLCarveth:
When running some code in the playground, there is no way to stop the execution of a running program (a la Ctrl+C in the terminal). This can lead to some weird behaviour. For example, set an interval to log `A` to the console periodically and run the code. Changing the output to `B` causes `A` and `B` to both be output to the console.

![image](https://user-images.githubusercontent.com/23156861/191822678-b15e9449-f941-49d2-bcd2-f918f1fb286a.png)

**Repro:**
1. Simply run the following line of code in the playground, and run the code.
```
let interval = setInterval(() => { console.log('A') }, 1000);
```
2. Modify the code, and 'run' again:
```
let interval = setInterval(() => { console.log('B') }, 1000);
```

**Expected behavior:**
I expect only 'B' to be printed.
**Actual behavior:**
'B' and 'A' are being printed.

orta:
This is kind of the halting problem, but I dont think there's a way in the JS runtime to eval your code and be able to stop all side-effects like your setInterval from being ran. This is not really feasible without [shadowrealms](https://github.com/tc39/proposal-shadowrealm) I expect and that brings with it a whole host of other issues which make this not really doable.

PR #3617, opened 2026-07-26
Playground: exit loops which block the page instead of freezing the tab when running code

dinesh3018:
## The problem

Open https://www.typescriptlang.org/play, type:

```ts
while (true) {}
```

and press **Run**. The tab freezes permanently — the only way out is killing the tab, losing the code and any unsaved state. Same story for `for(;;)`, accidental `i--` instead of `i++`, and every other runaway loop a learner writes. (Filing this as a PR rather than an issue per the issue-template guidance that non-critical reports should come as PRs; the underlying bug context is below.)

The cause is that the Run button evals the emitted JS directly on the page's main thread (`packages/playground/src/sidebar/runtime.ts`) — the `try`/`catch` there can handle a thrown error, but nothing can interrupt non-termination, which is what was concluded in #2541 ("Add ability to stop a running script", closed as not feasible without ShadowRealms). Related Run-UX discussion: #3155.

## The fix

The same approach JSBin ([`loop-protect`](https://github.com/jsbin/loop-protect)), CodePen, freeCodeCamp and the Babel REPL ([babel/website#352](https://github.com/babel/website/issues/352)) converged on for main-thread-eval playgrounds — instrument loops before eval, since a Worker/iframe runner would break DOM access and the live console capture:

- A new `insertLoopProtection` (in `packages/playground/src/sidebar/loopProtection.ts`) parses the emitted JS with the already-loaded TypeScript AST and inserts `if (__tsPlaygroundLoopProtection.hit(id, line)) break;` at the top of every `for`/`while`/`do` loop body. Unbraced bodies are wrapped in a block; nothing is ever inserted *before* a loop, so `if (x) while (y) z()`, `else`-bodies and labelled loops stay valid. No newlines are added, so runtime error positions still match the JS the user can see.
- The runtime guard tracks a "synchronous burst": starting a burst arms a zero-delay timeout, and until the event loop gets a turn every guard check belongs to that burst. Once a burst has blocked the page for over **1000ms**, every guarded loop exits and the Logs pane reports which line looked infinite (subsequent loops in the same exhausted burst get distinct "exited early" copy rather than being blamed as infinite). This also catches nested infinite loops and loops whose individual iterations are slow, while never touching `await`-loops, `setInterval` callbacks, or ordinary sub-second loops.
- Guards degrade safely: code that doesn't parse, and edge cases like declaration-bodied loops (`while (true) let x = 1;` — a runtime SyntaxError which wrapping would legalize) are left byte-for-byte untouched so they behave exactly as today.
- A **"Disable Loop Protection"** setting (applies on the next Run, so it's rendered without the restart-required banner) is there for intentionally long-running loops, with copy in `en/playground.ts` and entries in the playground handbook ("Settings Panel" and "Running Code").
- Known blind spots, same as all the prior art: infinite recursion (stack-overflows on its own) and long non-loop synchronous work. Stray `setInterval`s from earlier runs (#2541's other half) are intentionally out of scope.

## Testing

- `pnpm --filter=@typescript/playground test` — 18 tests, including: nested/unbraced/labelled loops, `for-of` over an endless generator, `try/finally` bodies, loops in single-statement position, a finite loop in a `setTimeout` callback queued behind an over-budget script (must *not* be broken), an over-budget `async` loop that yields (must *not* be broken), and line-number/`SyntaxError`-preservation checks. The tests wrap the guard in a hard deadline that throws, so a protection regression fails the suite instead of hanging jest/CI.
- `pnpm --filter=@typescript/playground build` (tsc) is clean, the full `pnpm bootstrap` build passes, and new/changed files pass the repo prettier config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/playground-handbook/copy/en/Running Code.md
+++ b/packages/playground-handbook/copy/en/Running Code.md
@@ -8,6 +8,7 @@ In the editor toolbar, the one which is not visible because you're reading this 
 
 - Take the code in the editor and convert it to JS
 - Remove references to `"reflect-metadata"` if you are using decorators
+- Add guards to loops so that code which blocks the page for more than a second exits its loops instead of freezing the tab (can be turned off via "Disable Loop Protection" in Settings)
 - Run that code within the context of your current browser session
 - Capture any `console.log`, `.error`, `.warn` and `.debug` calls and show them in the sidebar "Logs" tab.
 
--- a/packages/playground-handbook/copy/en/Settings Panel.md
+++ b/packages/playground-handbook/copy/en/Settings Panel.md
@@ -12,6 +12,10 @@ Turns off [Type Acquisition](/play#handbook-5) which means that when importing c
 
 When the editor loses focus, or compiler flags change, the Playground will replace the URL in your browser. This doesn't change the behavior of the back button, but it does add history entries in the browser. You can turn off this behavior via this setting, and you can use the 'Share' (or press <kbd>cmd/ctrl</kbd> + <kbd>s</kbd>) to copy the sharable URL.
 
+**"Disable Loop Protection"**
+
+When running code, the Playground adds a small guard to every loop so that a loop which blocks the page for over a second (like `while (true) {}`) exits instead of freezing the browser tab. If your code intentionally runs long loops, you can turn the guards off via this setting — it applies the next time you press "Run", no reload needed.
+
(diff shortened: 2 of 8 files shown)


dinesh3018:
Closing — this was opened from the wrong account by mistake. Apologies for the noise.

PR #3619, opened 2026-07-28
fix(documentation): make mobile sidebar dismissible

praffiii:
## Summary

Makes the mobile documentation sidebar fully dismissible while preserving its existing design, animation, and desktop behavior.

## What

- Adds dismissal through the hamburger trigger, backdrop, navigation links, and Escape.
- Restores focus to the hamburger trigger after dismissal.
- Exposes the drawer state with accurate accessible attributes and labels.
- Prevents interaction and scrolling behind the open drawer while keeping the sidebar scrollable.
- Safely handles Escape when the playground samples popover is absent.
- Adds focused mobile and desktop regression coverage.

## Why

The reason I made these changes is that navigating the documentation becomes difficult when I split my screen between the browser and my IDE.

## Validation

- `pnpm --filter=typescriptlang-org test`
- `pnpm test`
- Manually tested at 390×844 in Helium:
  - trigger, backdrop, navigation-link, and Escape dismissal
  - focus restoration and accessible state
  - background interaction and scroll locking
  - independently scrollable sidebar
  - mobile-to-desktop and desktop-to-mobile resizing
  - repeated open/close cycles with no runtime errors

### DIFF
--- a/packages/typescriptlang-org/src/components/layout/Sidebar-mobile.test.ts
+++ b/packages/typescriptlang-org/src/components/layout/Sidebar-mobile.test.ts
@@ -1,0 +1,133 @@
+/** @jest-environment jsdom */
+
+import { setupMobileSidebar } from "./Sidebar-mobile"
+
+const renderSidebar = (backgroundInert = false) => {
+  document.body.innerHTML = `
+    <div id="site-root">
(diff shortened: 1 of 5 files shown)


praffiii:

PR #3621, opened 2026-07-31
docs: rm error code example and add more practical one and change casts to type assertions

legend80s:
What's the problem of current docs?

## 1. Type error and Inappropriate example

```js
// You can specify an HTML Element with DOM properties
/** @type {HTMLElement} */
var myElement = document.querySelector(selector)
element.dataset.myData = ""
```

1 `element` should be `myElement`

2 type error and misplaced Example

```
Type 'HTMLElement | null' is not assignable to type 'HTMLElement'.
  Type 'null' is not assignable to type 'HTMLElement'.
```

This example doesn't fit the first use case of `@type` — it's better suited for demonstrating casts. It should be removed from the first part of  `@type` introduction.

> It's a perfect example for `cast`:

```js
// You can specify an HTML Element with DOM properties
const myElement = /** @type {HTMLElement} */ (
  document.querySelector("#selector")
)

myElement.dataset.myData = ""
```

No more errors and the inline usage of `@type` assertion is well explained.

## 2. Rename "Casts" to "Type Assertions"

It's functionally identical to TypeScript's `as` — simply written as inline `@type` in JSDoc!

C-style casts imply a runtime type conversion, whereas TypeScript's `as` syntax and JSDoc's inline `@type` are simply type assertions that do not affect runtime behavior. Thus, "Type Assertions" is a more accurate and less misleading term.

## 4. Add JSDoc equivalent for `as unknown as T`

The original docs don't show how to write nested inline assertions in JSDoc — a pattern many users don't know about.

```ts
const x = "hello" as unknown as number;
```

```js
const x = /** @type {number} */ (/** @type {unknown} */ ("hello"))
```

### DIFF
--- a/packages/documentation/copy/en/javascript/JSDoc Reference.md
+++ b/packages/documentation/copy/en/javascript/JSDoc Reference.md
@@ -78,11 +78,6 @@ var win;
 
 /** @type {PromiseLike<string>} */
 var promisedString;
-
-// You can specify an HTML Element with DOM properties
-/** @type {HTMLElement} */
-var myElement = document.querySelector(selector);
-element.dataset.myData = "";
 ```
 
 `@type` can specify a union type &mdash; for example, something can be either a string or a boolean.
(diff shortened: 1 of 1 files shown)


legend80s:
> @legend80s please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

legend80s:

PR #3628, opened 2026-08-16
Update "types" default value docs

### LINKED ISSUE #63753: TypeScript doesn't add "node" to "types" in compilerOptions, even though node_modules/@types/node exists
### 🔎 Search Terms

"typescript default config values", "default tsconfig", "What config is used when there's no tsconfig", "What is tsc part of?", "Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.", "After installing tsx it broke process", "Is tsconfig part of the typescript repository?"

- Versions: ``"typescript": "^7.0.2"`` was the version configured

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
console.log(process.env)
```

### 🙁 Actual behavior

> Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.

### 🙂 Expected behavior

The documentation here https://www.typescriptlang.org/tsconfig/#types says:

> By default all visible ”@types” packages are included in your compilation. Packages in ``node_modules/@types`` of any enclosing folder are considered visible. For example, that means packages within ``./node_modules/@types/``, ``../node_modules/@types/``, ``../../node_modules/@types/``, and so on.

So, if ``@types/node`` is present, why doesn't it include it

## Description

The documentation here https://www.typescriptlang.org/tsconfig/#types says:

> By default all visible ”@types” packages are included in your compilation. Packages in ``node_modules/@types`` of any enclosing folder are considered visible. For example, that means packages within ``./node_modules/@types/``, ``../node_modules/@types/``, ``../../node_modules/@types/``, and so on.

So, I should ever see

```
Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
```

But I do see it, why does it happen, how can I debug this issue. I've tried basic things but I feel like either there's some cache file or something, or I do not understand why it suddenly came back again.

## Re-production Steps

Unclear but

- Create a file and name it ``index.tsx``
- ``npm install --save-dev ``@types/node``
- Add ``console.log(process.env)`` in the file
- Restart VS Code

not guaranteed, I need help debugging

karl-police:
``types`` defaults to ``[]``, however this is not mentioned in the documentation and confused me, so

Fixes https://github.com/microsoft/TypeScript/issues/63753

### DIFF
--- a/packages/tsconfig-reference/copy/en/options/types.md
+++ b/packages/tsconfig-reference/copy/en/options/types.md
@@ -3,7 +3,7 @@ display: "Types"
 oneline: "Specify type package names to be included without being referenced in a source file."
 ---
 
-By default all _visible_ "`@types`" packages are included in your compilation.
+By default ``types`` is set to ``[]``. For versions below TypeScript 6.0, by default all _visible_ "`@types`" packages are included in your compilation.
 Packages in `node_modules/@types` of any enclosing folder are considered _visible_.
 For example, that means packages within `./node_modules/@types/`, `../node_modules/@types/`, `../../node_modules/@types/`, and so on.
 
--- a/packages/tsconfig-reference/copy/en/options/types.md
+++ b/packages/tsconfig-reference/copy/en/options/types.md
@@ -3,7 +3,7 @@ display: "Types"
 oneline: "Specify type package names to be included without being referenced in a source file."
 ---
 
-By default ``types`` is set to ``[]``. For versions below TypeScript 6.0, by default all _visible_ "`@types`" packages are included in your compilation.
+By default `types` is set to `[]`. For versions below TypeScript 6.0, by default all _visible_ "`@types`" packages are included in your compilation.
 Packages in `node_modules/@types` of any enclosing folder are considered _visible_.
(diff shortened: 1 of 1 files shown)


packages/tsconfig-reference/copy/en/options/types.md
```diff
@@ -3,7 +3,7 @@ display: "Types"
 oneline: "Specify type package names to be included without being referenced in a source file."
 ---
 
-By default all _visible_ "`@types`" packages are included in your compilation.
+By default ``types`` is set to ``[]``. For versions below TypeScript 6.0, by default all _visible_ "`@types`" packages are included in your compilation.
```
RyanCavanaugh:
```suggestion
By default `types` is set to `[]`. For versions below TypeScript 6.0, by default all _visible_ "`@types`" packages are included in your compilation.
```

RyanCavanaugh (approved):

jakebailey (approved):

PR #3629, opened 2026-08-18
Fix syntax error in AsyncDisposable example

MantasEdine:
The AsyncDisposable example in the variable declarations reference has a syntax error: `const db = this.db:` uses a colon instead of a semicolon, so the example doesn't compile.

Separately, I noticed the `transfer` function uses `using tx` on an async-disposable resource , should that be `await using tx` so the async disposal is actually awaited? Happy to include that fix too if you'd like !


jakebailey (dismissed):

jakebailey:
Yeah, I believe you're right on the second count, so I'd fix it

jakebailey (approved):

PR #3642, opened 2026-09-03
docs: fix DOM type definitions link

seanxuu:
## What changed

- Update the DOM Manipulation tutorial to link to the current DOM type-definition baseline.

## Why

The previous link pointed to `microsoft/TypeScript/src/lib/dom.generated.d.ts`, which was removed in the TypeScript 7 repository layout. The replacement points to the maintained `TypeScript-DOM-lib-generator` baseline with the same generated DOM definitions.

## Validation

- `git diff HEAD^ HEAD --check`
- Confirmed the replacement URL returns HTTP 200.

References microsoft/TypeScript#64118

PR #3649, opened 2026-09-08
Fix broken link to DOM d.ts file

tyler-breisacher-zipline:


tyler-breisacher-zipline:
I'm not sure if this is the right place to link to but I think so.

jakebailey:
This is identical to #3642

